PR #19, opened 2023-02-07
Add i18next

bruce-glazier:
### Introduce [i18next](https://www.i18next.com)
Using the i18next.t() function in place of hard-coded strings will pave the way for text extraction, integration with an automated translation provider and delivery of translations back into the repo.


package.json
```diff
@@ -45,5 +46,8 @@
     "react-native-device-info": "^10.3.0",
     "react-native-keychain": "^8.1.1"
   },
-  "dependencies": {}
+  "dependencies": {
+    "i18next": "^22.4.9",
+    "react-i18next": "^12.1.5"
```
markdlv, line 51: This is going to be a tricky thing- ideally, we want to be _very_ selective about `dependencies`.  If a consumer of this setup is using `i18next` themselves, that puts their version and ours at odds and it can be confusing to the developer "who wins" on exact version.  The answer is obviously in yarn.lock, but you can imagine how an unintended version might slip by someone.

For now: default to always using peer dependencies.  Over time, we can move things to dependencies, or also "optional dependencies" strategy ([example](https://wix.github.io/react-native-ui-lib/docs/getting-started/setup#optional-dependencies)).  But to start, I like putting these dependencies in the consumer-developer's face so they can stay in full control (and reduce surprises).

lib/package.json
```diff
@@ -0,0 +1,3 @@
+{
+    "name": "lib"
+}
```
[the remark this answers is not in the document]
bruce-glazier: It enables imports direct from `lib` without having to use relative paths, i.e, `import i18n from 'lib/i18n'`

package.json
bruce-glazier: 👍 Thanks for the explanation.

markdlv (approved):

PR #21, opened 2023-02-09
Add BaseImage component

bruce-glazier:
## Changes

Adds BaseImage component using [react-native-fast-image](https://github.com/DylanVann/react-native-fast-image) which provides several performance advantages over using React Native's Image directly.

## Screenshots

https://user-images.githubusercontent.com/76954025/217906981-9974cd1a-5d34-40cc-ba0b-79f26b79d0c5.mov

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -82,6 +82,15 @@ PODS:
     - hermes-engine/Pre-built (= 0.71.1)
   - hermes-engine/Pre-built (0.71.1)
   - libevent (2.1.12)
+  - libwebp (1.2.4):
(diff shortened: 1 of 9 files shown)


bruce-glazier:
This would have been useful for caching and improving performance when fetching remote images however I am not going to proceed with adopting `react-native-fast-image` due to the amount of unresolved open bugs. Just attempting to write unit tests I hit a bug where the callbacks (necessary to provide a placeholder image) never fire and I found several open bugs for this. There is even a PR over two years old with the fix that no one has merged. We can revisit this down the road if we find we really need something like this.

PR #23, opened 2023-02-09
Refactor awkward username param, and allow for loading stored auth token

markdlv:


src/common/SecureStore.ts
```diff
@@ -35,9 +35,9 @@ export class SecureStore<Stored> {
    * Stores an object as { username, password } at key com.lifeomic.securestore/<instanceIdentifier>
    * @param object
    */
-  async setObject(itemId: string, object: Stored) {
```
markdlv, line 38: The way I'd describe the original intent was this - "Secured storage involves a username/password, so consumers can pass a username as this first param - then they get it back as part of the object in `getObject`."  Realizing that's pretty awkward (namely: you get that extra `username` prop that's not in your `Stored` type from `getObject) - this is how I'd characterize this PR: "The fact that storage includes a username & password is an implementation detail of using keychain for iOS.  This should not leak to SecureStore consumers, who just want to securely store/retrieve a typed object."

src/hooks/useOAuthFlow.tsx
```diff
@@ -83,6 +93,10 @@ export const OAuthContextProvider = ({
     [authConfig, clearAuthResult, storeAuthResult],
   );
 
+  useEffect(() => {
+    initialize();
+  }, [initialize]);
+
```
sternetj, line 99: Could this just be moved into `useAuth`?

sternetj (approved):

loscm:
Requested review from a random `@lifeomic/security` team member, @indigocarmen, due to term: **auth**.

src/hooks/useOAuthFlow.tsx
markdlv: The `initialize` call?  It could, but it creates some testing challenges.  When you have a no-deps, first-render useEffect, it makes it more challenging to test.  In this repo, I was running into errors about that first render not being in an `act` but having a side effect.  I could see there being some really nice advantages to externalizing that initialization (perhaps optionally, long-term).  I do agree it feels weird in here - or will long-term.. let's definitely keep playing with this as we go.

PR #24, opened 2023-02-09
fix: Add skeleton of Settings screen, along with i18next fixes

markdlv:
## Screenshots
Simulator Screen Shot - iPhone 14 - 2023-02-09 at 18 50 34


example/ios/Podfile.lock
```diff
@@ -593,11 +593,11 @@ SPEC CHECKSUMS:
   Flipper-RSocket: d9d9ade67cbecf6ac10730304bf5607266dd2541
   FlipperKit: cbdee19bdd4e7f05472a66ce290f1b729ba3cb86
   fmt: ff9d55029c625d3757ed641535fd4a75fedc7ce9
-  glog: 791fe035093b84822da7f0870421a25839ca7870
+  glog: 04b94705f318337d7ead9e6d17c019bd9b1f6b1b
   hermes-engine: 922ccd744f50d9bfde09e9677bf0f3b562ea5fb9
   libevent: 4049cae6c81cdb3654a443be001fb9bdceff7913
   OpenSSL-Universal: ebc357f1e6bc71fa463ccb2fe676756aff50e88c
-  RCT-Folly: 85766c3226c7ec638f05ad7cb3cf6a268d6c4241
```
markdlv, line 600: Wonder if we're battling on xcode version- will follow up in slack.

lib/i18n/index.ts
```diff
@@ -9,8 +9,9 @@ declare module 'i18next' {
 }
 
 // Skeleton instance will enable building i18n ready components
-function configure() {
+export function configure() {
   return i18next.use(initReactI18next).init({
+    compatibilityJSON: 'v3',
```
markdlv, line 14: Without this, I got [errors about requiring a plural polyfill](https://github.com/i18next/i18next/issues/1671).  Let's hold off on that (polyfills) for now.

package.json
```diff
@@ -28,10 +28,12 @@
     "@types/react-native": "^0.71.0",
     "conventional-changelog-conventionalcommits": "^5.0.0",
     "eslint": "^8.33.0",
+    "i18next": "^22.4.9",
     "jest": "^29.4.1",
     "metro-react-native-babel-preset": "^0.74.1",
     "prettier": "^2.8.3",
     "react": "^18.2.0",
+    "react-i18next": "^12.1.5",
```
markdlv, line 36: All peerDependencies also need to be added as devDependencies in order to work locally.

bruce-glazier (approved):

src/hooks/useOAuthFlow.tsx
```diff
@@ -59,7 +59,9 @@ export const OAuthContextProvider = ({
       const { onSuccess, onFail } = params;
       if (!isLoggedIn || !authResult?.refreshToken) {
         await clearAuthResult();
-        onSuccess();
+        if (onSuccess) {
+          onSuccess();
+        }
```
sternetj, line 64: Bit of a nit but I would prefer `onSuccess?.()` since we don't care if execution succeeds and it reduces the vertical space and branching flow when reading.

sternetj (approved):

src/hooks/useOAuthFlow.tsx
markdlv: Love it.  I want to option-alize all of these `on*` methods - I'll get that in an all-in-one PR here in a sec

PR #26, opened 2023-02-10
(Over) simplified api clients hook, to start

markdlv:
Hopefully obvious I'm not trying to add all the features (auth refresh) or edge cases (401 error handling) here.  For the components that we all agree we don't want to migrate as-is (hopefully not TOO many of these...), I think it makes sense to introduce very small bits like this.  i.e. Auth refresh & 401 error handling can be separate PRs.

I manually tested this out like so, but couldn't find a way to put it into a story yet - that'll also come in subsequent PRs.
```
const { httpClient } = useAPIClients();

  useEffect(() => {
    httpClient.get('/v1/accounts').then(response => {
      console.warn('accounts', response.data);
    });
  }, []);
```

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -10,6 +10,7 @@
     "build-storybook": "build-storybook"
   },
   "dependencies": {
+    "axios": "^1.3.2",
     "react": "18.2.0",
     "react-native": "0.71.1",
     "react-native-app-auth": "^6.4.3",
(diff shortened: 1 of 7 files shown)


sternetj (approved):

sternetj (approved):

bruce-glazier (approved):

loscm:
Requested review from a random `@lifeomic/security` team member, @indigocarmen, due to term: **auth**.

PR #30, opened 2023-02-10
fix: Add ProfileImage component

bruce-glazier:
## Changes

  - Port over the ProfileImage component

## Screenshots

image


src/common/utils/colors.ts
```diff
@@ -0,0 +1,19 @@
...
+  highlight: '#F2F2F6',
+  white: '#FFFFFF',
+
+  announcement: {
+    gradientOrangeMedium: '#F2994A',
+    gradientOrangeLight: '#F2C94C',
+  },
+} as const;
+
+export default colors;
```
markdlv, line 19: I've been trying to avoid this file in hopes that we can introduce it as part of the eventual `useBranding` PR.  Is it possible to remove this, or does the component look ridiculous without it?

bruce-glazier: I will give it a shot and see what it looks like.

src/components/ProfileImage.tsx
```diff
@@ -0,0 +1,162 @@
...
+      displayName: string | null;
+      picture?: string | null;
+    };
+  };
+  customBackgroundStyle?: StyleProp<ViewStyle>;
+  imageUri?: string;
+}
+
+export const imageWrapperStyles: StyleProp<ViewStyle> = {
+  backgroundColor: '#02BFF1', // light blue
```
bruce-glazier, line 31: Perhaps we can default to this color when useBranding is introduced and the specified color is not defined in the branding, i.e, `brand.styles.image.borderColor ?? '#02BFF1',`

markdlv:
> This seems pretty good but it does make me think we should stop and consider where we want to go with style support.

Yep, great discussions on this yesterday.  Let's let this PR and the [gradient banner one](https://github.com/lifeomic/react-native-components/pull/31) hang out in PR branches until we have style/theme support in

bruce-glazier:
Closing for now until we revisit notifications.

PR #31, opened 2023-02-10
Add gradient banner component

bruce-glazier:
## Changes

  - Port over the SocialBanner component renamed to GradientBanner to be more generic

### DIFF
--- a/example/storybook/stories/GradientBanner.stories.tsx
+++ b/example/storybook/stories/GradientBanner.stories.tsx
@@ -1,0 +1,17 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react-native';
+import { text, color } from '@storybook/addon-knobs';
+import GradientBanner from 'src/components/GradientBanner';
+
+storiesOf('SocialBanner', module).add('demo', () => {
+  const gradientColor1 = color('gradientColor1', 'red');
+  const gradientColor2 = color('gradientColor2', 'orange');
+  const message = text('message', 'Today is national pie day.');
+
+  return (
+    <GradientBanner
+      gradientColors={[gradientColor1, gradientColor2]}
+      text={message}
+    />
+  );
+});
--- a/src/components/GradientBanner.tsx
+++ b/src/components/GradientBanner.tsx
@@ -1,0 +1,87 @@
+import React from 'react';
+import { StyleSheet, View, Text, Dimensions } from 'react-native';
+import {
+  Svg,
+  Path,
+  Defs,
+  LinearGradient,
+  Stop,
+  StopProps,
+  PathProps,
+} from 'react-native-svg';
+import times from 'lodash/times';
+
+const { width: windowWidth } = Dimensions.get('window');
+const styles = StyleSheet.create({
+  viewContainer: {
+    justifyContent: 'center',
+    alignContent: 'center',
+    alignItems: 'center',
+    marginVertical: 9,
+  },
+  labelText: {
+    color: '#FFFFFF',
+    marginBottom: 4,
+    fontSize: 10,
+    textTransform: 'uppercase',
+    fontWeight: '700',
+  },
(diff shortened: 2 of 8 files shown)


src/components/GradientBanner.tsx
```diff
@@ -0,0 +1,87 @@
...
+curvesSequenceTimes.forEach(() => {
+  const pointOne = initialPoint + axisXdistance;
+  const pointTwo = pointOne + axisXdistance;
+  const pointThree = pointTwo + axisXdistance;
+  const drawingSection = `28.8198 ${initialPoint} 28.8198C${pointOne} 28.8198 ${pointOne} 33.5847 ${pointTwo} 33.5847C${pointThree} 33.5847 ${pointThree}`;
+  curvesSectionsAtWidth.push(drawingSection);
+  initialPoint = pointThree + axisXdistance;
+});
+const drawingSectionString = curvesSectionsAtWidth.join(',').replace(',', ' ');
+const path: PathProps['d'] = `M${windowWidth} 0.44751H0.666992V30.011C2.2553 29.2963 4.08186 ${drawingSectionString} 28.8198 ${initialPoint} 28.8198V0.44751H${windowWidth}Z`;
```
namick, line 47: Interesting to see all these hardcoded values.  Where are these coming from?  Is this really how it is done?

bruce-glazier: I lifted this component as is so I'm not sure what the original author used to generated those values but it seems to work as expected.

example/storybook/stories/GradientBanner.stories.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react-native';
+import { text, color } from '@storybook/addon-knobs';
+import GradientBanner from 'src/components/GradientBanner';
```
sternetj, line 4: We would want this to be exported from src

src/components/GradientBanner.tsx
```diff
@@ -0,0 +1,87 @@
...
+const curvesSequenceTimes = times(windowWidth / (axisXdistance * 3));
+curvesSequenceTimes.forEach(() => {
+  const pointOne = initialPoint + axisXdistance;
+  const pointTwo = pointOne + axisXdistance;
+  const pointThree = pointTwo + axisXdistance;
+  const drawingSection = `28.8198 ${initialPoint} 28.8198C${pointOne} 28.8198 ${pointOne} 33.5847 ${pointTwo} 33.5847C${pointThree} 33.5847 ${pointThree}`;
+  curvesSectionsAtWidth.push(drawingSection);
+  initialPoint = pointThree + axisXdistance;
+});
+const drawingSectionString = curvesSectionsAtWidth.join(',').replace(',', ' ');
```
sternetj, line 46: This seems odd, why not just do a `join(' ')`?

src/components/GradientBanner.tsx
```diff
@@ -0,0 +1,87 @@
...
+const path: PathProps['d'] = `M${windowWidth} 0.44751H0.666992V30.011C2.2553 29.2963 4.08186 ${drawingSectionString} 28.8198 ${initialPoint} 28.8198V0.44751H${windowWidth}Z`;
+
+const GradientBanner = ({
+  gradientColors,
+  text,
+}: {
+  gradientColors: Array<StopProps['stopColor']>;
+  text: string | null;
+}) => {
+  const height = 34;
```
sternetj, line 56: This should probably be a prop and its default.

sternetj (commented):
I wonder if this is something we will want with the new brand styles we are moving towards. I feel like maybe we could ship the circle announcement notification without this and add it in once we work with the design team on what it should look like and if they want to still use this style. Might be a good question for Jay.

bruce-glazier:
Closing for now but may re-open in the future.

PR #32, opened 2023-02-13
Two hooks for accounts, and show account name on Settings screen

markdlv:
## Changes
- Add react-query
- useAccounts, useActiveAccount
- Settings screen shows account name

## Screenshots
Simulator Screen Shot - iPhone 14 Pro - 2023-02-13 at 10 48 59

### DIFF
--- a/example/storybook/stories/AppDemo.stories.tsx
+++ b/example/storybook/stories/AppDemo.stories.tsx
@@ -1,26 +1,39 @@
 import React, { FC } from 'react';
 import { Text } from 'react-native';
 import { storiesOf } from '@storybook/react-native';
-
 import { ActivityIndicator } from 'react-native';
 import { NavigationContainer } from '@react-navigation/native';
 import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import { QueryClient, QueryClientProvider } from 'react-query';
 import { action } from '@storybook/addon-actions';
+import {
+  ActiveAccountContextProvider,
+  APIClientsContextProvider,
+  AuthContextProvider,
+  useAuth,
+  OAuthLoginButton,
+  OAuthContextProvider,
+  SettingsScreen,
+} from '../../../src';
 import { authConfig } from './OAuth.stories';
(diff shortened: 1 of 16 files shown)


example/storybook/stories/AppDemo.stories.tsx
```diff
@@ -1,29 +1,39 @@
...
 import {
+  ActiveAccountContextProvider,
+  APIClientsContextProvider,
   AuthContextProvider,
   useAuth,
   OAuthLoginButton,
   OAuthContextProvider,
   SettingsScreen,
-} from '../../..';
+} from '../../../src';
```
markdlv, line 17: @bruce-glazier I'm finding this to be really useful.  Like we talked about offline, requiring them from an index file like this helps us avoid accidentally forgetting to export in a way that the npm lib will be able to consume.  I'm still in support of the `src/` & `lib/` paths - but just wanted to note we'll want this kind of import in some areas as well.

example/storybook/stories/AppDemo.stories.tsx
```diff
@@ -1,29 +1,39 @@
...
+          <OAuthContextProvider authConfig={authConfig}>
+            <ActiveAccountContextProvider>
+              <NavigationContainer>
+                <RootStack />
+              </NavigationContainer>
+            </ActiveAccountContextProvider>
+          </OAuthContextProvider>
+        </APIClientsContextProvider>
+      </AuthContextProvider>
+    </QueryClientProvider>
```
markdlv, line 36: This isn't going to work as a story much longer at all.  You run into issues with SafeAreaView (storybook defines its own), etc.  My thought is we'll want the `example` app to run one of two apps: a components-dev app (storybook) or a `demo` app that is an example usage of all the components in a real app setting.  An env var and/or different package.json scripts can facilitate which one to run.

bruce-glazier (dismissed):

bruce-glazier (approved):

sternetj (approved):

PR #37, opened 2023-02-16
Add GraphQL Client Provider

bruce-glazier:
## Changes

  - Adds GraphQL Client Provider leveraging `graphql-request` and `graphql` libraries
  - Adds nock to test suite to mock responses to graphql requests


loscm:
Requested review from a random `@lifeomic/security` team member, @securitybites, due to term: **access**.

markdlv:
> `Account does not have access to oauth client`

I know what this is - I'll follow up in slack 👍

src/common/testHelpers/mockGraphQLResponse.ts
```diff
@@ -0,0 +1,26 @@
...
+// Enable debug for nock by setting the debug variable
+// i.e., DEBUG=nock.* yarn test
+
+const mockGraphQLResponse = (
+  url: string,
+  assertRequestHeaders?: Record<string, string | undefined>,
+  response: Object = {},
+  responseCode: number = 200,
+) => {
+  return nock(url)
```
bruce-glazier, line 12: This is super basic and intended to reduce copy/paste between graphql hook tests. We may want to extend it slightly but more complex cases should just use nock directly (assuming we agree to take this approach).

example/yarn.lock
```diff
@@ -8459,6 +8500,13 @@ node-dir@^0.1.17:
   dependencies:
     minimatch "^3.0.2"
 
+node-fetch@2.6.7:
```
markdlv, line 8503: I was pretty excited about the graphql & httpClient re-using the same axios instance as their fetch impl.  Not necessarily in this PR (in fact I think this is good as is), but how would you feel about experimenting with the [custom fetch](https://www.npmjs.com/package/graphql-request#using-a-custom-fetch-method) option graphql-request offers?  The reason the re-use excited me was to have a single place to a) attach the authorization header and also (potentially) b) a single place to detect 401s.

src/common/testHelpers/mockGraphQLResponse.ts
markdlv: I can dig how nock gets us to _fully_ test the lib at hand, instead of mocking things in such a way that you can miss some behavior that's worth testing.  I've seen nock usage get hairy & hard to maintain, and I've seen requests accidentally not mocked and they leak out from unit tests.  As long as we watch those 2 things really closely, I think this should work out great.

src/hooks/useNotifications.tsx
```diff
@@ -0,0 +1,81 @@
...
+                  id
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+`;
```
markdlv, line 44: 🔥 perfect

markdlv:
Sorry about all the conflicts - once those are resolved, I think this is looking great.

One thing I'm thinking about, don't want to burden you with this on notifications screen but just plant the seed mentally: we don't want too much going into the _screen_ functions.  i.e. It'd be great if there was a notifications _component_, storybook'd etc., that the screen used.  That allows consumers to use the entire screen, drop down a level and use the component, or drop all the way down to just using the data/hook.  I'm going to try and take that approach with HomeScreen if/where it's easy

src/common/testHelpers/mockGraphQLResponse.ts
bruce-glazier: If we make the baseURL unique per test we should hopefully reduce any collisions. Definitely something to keep in mind.

example/yarn.lock
bruce-glazier: That seems like really solid reasoning to me. I'll see if I can sneak that in on another PR.

sternetj (approved):

namick (approved):

markdlv (approved):

sternetj (approved):

PR #38, opened 2023-02-16
chore: Providers, RootStack, dummy LoginScreen

markdlv:
## Changes
- Define simple providers & stack helpers
- Reuse those so that AppDemo.tsx looks more like what we'll want in react-native-starter

### DIFF
--- a/src/common/LoggedInProviders.tsx
+++ b/src/common/LoggedInProviders.tsx
@@ -1,0 +1,13 @@
+import React from 'react';
+
+import { ActiveAccountContextProvider } from '../hooks/useActiveAccount';
+
+export const LoggedInProviders = ({
+  children,
+}: {
+  children?: React.ReactNode;
+}) => {
+  return (
+    <ActiveAccountContextProvider>{children}</ActiveAccountContextProvider>
+  );
+};
--- a/example/AppDemo.tsx
+++ b/example/AppDemo.tsx
@@ -1,71 +1,13 @@
 import React, { FC } from 'react';
-import { ActivityIndicator, Text } from 'react-native';
 import { authConfig } from './storybook/stories/OAuth.stories';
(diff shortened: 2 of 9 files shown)


loscm:
Requested review from a random `@lifeomic/security` team member, @bishopb, due to term: **root**.

src/common/LoggedInProviders.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import { ActiveAccountContextProvider } from '../hooks/useActiveAccount';
+
+export const LoggedInProviders = ({
```
sternetj, line 5: I'm not sure about this name but it seems fine for now. To me this feels like the one provider that any consumer will have to import to be able to use our services. Something like `LifeomicPlatformProvider` might be appropriate but I am good to kick this down the road.

sternetj (approved):

bruce-glazier (approved):

src/common/LoggedInProviders.tsx
markdlv: I know what you're saying, but somehow I'd love to convey that anyone is welcome to (and encouraged) re-implement these provider lists.  I could almost see RootProviders & LoggedInProviders being prefixed with `Default` or something.  `LifeomcPlatformProvider` might accidentally convey that it _must_ be used.  Idk- maybe /** */ docs above these are going to be what take this from an blunt import name to a well-explained thing.

markdlv: Also worth mentioning, I think the reason we previously named this as such is because of how it's used:
```
if (isLoggedIn) {
    return (
      <LoggedInProviders>
         ...
```

Whereas if the user is not logged in, these are altogether not in the render/stack.

PR #39, opened 2023-02-16
chore: Support for account selection

markdlv:
## Changes
- Add AccountSelectionScreen
- Integrate account selection into SettingsScreen

## Screenshots
https://user-images.githubusercontent.com/1445395/219350196-83c09e5f-0854-4efe-a055-aa079e99fa69.mp4

### DIFF
--- a/src/hooks/useActiveAccount.tsx
+++ b/src/hooks/useActiveAccount.tsx
@@ -46,7 +46,8 @@ export const ActiveAccountContextProvider = ({
 }: {
   children?: React.ReactNode;
 }) => {
-  const accounts = useAccounts();
+  const accountsResult = useAccounts();
+  const [accountsWithProduct, setAccountsWithProduct] = useState<Account[]>([]);
   const [activeAccount, setActiveAccount] = useState<ActiveAccountProps>({});
 
   /**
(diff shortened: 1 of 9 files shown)


src/hooks/useActiveAccount.tsx
```diff
@@ -56,28 +57,28 @@ export const ActiveAccountContextProvider = ({
     if (activeAccount?.account?.id) {
       return;
     }
-    const accountsWithProduct = filterAccounts(accounts.data);
-    if (accountsWithProduct.length < 1) {
+    const filteredAccounts = filterAccounts(accountsResult.data);
```
sternetj, line 60: The logic in this hook looks very similar to the logic in `setActiveAccountId` I wonder if we could move the `useEffect` below `setActiveAccountId` and call the callback function instead.

src/hooks/useUserProfile.test.tsx
```diff
@@ -1,4 +1,3 @@
-// import React from 'react';
```
sternetj, line 1: Should we change the file extension?

sternetj (approved):

src/hooks/useUserProfile.test.tsx
markdlv: Looks like there are a few `.tsx` that could be `.ts` - I'll grab them all in a separate PR

PR #41, opened 2023-02-17
Style/theme/brand config with example

namick:
## Changes

  - Add a `useBrandConfig` hook to allow the customization of components in this library.  We will have a basic style and theme, but both can be customized by passing config into the provider:
  ```<BrandConfigProvider style={customStyle} theme={customTheme} />```
  - Add a `Text` component that shows this functionality

TODO:
  - Rename and finish the theme customization
  - Add a generic type for customizing the brand config
  - Add a storybook story
  - Documentation


src/hooks/useLifeOmicConfig.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { TextStyles, defaultTextStyles } from 'src';
+import * as theme from '../theme';
+
+export interface AppConfig {
```
markdlv, line 5: Can this be `BrandConfig` or something like that?

src/hooks/useLifeOmicConfig.tsx
```diff
@@ -0,0 +1,41 @@
...
+  };
+
+  return (
+    <LifeOmicConfigContext.Provider value={context}>
+      {children}
+    </LifeOmicConfigContext.Provider>
+  );
+}
+
+export const useLifeOmicConfig = () => React.useContext(LifeOmicConfigContext);
```
markdlv, line 41: Let's stick with `useAppConfig` when we get to that hook - but this seems different.  Can this be `useBranding`?

src/theme/colors.ts
```diff
@@ -0,0 +1,65 @@
+// TODO: Replace with our real default colors
```
markdlv, line 1: This looks great!  Great thoughts offline here.  It's important (to me) that this lib represents the "LifeOmic mobile app shell."  It should _not_ end up being the new Outcomes repo.  So ideally, these colors/spacing/etc. represent sensible defaults for us to develop components against.  And then we're overriding those in our Outcomes repo.

src/components/Text.test.tsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react-native';
+import { Text } from './Text';
+import { BrandConfig, BrandConfigProvider } from 'src/hooks/useBrandConfig';
+import { colors } from '../theme';
+
+describe('Without configuration', () => {
+  test('Renders text with default style (baseTheme.colors.text)', () => {
+    render(
+      <BrandConfigProvider>
```
sternetj, line 10: Shouldn't need to add the provider here. It might be a good idea to remove it so we can ensure that themed components can be used with the default context value.

sternetj (approved):

src/components/Text.test.tsx
namick: Oh, good call!

loscm:
Requested review from a random `@lifeomic/security` team member, @indigocarmen, due to term: **shell**.

src/components/Text.tsx
```diff
@@ -0,0 +1,38 @@
...
+  subHeading?: TextStyle;
+}
+
+type Variant = keyof TextStyles;
+
+export const defaultTextStyles: TextStyles = {
+  base: {
+    color: colors.text,
+  },
+};
```
markdlv, line 38: Something that might be worth prototyping, as an additive alternative to components using `useBrandConfig` like this, is an HOC that captures:
* useBrandConfig usage, but
* is injected with a mapping between expected brandConfig props & the component's props

In my mind, that'd allow some flexibility like:
- While some components might be coupled to `useBrandConfig`, all don't have to be.  i.e. There's a "component purity" factor here in my mind
- Using regular RN components (or some other component lib) instead of ours.  This particularly comes to mind for components that aren't unique to LifeOmic, like this `Text` example

src/index.ts
```diff
@@ -11,3 +11,4 @@ export * from './components';
 export * from './hooks';
 export * from './screens';
 export * from './navigators';
+export * from './common/SecureStore';
```
markdlv, line 14: I think you can back this change out.  `SecureStore` is already exported from `./common`

src/theme/colors.ts
markdlv: I'm curious whether this folder should `defaultTheme`, `baseTheme` - something to that effect?  Esp. if it's wireframe-y, I'd think we always want/expect something layered on top of the default values provided in these files.

src/hooks/useBrandConfig.tsx
```diff
@@ -0,0 +1,47 @@
...
+  },
+  theme,
+});
+
+interface Props {
+  styles?: BrandConfig['styles'];
+  children: React.ReactNode;
+}
+
+export function BrandConfigProvider({ styles, children }: Props) {
```
markdlv, line 25: So one can say `<BrandingConfigProvider styles={myStyleOverrides}>...` - very cool.  And eventually `theme` will also be injectable.  And those `styles` and `theme` props can be extensions of the types provided in this lib, with additional props etc.  Then if `useBrandConfig` had a generic, one could get back their own type.  Excited to see all this come together in storybook story/demo in some future PR

namick: Correct!  See `TODO:` list at top.  😄

src/hooks/useBrandConfig.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { StyleSheet } from 'react-native';
+import { TextStyles, defaultTextStyles } from 'src/components/Text';
+import * as theme from '../theme';
+
+export interface BrandConfig {
+  styles: {
+    Text?: TextStyles | undefined;
+  };
+  theme: {};
+}
```
markdlv, line 11: I'm eager to find/name strategies for ensuring this type doesn't balloon with props that we later want to refactor/remove as a breaking change.  The only I can think of are:
- Generally keep this type super simple and be cautious about adding to it (maybe need a comment)
- I wonder if we could separate out different groups into different types that could be (optionally) decorated/merged into a generic type used by this provider/context/hook?  Can explain further offline

markdlv (approved):
Lot of thoughts/ideas I commented in here, but removing `export * from './common/SecureStore';` is the only important one for this PR before merging.

src/hooks/useBrandConfig.tsx
namick: Agree 💯 . The idea here (in contrast to some ways we have done this in the past), would be that `styles:` property is composed from collections of styles imported from each component.  

All the style typing information should live with each component and be collected together here.  Might even be good to create a helper to build this up in the future if it gets to be too big here.

markdlv (approved):

PR #44, opened 2023-02-17
chore: useActiveProject and useSubjectProjects

markdlv:


src/hooks/useSubjectProjects.tsx
```diff
@@ -0,0 +1,33 @@
...
+    `${account?.id}-projects`,
+    () =>
+      httpClient
+        .get<ProjectsResponse>(
+          `/v1/projects?id=${subjects?.map((s) => s.projectId).join(',')}`,
+          { headers: accountHeaders },
+        )
+        .then((res) => res.data.items),
+    {
+      enabled: !!accountHeaders && !!subjects?.[0]?.projectId,
```
sternetj, line 30: Should this be: 

```suggestion
      enabled: !!accountHeaders && !!subjects?.some(subject => subject.projectId),
```

sternetj (approved):

bruce-glazier (approved):

src/hooks/useSubjectProjects.tsx
markdlv: Yea good catch.  Well, since we're sending all `projectId`s in the URL, I suppose it should even be `.every(...)`.  In my mind, a record in here without `projectId` would "never" happen.  But this change is much better defensive wise - will fix up in upcoming PR

PR #46, opened 2023-02-17
chore: Checkpoint on show app tiles on HomeScreen

markdlv:
## Changes
I have some unit testing to catch up on here, WIP.

## Screenshots
https://user-images.githubusercontent.com/1445395/219818565-bf4df7f3-3ec2-4381-9345-a86c39a9133a.mov


example/package.json
```diff
@@ -17,15 +17,17 @@
     "react-native-config": "^1.5.0",
     "react-native-device-info": "^10.3.0",
     "react-native-keychain": "^8.1.1",
-    "react-query": "^3.39.3"
+    "react-native-webview": "^11.26.1",
+    "react-query": "^3.39.3",
+    "reactotron-react-native": "^5.0.3"
```
markdlv, line 22: Once I added this (for debugging a 403), it was going to be annoying to disentangle from the rest of this.  I figured few enough changes no one would mind.

package.json
```diff
@@ -66,6 +66,7 @@
     "react-query": "^3.39.3"
   },
   "dependencies": {
+    "react-native-webview": "^11.26.1",
```
markdlv, line 69: TODO: this should be a peer dep because it requires a pod install

namick (approved):

src/components/tiles/Tile.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const Tile = ({ onPress, id, title, children }: TileProps) => {
+  if (onPress) {
+    return (
+      <TouchableOpacity testID={tID(`tile-button-${id}`)} onPress={onPress}>
+        <ViewTile id={id} title={title} />
+        {children}
+      </TouchableOpacity>
+    );
+  }
```
sternetj, line 29: This can make styling difficult when we are using a `Tile` internally that doesn't provide a clickable behavior. I think it might be better to always use a `TouchableOpacity` component but set `disabled` to `true` if `onPress` is falsy.

sternetj (approved):

bruce-glazier (approved):

src/components/tiles/Tile.tsx
markdlv: Makes sense - love it.  Will pickup in upcoming PR

PR #53, opened 2023-02-23
PHC-4075 - theme and style updates

namick:
## Changes

  - Added injectable theme and style customizations
  - Ability to scale default styles for many components with typing 
  - Used classes to clean up the style merging and prevent ballooning of type information

Still a WIP

## Screenshots

https://user-images.githubusercontent.com/220893/220939426-4524076a-1703-4ba1-99e2-2f7b6f0f188a.mov


example/storybook/stories/BrandConfig/BrandConfig.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    const styles = new Styles({ theme });
+    styles.Text.mergeStyles(customTextStyles);
+
+    return (
+      <BrandConfigProvider styles={styles}>
+        <Examples content={content} />
+      </BrandConfigProvider>
+    );
+  });
```
namick, line 104: I'll add better documentation, but this is basically the customization process for the theme and any components that want individually styled.

example/storybook/stories/BrandConfig/BrandConfig.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+    });
+
+    const theme = new Theme(themeConfig);
+
+    return (
+      <BrandConfigProvider theme={theme}>
+        <Examples content={content} />
+      </BrandConfigProvider>
+    );
+  })
```
namick, line 49: The Theme object can be injected directly into the provider if no component customization is needed.

src/components/BrandConfigProvider/Styles.ts
```diff
@@ -0,0 +1,18 @@
...
+  Text: TextStyles;
+
+  private _theme: Theme;
+
+  constructor({ theme }: Props = {}) {
+    this._theme = theme || new Theme();
+
+    this.Text = new TextStyles(this._theme);
+  }
+}
```
namick, line 18: This is where we can add style for any other components.  They will get a reference to the them injected into them.

src/themes/default/spacing.ts
```diff
@@ -8,6 +8,6 @@ export const spacing = {
   extraLarge: 32,
```
markdlv, line 8: The nesting here with `src/themes/default` might accidentally imply we'd have other folders inside the `themes` folder within the SDK.  But we don't intend on doing that atm, right?  I don't have a better suggestion and dig this for now, just noting the potential confusion.

namick: Correct, maybe better to just use `src/defaultTheme/`... However, thinking about now, do we need this at the root of `src`. Since it isn't meant to imported directly, but instead accessed through `useBrandConfig` then maybe we just put it with the other branding files.. next to the `Theme` class which is it's only direct consumer.

example/storybook/stories/BrandConfig/BrandConfig.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+      heading: {
+        color: 'purple',
+      },
+    });
+
+    const styles = new Styles({ theme });
+    styles.Text.mergeStyles(customTextStyles);
+
+    return (
+      <BrandConfigProvider styles={styles}>
```
markdlv, line 100: This is interesting- skimming through I expected `<BrandConfigProvider theme={theme} styles={styles}>` here, but see above how theme is injected into the Styles ctor.  I'll probably know better just reading on, but can you speak to the advantage of that over keeping them as separate params?

example/storybook/stories/BrandConfig/Examples.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+
+import { Text as RNText, TextStyle } from 'react-native';
+import { Text } from '../../../../src/components/Text';
+import { View, ViewStyle } from 'react-native';
```
markdlv, line 5: Any reason not to combine this `react-native` import with the one on line 3?

example/storybook/stories/Text.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+const style: ViewStyle = {
+  flex: 1,
+  justifyContent: 'center',
+  alignItems: 'center',
+};
+
+storiesOf('Text', module)
+  .addDecorator((story) => <BrandConfigProvider>{story()}</BrandConfigProvider>)
+  .addDecorator((story) => <View style={style}>{story()}</View>)
+  .add('default', () => <Text>Example textual content.</Text>, {
```
markdlv, line 17: Any reason not to throw `style` and `variant` props in via story knobs here?

example/storybook/stories/BrandConfig/Examples.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <>
+      <Container description="heading">
+        <Text variant="heading">{content}</Text>
+      </Container>
+      <Container description="subHeading">
+        <Text variant="subHeading">{content}</Text>
+      </Container>
+      <Container description="body">
+        <Text variant="body">{content}</Text>
+      </Container>
```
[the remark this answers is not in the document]
namick: I was thinking we would probably just have a few in examples, only enough to exercise the features of BrandConfigProvider. Seems like each component should also have it's own set of stories.  Almost like visual unit tests

However, if we can get to some of the more advanced storybook features, we should be able to switch up themes and styles on a global level, so that all stories are affected. Maybe to test out light vs dark themes on all components.

src/components/BrandConfigProvider/Styles.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class Styles {
+  Text: TextStyles;
+
+  private _theme: Theme;
+
+  constructor({ theme }: Props = {}) {
+    this._theme = theme || new Theme();
+
+    this.Text = new TextStyles(this._theme);
```
markdlv, line 16: Right on - so all components' styles will need to be worked into this class.  And customers/we can subclass this within apps if we want to have more props (e.g. for components that are only in that app).  Let's be crazy careful about what goes in here - just want to avoid refactors that amount to breaking changes.  If we do that well, I think this is going to be a nice approach.

example/storybook/stories/BrandConfig/BrandConfig.stories.tsx
namick: It needs to be changed. Originally the idea was to be able to pass in either, the theme and/or the styles. However, the Style ended up being dependent on the Theme already being configured when it is constructed. 

Currently you can pass in either the theme or the style. But if you have a theme and a style they need to be put together first.  Not ideal.

I plan to fix it possibly by introducing a BrandConfigSetup object that can take both configs and instantiate the other two properly

src/components/Text/TextStyles.ts
```diff
@@ -0,0 +1,38 @@
...
+      },
+    };
+
+    this.mergeStyles(defaultStyles);
+  }
+
+  mergeStyles(styles: Partial<TextStyles>) {
+    this._variants.forEach((variant) => {
+      Object.assign(this[variant], styles[variant]);
+    });
```
markdlv, line 36: I'm trying to decide how I feel about each component needing its own `*Styles` class.  I especially don't know if I like each having to implement a `mergeStyles` like this.  I like how only the component can/should know how it can merge styles into its internal styles representation - it's the class & method I'm not sure about.  For example, if I extrapolate this out into 10s of components, a consumer might have to do something like this in order to brand the app:
```
... create theme
const styles = new Styles({ theme });
styles.Text.mergeStyles(customTextStyles);
styles.Component2.mergeStyles(component2StyleOverrides);
styles.Component3.mergeStyles(component3StyleOverrides);
styles.Component4.mergeStyles(component4StyleOverrides);
styles.Component5.mergeStyles(component5StyleOverrides);
styles.Component6.mergeStyles(component6StyleOverrides);
... 10-15 more

return (<BrandConfigProvider styles={styles}>...)
```

I wonder if we should get ahead of that and make `mergeStyles` more of an impl detail that consumers don't need to worry about?  Something as simple as moving that merge logic to the ctor might go a long way:
```
... create theme
const styleOverrides = {
  Text: { body: { color: 'red' } },
  Component2: ...
  Component3: ...
};
const styles = new Styles({ theme, styleOverrides });
return (<BrandConfigProvider styles={styles}>...)
```
Not necessarily less LOC, just less method calls in my mind.

markdlv (commented):
That storybook demo is coming along nicely!

src/components/Text/Text.test.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+
+import { render, screen } from '@testing-library/react-native';
+import { Text } from './Text';
+import { BrandConfigProvider } from '../BrandConfigProvider';
+import { colors } from '../../themes/default';
+import { Styles } from '../BrandConfigProvider/Styles';
+
+describe('Without configuration', () => {
+  test('Renders text with default style (baseTheme.colors.text)', () => {
```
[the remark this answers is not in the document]
namick: Yes, I like that better too!

src/themes/default/spacing.ts
markdlv: Oh, so it wouldn't have a folder name at all - just `colors.ts` and `spacing.ts` files as siblings to `Theme` - love it.  We could prefix like `baseColors.ts` too - I personally like that better but either way love the idea of removing the folder.

src/components/Text/TextStyles.ts
namick: I think we are thinking along the same lines here.  I'm sure we have not arrived on the final API from a customer perspective and I don't expect `.megeStyles()` to be part of that.  Just building with iteration and creating simple primitives.

My next PR will have more on this, but in short I'm thinking that we'll want to support a number of use cases from just tweaking a few colors, to redesigning every component to match a consumer brand.

src/components/BrandConfigProvider/Styles.ts
namick: Yes, I think that is this class's one job, to provide access to all the component styles, and thus will need a handle on them.  This could be inverted by having the component style objects register themselves with this class.  That would make it so no changes are needed here when a new component is added.

markdlv:
Great offline conversation on how to progress on this PR.  Apart from minor cleanup, my suggestion is to just back out the `style` stories that use `mergeStyles`, and remove the BrandConfigProvider's acceptance of the `styles` prop.  That'd make for a simpler starting point, just allowing `theme` injection (which is looking great as is).  Like we talked about, this would allow us to evolve the interface for globally affecting components' default styles separately.  That's not strictly needed atm, so it's nice to push that off this early PR.

src/components/BrandConfigProvider/Theme.ts
```diff
@@ -0,0 +1,24 @@
...
+  colors: defaultTheme.Colors = defaultTheme.colors;
+  spacing: defaultTheme.Spacing = defaultTheme.spacing;
+
+  constructor({ colors, spacing }: Props = {}) {
+    this.mergeColors(colors || {});
+    this.mergeSpacing(spacing || {});
+  }
+
+  mergeColors(colors: Partial<defaultTheme.Colors>) {
+    Object.assign(this.colors, colors);
```
bruce-glazier, line 18: Not sure if needed but may want to do a deep merge here?

PR #56, opened 2023-02-24
fix: Refresh auth token on 401 http errors

markdlv:
## Changes
- Allow for API consumers to report 401s to useAuth
- Have useHttpClient report 401s to useAuth

### Refresh retries:
https://user-images.githubusercontent.com/1445395/221244825-8ec42efe-abc5-487f-86e2-eb70a950d0a3.mov

### Refresh fail:
https://user-images.githubusercontent.com/1445395/221244903-d365095a-569d-4618-afb3-ece98251f373.mov

☝️ waiting for future PR on `Alert` (user error message dialog) usage, error reporting, etc.

### DIFF
--- a/src/hooks/useAuth.tsx
+++ b/src/hooks/useAuth.tsx
@@ -16,6 +16,7 @@ export interface AuthStatus {
   storeAuthResult: (authResult: AuthResult) => Promise<void>;
   clearAuthResult: () => Promise<void>;
   initialize: (refreshHandler: RefreshHandler) => Promise<void>;
+  refreshForAuthFailure: (error: Error) => Promise<void>;
 }
 
 export type AuthResult = Omit<RefreshResult, 'additionalParameters'>;
@@ -27,6 +28,7 @@ const AuthContext = createContext<AuthStatus>({
   storeAuthResult: (_) => Promise.reject(),
   clearAuthResult: () => Promise.reject(),
   initialize: (_) => Promise.reject(),
+  refreshForAuthFailure: (_) => Promise.reject(),
 });
 
 const secureStorage = new SecureStore<AuthResult>('auth-hook');
@@ -78,6 +80,9 @@ export const AuthContextProvider = ({
 
   const refreshAuthResult = useCallback(
     async (_refreshHandler: RefreshHandler, _authResult: AuthResult) => {
+      if (__DEV__) {
+        console.warn('Attempting access token refresh');
+      }
       try {
         setLoading(true);
         const refreshResult = await _refreshHandler(_authResult);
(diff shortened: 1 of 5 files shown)


src/navigators/RootStack.tsx
```diff
@@ -19,7 +19,7 @@ export type NotLoggedInRootParamList = {
 export function RootStack() {
   const { isLoggedIn, loading } = useAuth();
 
-  if (loading) {
+  if (!isLoggedIn && loading) {
```
markdlv, line 22: This might evolve, but for now it provides the experience we're used to.  Without this change, any token refresh would bring the user back to the `Waiting for authorization` full-screen (below) and then home screen afterwards (if the refresh succeeded).  That might be desired functionality for some apps, so I could see this becoming configurable someday.  The approach in this PR means components/screens may still be on screen after errors (even 401s) and need to be prepared for error display and/or refresh controls to accommodate that scenario.  Again this is what we're used to so that can be the default initially.

loscm:
Requested review from a random `@lifeomic/security` team member, @bishopb, due to term: **auth**.

src/hooks/useAuth.tsx
```diff
@@ -118,30 +123,39 @@ export const AuthContextProvider = ({
...
+    [loading, refreshHandler, authResult, refreshAuthResult],
+  );
 
   useEffect(() => {
     if (currentAppState === 'active') {
       refreshIfNeeded();
     }
   }, [currentAppState, refreshIfNeeded]);
 
+  const refreshForAuthFailure = useCallback(async () => {
```
sternetj, line 147: I feel like we might want a different name here (something like `refreshAuth` or `refresh`), since the current name does not convey that it will always refresh the token because it sets `skipExpirationCheck` to true and because it suggests how/when it should be used.

sternetj (approved):

src/hooks/useAuth.tsx
markdlv: Yea I wrestled with the naming of this method.  What I want this to be is is combo of "report auth failure" and "refresh token if not already refreshing - and if refreshing fails I expect isLoggedIn to change".  I started with the former, `reportAuthFailure` because I don't want to indicate consumers have more power than they do.  Note that `skipExpirationCheck=true` does NOT mean it will always refresh; namely if `loading=true` the invocation will be essentially ignored.  Lmk if you or others still feel a more straight forward method name is better here - if we go that route, I think we should just expose `refreshIfNeeded`.  I personally still like this extra method that abstracts the internals a bit.  I could see this method evolving on its own, e.g. with debouncing etc.

src/hooks/useHttpClient.tsx
```diff
@@ -45,16 +46,38 @@ export const HttpClientContextProvider = ({
...
+    responseInterceptorId = axiosInstance.interceptors.response.use(
+      undefined,
+      async function (error: Error) {
+        if (axios.isAxiosError(error)) {
+          if (__DEV__) {
+            console.warn('Request Failed: ', error.toJSON());
+          }
+
+          if (error.response?.status === 401) {
+            await refreshForAuthFailure(error);
```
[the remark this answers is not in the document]
markdlv: When `refreshForAuthFailure` starts a refresh, it sets `loading` to true and if `refreshForAuthFailure` is called again while `loading=true` it'll be ignored.  I think the answer is still yes, though- more requests could come in before loading is set to true,and we could also simply have multiple refreshes one after the other succeed if they were timed just right/wrong (which would be wasteful).  My thought here is we can add debouncing/throttling to `refreshForAuthFailure` if/when we run into this - but to start simple in this PR.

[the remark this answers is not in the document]
bruce-glazier: Cool, thanks for the explanation. 👍

bruce-glazier (approved):

PR #57, opened 2023-02-24
PHC-4075 - Add Custom Theme Support

namick:
## Changes

- [x] Add theme readme
- [x] Add ability to customize the base theme

## Screenshots

Screenshot 2023-02-28 at 6 24 20 AM

## Stories

https://user-images.githubusercontent.com/220893/221843085-f5762034-56ea-4157-a430-31e58361ec4e.mov


sternetj (dismissed):

bruce-glazier (dismissed):

namick:
This is ready for review now.

I didn't use a custom `Text` component as an example in this PR.  Instead I just created an example screen with react-native `Text` and `View` components within the story itself to keep things more focused on just the Theme feature.

I also updated the colors to give us a wireframe looking `base` theme with blue-gray primary and secondary color scheme.  We can update it or add colors as needed when building out components.

docs/theme.md
```diff
@@ -0,0 +1,49 @@
+# How to Customize the Theme
+
+The components in this library come with a default set of colors and spacing.
+You can customize these properties (and add your own) by injecting a `Theme`
```
markdlv, line 4: Nit- `a[n] object`

src/components/BrandConfigProvider/BrandConfigProvider.tsx
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  const context = {
+    theme,
+  };
+
+  return (
+    <BrandConfigContext.Provider value={context}>
+      <ThemeProvider theme={theme}>{children}</ThemeProvider>
+    </BrandConfigContext.Provider>
```
markdlv, line 26: I like it!  One can not use the `BrandConfigProvider` and use ThemeProvider (and any other providers encapsulated in here in the future).  But the purpose of BrandConfigProvider is to keep all these things coordinated with each other- I think this is going to work out well.

src/components/BrandConfigProvider/theme/base/colors.ts
```diff
@@ -0,0 +1,32 @@
...
+  background: palette.neutral200,
+
+  primary: palette.blueGray800,
+  secondary: palette.blueGray200,
+
+  error: palette.error500,
+  errorBackground: palette.error100,
+};
+
+export type Colors = Partial<typeof colors>;
```
markdlv, line 32: Are you intentionally not exporting a `Palette` type here (as a sibling to Colors) to let it evolve a bit more before exporting?  Sounds good if so 👍

markdlv (dismissed):
Nice!

src/components/BrandConfigProvider/theme/base/colors.ts
namick: Yes, I tried to minimize everything. We currently have a semantic name for every color, so exposing palette is redundant. We can grow this API if we need in the future.

src/components/BrandConfigProvider/BrandConfigProvider.tsx
namick: Yes, much better!

src/components/BrandConfigProvider/theme/Theme.ts
```diff
@@ -0,0 +1,24 @@
+import * as baseTheme from './base';
+
+interface Props {
+  colors?: Partial<baseTheme.Colors>;
+  spacing?: Partial<baseTheme.Spacing>;
+}
+
+export class Theme {
+  colors: baseTheme.Colors = baseTheme.colors;
+  spacing: baseTheme.Spacing = baseTheme.spacing;
```
sternetj, line 10: I think you might want to spread the baseTheme to avoid `mergeX` mutating the baseTheme values

src/components/BrandConfigProvider/theme/base/colors.ts
sternetj: Do we want `Partial` on this? Shouldn't the type of `colors` always contain all fields?

src/components/BrandConfigProvider/theme/base/spacing.ts
```diff
@@ -8,6 +8,6 @@ export const spacing = {
   extraLarge: 32,
   huge: 48,
   massive: 64,
-} as const;
+};
 
-export type Spacing = keyof typeof spacing;
+export type Spacing = Partial<typeof spacing>;
```
sternetj, line 13: Same comment about `Partial` above.

src/components/BrandConfigProvider/theme/Theme.ts
namick: Thanks!

src/components/BrandConfigProvider/theme/base/colors.ts
namick: Yes, correct.  Fixed.

sternetj (approved):

markdlv (approved):

PR #59, opened 2023-02-24
README badges

markdlv:


hemp (dismissed):

package.json
```diff
@@ -13,6 +13,7 @@
     "lint:fix": "yarn lint --fix",
     "test": "yarn jest",
     "test:ci": "yarn lint && yarn jest --coverage --runInBand --silent --ci --logHeapUsage",
+    "coverage": "cat ./coverage/lcov.info | coveralls",
```
sternetj, line 16: What does this do exactly? Looking at the library it pushes it to coveralls.io but isn't that what the github action is for?

sternetj (approved):

namick (dismissed):

package.json
markdlv: 🤔 spot on.  I'm not even calling `yarn coverage` within this PR 🤦.  This is nice for testing/debugging, but it doesn't really need to be in package.json/yarn.lock - removing.

sternetj: Gotcha 👍, I was confused and thought you knew something I didn't because the action's own docs points you at that library and I thought maybe it called that command internally or something.

sternetj (approved):

package.json
markdlv: 👍 glad to turn around and add it back if the master merge fails, but pretty sure the Github Action is self-sufficient. We'll see!

PR #60, opened 2023-02-27
Add no internet toast message

bruce-glazier:
### Adds NoInternetToast
A toast message will be shown when there is no internet on the device and hidden when the connection is restored. A user can swipe the message off screen but it will reappear until internet connection is restored.

### Setup
- No additional setup required if importing/using RootProviders directly. Otherwise include the NoInternetToast as a sibling alongside the root NavigationContainer. Refer to the setup docs for react-native-toast-message [here](https://github.com/calintamas/react-native-toast-message/blob/main/docs/navigation-usage.md) for more info.

Additional setup and usage docs can be found on in `react-native-toast-message` repo [here](https://github.com/calintamas/react-native-toast-message).
      
## Next Up
- Jest test coverage

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "@babel/preset-env": "^7.20.2",
     "@lifeomic/eslint-plugin-i18next": "^2.0.1",
     "@react-native-community/eslint-config": "^3.2.0",
+    "@react-native-community/netinfo": "^9.3.7",
     "@react-navigation/bottom-tabs": "^6.5.5",
     "@react-navigation/native": "^6.1.3",
     "@react-navigation/native-stack": "^6.9.9",
@@ -51,6 +52,7 @@
     "react-native-keychain": "^8.1.1",
     "react-native-safe-area-context": "^4.5.0",
     "react-native-screens": "^3.19.0",
+    "react-native-toast-message": "^2.1.6",
     "react-native-webview": "^11.26.1",
     "react-query": "^3.39.3",
     "react-test-renderer": "^18.2.0",
@@ -59,6 +61,7 @@
     "typescript": "^4.9.4"
   },
   "peerDependencies": {
+    "@react-native-community/netinfo": "^9.3.7",
     "@react-navigation/bottom-tabs": "^6.5.5",
     "@react-navigation/native": "^6.1.3",
     "@react-navigation/native-stack": "^6.9.9",
@@ -74,6 +77,7 @@
     "react-native-keychain": "^8.1.1",
     "react-native-safe-area-context": "^4.5.0",
     "react-native-screens": "^3.19.0",
+    "react-native-toast-message": "^2.1.6",
     "react-native-webview": "^11.26.1",
     "react-query": "^3.39.3"
   },
--- a/src/common/RootProviders.tsx
+++ b/src/common/RootProviders.tsx
@@ -9,6 +9,8 @@ import { HttpClientContextProvider } from '../hooks/useHttpClient';
 import { OAuthContextProvider } from '../hooks/useOAuthFlow';
 import { ActiveProjectContextProvider } from '../hooks/useActiveProject';
 import { GraphQLClientContextProvider } from '../hooks/useGraphQLClient';
+import Toast from 'react-native-toast-message';
+import { NoInternetToastProvider } from '../hooks/NoInternetToastProvider';
 
 const queryClient = new QueryClient();
 
@@ -27,9 +29,12 @@ export function RootProviders({
             <OAuthContextProvider authConfig={authConfig}>
               <ActiveAccountContextProvider>
                 <ActiveProjectContextProvider>
-                  <SafeAreaProvider>
-                    <NavigationContainer>{children}</NavigationContainer>
-                  </SafeAreaProvider>
+                  <NoInternetToastProvider>
+                    <SafeAreaProvider>
+                      <NavigationContainer>{children}</NavigationContainer>
+                      <Toast />
+                    </SafeAreaProvider>
+                  </NoInternetToastProvider>
                 </ActiveProjectContextProvider>
               </ActiveAccountContextProvider>
             </OAuthContextProvider>
--- a/src/hooks/NoInternetToastProvider.tsx
+++ b/src/hooks/NoInternetToastProvider.tsx
@@ -1,0 +1,54 @@
+import React, { createContext, useEffect, useRef, useState } from 'react';
+import { useNetInfo } from '@react-native-community/netinfo';
+import { t } from 'i18next';
+import Toast, { ToastProps } from 'react-native-toast-message';
+
+export const NoInternetToastContext = createContext({});
+
+export const NoInternetToastProvider = ({
+  children,
+  toastOverrides,
+  forceShowToast,
+}: {
+  children?: React.ReactNode;
+  toastOverrides?: ToastProps;
+  forceShowToast?: boolean;
+}) => {
+  const { isConnected } = useNetInfo();
+  const hiddenByAPI = useRef<boolean>(false);
+  const [forceRefresh, setForceRefresh] = useState<boolean>(false);
+
+  useEffect(() => {
+    if (isConnected === false || forceShowToast === false) {
+      hiddenByAPI.current = false;
+      Toast.show({
+        type: 'error',
+        text1: t('connection-error-title', 'No internet connection ❌'),
+        text2: t(
+          'connection-error-body',
+          'This app requires an internet connection.',
+        ),
+        autoHide: false,
+        onHide: () => {
+          if (hiddenByAPI.current === false) {
+            // If dismissed by a user force a rerender
+            // so the toast is shown again
+            setForceRefresh((currentValue) => !currentValue);
+          } else {
+            hiddenByAPI.current = false;
+          }
+        },
+        ...toastOverrides,
+      });
+    } else if (isConnected === true || forceShowToast === true) {
+      hiddenByAPI.current = true;
+      Toast.hide();
+    }
+  }, [isConnected, toastOverrides, forceRefresh, forceShowToast]);
+
+  return (
+    <NoInternetToastContext.Provider value={{}}>
+      {children}
+    </NoInternetToastContext.Provider>
+  );
+};
--- a/src/hooks/NoInternetToastProvider.tsx
+++ b/src/hooks/NoInternetToastProvider.tsx
@@ -5,21 +5,24 @@ import Toast, { ToastProps } from 'react-native-toast-message';
 
 export const NoInternetToastContext = createContext({});
 
+class NoInternetToastProviderProps {
+  children?: React.ReactNode;
+  toastOverrides?: ToastProps;
+  /** Simulate the connection state (useful for testing) */
+  _isConnected?: boolean;
+}
+
 export const NoInternetToastProvider = ({
   children,
   toastOverrides,
-  forceShowToast,
-}: {
-  children?: React.ReactNode;
-  toastOverrides?: ToastProps;
-  forceShowToast?: boolean;
-}) => {
+  _isConnected,
+}: NoInternetToastProviderProps) => {
   const { isConnected } = useNetInfo();
   const hiddenByAPI = useRef<boolean>(false);
   const [forceRefresh, setForceRefresh] = useState<boolean>(false);
 
   useEffect(() => {
-    if (isConnected === false || forceShowToast === false) {
+    if (_isConnected === false || isConnected === false) {
       hiddenByAPI.current = false;
       Toast.show({
         type: 'error',
@@ -40,11 +43,11 @@ export const NoInternetToastProvider = ({
         },
         ...toastOverrides,
       });
-    } else if (isConnected === true || forceShowToast === true) {
+    } else if (_isConnected === true || isConnected === true) {
       hiddenByAPI.current = true;
       Toast.hide();
     }
-  }, [isConnected, toastOverrides, forceRefresh, forceShowToast]);
+  }, [isConnected, toastOverrides, forceRefresh, _isConnected]);
 
   return (
     <NoInternetToastContext.Provider value={{}}>
--- a/example/babel.config.js
+++ b/example/babel.config.js
@@ -22,6 +22,8 @@ module.exports = {
           '@react-navigation/bottom-tabs': path.resolve(__dirname, './node_modules/@react-navigation/bottom-tabs'),
           'react-native-safe-area-context': path.resolve(__dirname, './node_modules/react-native-safe-area-context'),
           'react-native-screens': path.resolve(__dirname, './node_modules/react-native-screens'),
+          'react-native-toast-message': path.resolve(__dirname, './node_modules/react-native-toast-message'),
+          '@react-native-community/netinfo': path.resolve(__dirname, './node_modules/@react-native-community/netinfo'),
           'graphql-request': path.resolve(
             __dirname, 
             './node_modules/graphql-request'
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -309,6 +309,8 @@ PODS:
     - react-native-config/App (= 1.5.0)
   - react-native-config/App (1.5.0):
     - React-Core
+  - react-native-netinfo (9.3.7):
+    - React-Core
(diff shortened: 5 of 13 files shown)


bruce-glazier:
Planning to add storybook and jest tests in follow up PRs.

src/hooks/useDetectNetworkConnection.tsx
```diff
@@ -0,0 +1,67 @@
...
+}: {
+  children?: React.ReactNode;
+  toastOverrides?: ToastProps;
+}) => {
+  const { isConnected } = useNetInfo();
+  const [forceCheck, setForceCheck] = useState<boolean>(false);
+  const toastRef = useRef(undefined);
+  const showToast = useCallback(
+    () =>
+      Toast.show(t('network-unstable', 'Your network is unstable.'), {
```
namick, line 31: Do we want to more accurately say something like, "Your Internet connection is unstable"?

This is a minor though, okay to keep it as is.  Probably just me who gets pedantic thinking, "what?  my network is perfectly stable,  it's my connection to the Internet that is a problem!" 😄

namick (dismissed):

src/hooks/useDetectNetworkConnection.tsx
bruce-glazier: I suppose the message is stating that your LAN is down. Most users will not make that distinction between network and internet however there still may be some! I will update it in a following PR.

src/common/RootProviders.tsx
```diff
@@ -1,3 +1,4 @@
+import { RootSiblingParent } from 'react-native-root-siblings';
```
markdlv, line 1: I see this lib in yarn.lock, but not package.json, which seems wrong.

Also, can you explain what this lib is doing in relation to the `DetectNetworkConnectionProvider`?

package.json
```diff
@@ -72,6 +75,7 @@
     "react-native-app-auth": ">=6.4.3",
     "react-native-device-info": "^10.3.0",
     "react-native-keychain": "^8.1.1",
+    "react-native-root-toast": "^3.4.1",
```
markdlv, line 78: Did you consider https://www.npmjs.com/package/react-native-toast-message vs. this lib?

src/common/RootProviders.tsx
bruce-glazier: > I see this lib in yarn.lock, but not package.json, which seems wrong.

`react-native-root-siblings` is a dependency of `react-native-root-toast` so it is installed when `react-native-root-toast` is installed.

> Also, can you explain what this lib is doing in relation to the DetectNetworkConnectionProvider?

RootSiblingParent is required to utilize 'root' components, i.e., the toast message which can be invoked via API anywhere. It is one of the first steps of setup for [react-native-root-toast](https://github.com/magicismight/react-native-root-toast).

As for why the `react-native-root-toast` library:
1. Mentioned in expo docs [here](https://docs.expo.dev/ui-programming/react-native-toast/).
2. The only toast library out of three that I tried that worked as expected without errors

package.json
bruce-glazier: I don't _think_ I tried this one. Maybe it is worth swapping out. I am getting some deprecated runtime warnings with the other library which I don't love.

src/hooks/useDetectNetworkConnection.tsx
```diff
@@ -0,0 +1,67 @@
...
+export const DetectNetworkConnectionProvider = ({
+  children,
+  toastOverrides,
+}: {
+  children?: React.ReactNode;
+  toastOverrides?: ToastProps;
+}) => {
+  const { isConnected } = useNetInfo();
+  const [forceCheck, setForceCheck] = useState<boolean>(false);
+  const toastRef = useRef(undefined);
```
markdlv, line 28: If I'm understanding the PR, this is combining the hook that's used if consumers want an `isConnected` prop to a user display (the toast).  I think ideally those should stay separate.  How would you feel about separating those two into a focused hook (possibly just migrating `useHasConnection`) and then a separate toast display component which utilizes its prop(s)?

src/hooks/useHttpClient.tsx
```diff
@@ -62,6 +64,7 @@ export const HttpClientContextProvider = ({
     responseInterceptorId = axiosInstance.interceptors.response.use(
       undefined,
       async function (error: Error) {
+        checkConnection?.();
```
sternetj, line 67: This seems a little concerning to me but it is probably okay for a first pass.

I view Toasts (and other forms of pop up messaging) as a design choice that sdk users should be able to decide on. By adding checkConnection to the http client, we are locking in sdk users in to having toasts in their app. The context provider is good because it seems like an sdk user could override the toast so it doesn't display, or make it look native to their app, but it doesn't give them a way to handle the scenario differently (splash page, logout, modal, status icon, etc.)

I feel like this might be a great place for optional dependencies that @markdlv mentioned at one point which would allow us to use toasts as the default behavior, if installed, or leave the handling up to the sdk user entirely.

sternetj (approved):

src/hooks/useDetectNetworkConnection.tsx
bruce-glazier: Since the toast is already its own component and has an API and netinfo already provides a hook for getting the connection state I am not sure what those hooks would provide. It wasn't my intention to use this like `useHasConnection` hook, i.e., that is a hook wrapper around an existing hook that doesn't really extend its functionality. In my mind developer's would just use netinfo directly for their connection related needs BUT I also exposed it here for a convenience but perhaps that just muddies the intention of this.

My intent was this would be a simple "detect network outage toast" with a way to force a recheck, not the go to hook for all network connection related matters. 

Does that all make sense? If so I think at least a rename is probably in order.

src/hooks/useHttpClient.tsx
bruce-glazier: 👍 Thanks for those thoughts, excellent points. I will look into this.

package.json
markdlv: 👍 the RootSiblingParent thing feels odd to me so I'm hopeful that other lib blows our minds.  It's published much more recently, has more weekly downloads, etc.  🤞

src/common/RootProviders.tsx
markdlv: If we stick to this lib, we can't/shouldn't rely on a dep of dep like that and will need to add `react-native-root-siblings` as a package.json entry / install.  As noted in another comment, I'm hopeful that other lib (react-native-toast-message) prototypes well, replaces this one in this PR, and then we won't need this.

src/hooks/useDetectNetworkConnection.tsx
markdlv: Ah, yes that totally makes sense.  I think you're spot on that the exposure of `checkConnection` and `isConnected` is what got us off on the wrong track.  I agree with @sternetj [here](https://github.com/lifeomic/react-native-sdk/pull/60/files#r1120169919) too so removing `checkConnection` exposure altogether would resolve that conversation as well.

markdlv:
Based on comments this is what I'm picking up can hopefully be fixed up in this PR:

- If prototyping goes well, replace react-native-root-toast with [react-native-toast-message](https://www.npmjs.com/package/react-native-toast-message), which eliminates the need for an explicit dep on react-native-root-siblings
- Rename etc. so this is a simpler component centered around a "not connected" toast
- Remove `checkConnection`, `isConnected`, and httpClient integration from this PR - to be handled later

src/hooks/useDetectNetworkConnection.tsx
sternetj: Yeah that definitely clears it up for me and makes perfect sense.

example/AppDemo.tsx
```diff
@@ -9,7 +9,9 @@ if (__DEV__) {
 const App: FC = () => {
   return (
     <RootProviders authConfig={authConfig}>
-      <RootStack />
+      <NoInternetToastProvider>
+        <RootStack />
+      </NoInternetToastProvider>
```
markdlv, line 14: Did you put this here (and not inside `RootProviders`) to give SDK users more of a choice about how to display no-internet message?  I'd like to keep this simpler in this first pass we're in, and put everything that is in the app shell inside RootProviders or RootStack.

src/hooks/useNoInternetToast.tsx
```diff
@@ -0,0 +1,49 @@
...
+      });
+    } else if (isConnected === true) {
+      Toast.hide();
+    }
+  }, [isConnected, forceRefresh, toastOverrides]);
+
+  return (
+    <NoInternetToastContext.Provider
+      value={{
+        refreshToast: () => setForceRefresh((currentVal) => !currentVal),
```
markdlv, line 40: You mentioned in the PR desc this will be used to put the toast back up upon user interaction, but can you explain more how that will work?  I wonder about instead experimenting with `visibilityTime` (e.g. 10 min), `autoHide=false`, `onPress` (e.g. can we cancel default behavior in there?), and maybe best of all `onHide` - maybe `onHide` could internally toggle setForceRefresh and then we wouldn't have to expose `refreshToast`?

example/AppDemo.tsx
bruce-glazier: Yeah, I chose not to include it by default to allow the app developer the choice but I will update it.

src/hooks/useNoInternetToast.tsx
bruce-glazier: Done!

example/AppDemo.tsx
bruce-glazier: Updated

src/common/RootProviders.tsx
```diff
@@ -29,6 +30,7 @@ export function RootProviders({
                 <ActiveProjectContextProvider>
                   <SafeAreaProvider>
                     <NavigationContainer>{children}</NavigationContainer>
+                    <NoInternetToast />
```
markdlv, line 33: I see - so, in the case that we want to utilize `<Toast />` for other purposes that are also global, they can be siblings to this.  👍 I dig the simplicity.  If we someday want a `<Toast />` here that tries to serve all "global toast" needs, we can always refactor into that later.

markdlv: Or no, shoot - reading through the [docs](https://github.com/calintamas/react-native-toast-message/blob/945189fec9746b79d8b5b450e298ef391f8022fb/docs/modal-usage.md) it sounds like the `<Toast />` element is really designed to be globally reused.  I get this impression esp. from that modal page's notes about `refs` etc.:

> The ref is tracked automatically; whichever <Toast /> instance last had its ref set will be used when showing/hiding.

The [react-native-model example](https://github.com/calintamas/react-native-toast-message/blob/945189fec9746b79d8b5b450e298ef391f8022fb/docs/modal-usage.md#notes-regarding-react-native-modal-or-nativestacknavigator) similarly makes me think we'd never want _sibling_ `<Toast />` elements (the internal ref tracking would cause only the last one to be used).  This makes me think this line really should just be `<Toast />`, right?  I better understand now what you were saying on how `NoInternetToast` isn't really a component - it's really just the `useEffect` that we need.  I'd personally turn to a context Provider that just renders its children for this.

bruce-glazier: Ahh, yes. Keeping the Toast and the NoInternetToast "effect" separate is a better path then. I've pushed another commit that I'm hopeful will be what we all want.

src/hooks/NoInternetToastProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+  children?: React.ReactNode;
+  toastOverrides?: ToastProps;
+  forceShowToast?: boolean;
+}) => {
+  const { isConnected } = useNetInfo();
+  const hiddenByAPI = useRef<boolean>(false);
+  const [forceRefresh, setForceRefresh] = useState<boolean>(false);
+
+  useEffect(() => {
+    if (isConnected === false || forceShowToast === false) {
```
markdlv, line 22: Should this be `&&` instead of `||` ?  That way, `forceShowToast` will always force execution into this block.

src/hooks/NoInternetToastProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+export const NoInternetToastContext = createContext({});
+
+export const NoInternetToastProvider = ({
+  children,
+  toastOverrides,
+  forceShowToast,
+}: {
+  children?: React.ReactNode;
+  toastOverrides?: ToastProps;
+  forceShowToast?: boolean;
```
markdlv, line 15: Yet another nit-pick, sorry I have so many on such a simple component: I don't love how we're exposing a property that's solely for testing purposes.  I know that's done sometimes, perhaps with `_` prefix, but I think it should be avoided whenever possible.  In this case, I wonder about just making the story show some text like "Turn off internet on your machine, or in your simulator, to see the No Internet toast".  It doesn't make for as smooth/intuitive of a story/demo as what you have, but between the story and not exposing extra props, I'd vote for the latter.  Alternatively, if we do take in a testing prop, I think just taking an injected `isConnected` override might be simpler.  I might be reading it wrong (lmk) but I feel like the `forceShowToast` checks get confusing in the useEffect.

markdlv (dismissed):
Approving in case I'm misreading the `forceShowToast` checks - overall looking good to me

src/hooks/NoInternetToastProvider.tsx
bruce-glazier: `isConnected` will always be true so if we make this an `&&` then we will never enter this block.

src/hooks/NoInternetToastProvider.tsx
bruce-glazier: I've been digging into alternatives but coming up empty. Unfortunately the connection state change events from `useNetInfo` do not work properly on an iOS simulator so we cannot rely on expecting folks to disable their connection to see this component. As far as I can tell our only options are to:

A) Remove the storybook example
B) Keep the extra prop but rename it and add JSDoc to explain it better, something like this:

image

src/hooks/NoInternetToastProvider.tsx
bruce-glazier: I rearrange the `_isConnected` to be the first evaluated to be sure the overwritten value is always honored first.

src/hooks/NoInternetToastProvider.tsx
namick: If we go for B, I would be a fan of a more annoyingly explicit name like `_isConnectedTestOverride` or `_simulatedIsConnectedState`

That would make this easier to understand:
```
if (_isConnected === false || isConnected === false) {
```
Without having to look elsewhere for context.

markdlv (approved):

src/hooks/NoInternetToastProvider.tsx
markdlv: > so we cannot rely on expecting folks to disable their connection to see this component

With a helpful storybook message, I'm not sure I agree with that.  But I feel like we've nit-picked this quite a bit and I think this PR is in great shape as-is.  By all means keep tweaking if desired, but also approving in case you want to merge & iterate in subsequent PRs.

namick (approved):
Also agree that this is great as is.

src/hooks/NoInternetToastProvider.tsx
bruce-glazier: > > so we cannot rely on expecting folks to disable their connection to see this component
> 
> With a helpful storybook message, I'm not sure I agree with that. But I feel like we've nit-picked this quite a bit and I think this PR is in great shape as-is. By all means keep tweaking if desired, but also approving in case you want to merge & iterate in subsequent PRs.

All I mean to say here is that this does not work on iOS simulators properly. It will show the message but never hide it. So I take that last statement back, we could instruct them accordingly. I'm going to merge this as-is since it has been growing and growing in history and circle back with some improvements.

bruce-glazier: > If we go for B, I would be a fan of a more annoyingly explicit name like `_isConnectedTestOverride` or `_simulatedIsConnectedState`
> 
> That would make this easier to understand:
> 
> ```
> if (_isConnected === false || isConnected === false) {
> ```
> 
> Without having to look elsewhere for context.

👍 Good advice, thanks. I may eliminate this prop in a follow up PR but if not, I will rename it.

PR #62, opened 2023-02-28
PHC-4075 - Add styles provider

namick:
## Changes

  -  [x] Initial styling system with typing
  -  [x] Add tests and stories
  -  [x] Clean up

## Screenshots

https://user-images.githubusercontent.com/220893/222241612-3a3a817b-1a6f-4b24-a223-eb38af620c10.mov

### DIFF
--- a/example/storybook/stories/BrandConfigProvider/BrandConfigProvider.stories.tsx
+++ b/example/storybook/stories/BrandConfigProvider/BrandConfigProvider.stories.tsx
@@ -1,12 +1,17 @@
 import React from 'react';
+
+import { View, ViewStyle } from 'react-native';
 import { storiesOf } from '@storybook/react-native';
+import { object, color } from '@storybook/addon-knobs';
 import { BrandConfigProvider, Theme } from 'src';
 import { ThemeExampleScreen } from './ThemeExampleScreen';
 import * as baseTheme from 'src/components/BrandConfigProvider/theme/base';
-
-import { object, color } from '@storybook/addon-knobs';
+import { ExampleBox, ExampleBoxStyles } from './ExampleBox';
+import { BrandConfigProviderStyles } from 'src/components/BrandConfigProvider/types';
 
 storiesOf('BrandConfigProvider', module)
+  .addDecorator((story) => <View style={centerView}>{story()}</View>)
+
   .add('Basic Theme Colors', () => {
     const customColors: Record<string, string> = {};
     Object.entries(baseTheme.colors).forEach(([key, value]) => {
@@ -21,6 +26,7 @@ storiesOf('BrandConfigProvider', module)
       </BrandConfigProvider>
     );
   })
+
   .add('Default Theme', () => {
     const customColors = object('theme.colors', baseTheme.colors);
     const customSpacing = object('theme.spacing', baseTheme.spacing);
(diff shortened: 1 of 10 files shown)


src/components/BrandConfigProvider/styles/StylesProvider.tsx
```diff
@@ -0,0 +1,53 @@
...
+export const useStyles = <T extends StylesBuilder>(
+  name: keyof ComponentStyles,
+  builder: T,
+  ...localStyles: PartialStylesProp<T>[]
+) => {
+  const { theme } = useTheme();
+  const { componentStyles } = React.useContext(StylesContext);
+
+  return React.useMemo(() => {
+    const styles = merge(
```
sternetj, line 26: Should we use `StyleSheet.merge` here?

src/components/BrandConfigProvider/styles/StylesProvider.tsx
```diff
@@ -0,0 +1,53 @@
...
+      componentStyles[name],
+      ...localStyles,
+    ) as StylesProp<T>;
+    return { styles };
+  }, [theme, builder, name, componentStyles, localStyles]);
+};
+
+interface Props {
+  componentStyles: RecursivePartial<ComponentStyles>;
+  theme: RecursivePartial<Theme>;
```
sternetj, line 38: This doesn't seem to be used in `StylesProvider`

sternetj (commented):
Looks really solid. I think there might be a small area of improvement we could make that would reduce some of the boilerplate even more. I think by creating our own wrapper for `StyleSheet.create` we should be able to define the component style name in a single place and adjust the types to grab that name and use it when we extend the `ComponentStyles` type.

src/components/BrandConfigProvider/styles/StylesProvider.tsx
namick: Good discussion offline.  Looks like `StyleSheet` doesn't give us anything.

example/storybook/stories/BrandConfigProvider/BrandConfigProvider.stories.tsx
```diff
@@ -32,4 +38,40 @@ storiesOf('BrandConfigProvider', module)
         <ThemeExampleScreen />
       </BrandConfigProvider>
     );
+  })
+
+  .add('Style', () => {
```
markdlv, line 43: This story is perfect for showing off what we're going for - nice!

example/storybook/stories/BrandConfigProvider/ExampleBox.tsx
```diff
@@ -0,0 +1,47 @@
...
+    text,
+  };
+};
+
+declare module 'src/components/BrandConfigProvider/types' {
+  interface ComponentStyles
+    extends ComponentNamedStyles<'ExampleBox', typeof defaultStyles> {}
+}
+
+export type ExampleBoxStyles = NamedStylesProp<typeof defaultStyles>;
```
markdlv, line 47: I really like how this file is in the `storybook` folder.  One can go too far and write too much code in this folder, but in this case it's a demonstration of how we'll write styled components.  In a way, this code is a "story" on the new `useStyles`.  Dig it

package.json
```diff
@@ -78,6 +79,7 @@
     "react-query": "^3.39.3"
   },
   "dependencies": {
+    "lodash": "^4.17.21",
```
markdlv, line 82: Nit-pick - if we have a chance to simplify down to `Object.assign` and bring this out later, I'd be in favor of that route.  Nbd in the meantime - nice that we're NOT adding as a peer dep here for now, since it might be temporary

markdlv (approved):
Nice!  LGTM

example/storybook/stories/BrandConfigProvider/ExampleBox.tsx
namick: Indeed.  It serves as a bit of executable documentation. (Not that we want to skip actual markdown documentation for the API which will be based off of this.)

PR #64, opened 2023-03-03
PHC-4075 - Add base style [OAuthLoginButton]

namick:
## Changes

  - Add base wireframe color scheme
  - Style the OAuthLoginButton
  - Add the DemoApp as a Storybook story

## Screenshots

https://user-images.githubusercontent.com/220893/222758453-5a3bcb32-e668-4e64-93c0-cde7795516c4.mov

### DIFF
--- a/example/storybook/stories/ExampleApp.stories.tsx
+++ b/example/storybook/stories/ExampleApp.stories.tsx
@@ -1,0 +1,10 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react-native';
+import { authConfig } from './OAuth.stories';
+import { RootProviders, RootStack } from 'src';
+
+storiesOf('Example App', module).add('DemoApp', () => (
+  <RootProviders authConfig={authConfig}>
+    <RootStack />
+  </RootProviders>
+));
--- a/src/components/BrandConfigProvider/theme/base/colors.ts
+++ b/src/components/BrandConfigProvider/theme/base/colors.ts
@@ -1,32 +1,61 @@
-const palette = {
-  neutral200: '#F4F2F1',
-  neutral300: '#D7CEC9',
-  neutral400: '#B6ACA6',
-  neutral600: '#564E4A',
-  neutral800: '#191015',
-
-  blueGray200: '#B0BEC5',
-  blueGray800: '#37474F',
-
-  error100: '#F2D6CD',
-  error500: '#C03403',
-};
-
+/**
+ * Base (wireFrame-like) color scheme
+ *
+ * The base color scheme uses a popular design system to
+ * dynamically generate all color tokens from the HTML color:
+ * "gray" - #808080	- rgb(128,128,128)
+ */
 export const colors = {
-  white: '#FFFFFF',
-  black: '#000000',
+  primary: 'rgb(0, 104, 116)',
+  onPrimary: 'rgb(255, 255, 255)',
+  primaryContainer: 'rgb(151, 240, 255)',
+  onPrimaryContainer: 'rgb(0, 31, 36)',
 
-  text: palette.neutral800,
-  textDim: palette.neutral600,
-  border: palette.neutral400,
-  separator: palette.neutral300,
-  background: palette.neutral200,
+  secondary: 'rgb(74, 98, 103)',
+  onSecondary: 'rgb(255, 255, 255)',
+  secondaryContainer: 'rgb(205, 231, 236)',
+  onSecondaryContainer: 'rgb(5, 31, 35)',
 
-  primary: palette.blueGray800,
-  secondary: palette.blueGray200,
+  tertiary: 'rgb(82, 94, 125)',
+  onTertiary: 'rgb(255, 255, 255)',
+  tertiaryContainer: 'rgb(218, 226, 255)',
+  onTertiaryContainer: 'rgb(14, 27, 55)',
 
-  error: palette.error500,
-  errorBackground: palette.error100,
+  error: 'rgb(186, 26, 26)',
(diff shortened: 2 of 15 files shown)


example/storybook/stories/ExampleApp.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react-native';
+import { authConfig } from './OAuth.stories';
+import { RootProviders, RootStack } from 'src';
+
+storiesOf('Example App', module).add('DemoApp', () => (
+  <RootProviders authConfig={authConfig}>
+    <RootStack />
+  </RootProviders>
+));
```
namick, line 10: Add the whole demo app as a storybook story.

src/components/BrandConfigProvider/theme/base/colors.ts
```diff
@@ -1,32 +1,118 @@
...
+    background: 'rgb(250, 253, 253)',
+    onBackground: 'rgb(25, 28, 29)',
+
+    surface: 'rgb(250, 253, 253)',
+    onSurface: 'rgb(25, 28, 29)',
+    surfaceVariant: 'rgb(219, 228, 230)',
+    onSurfaceVariant: 'rgb(63, 72, 74)',
+
+    outline: 'rgb(111, 121, 122)',
+    outlineVariant: 'rgb(191, 200, 202)',
```
namick, line 39: light

src/components/BrandConfigProvider/theme/base/colors.ts
```diff
@@ -1,32 +1,118 @@
...
+    background: 'rgb(25, 28, 29)',
+    onBackground: 'rgb(225, 227, 227)',
+
+    surface: 'rgb(25, 28, 29)',
+    onSurface: 'rgb(225, 227, 227)',
+    surfaceVariant: 'rgb(63, 72, 74)',
+    onSurfaceVariant: 'rgb(191, 200, 202)',
+
+    outline: 'rgb(137, 146, 148)',
+    outlineVariant: 'rgb(63, 72, 74)',
```
namick, line 92: dark

src/components/BrandConfigProvider/theme/base/colors.ts
namick: We should be able to refer to these colors visually by clicking back to this PR from a vscode context menu near this code.

src/components/OAuthLoginButton.tsx
```diff
@@ -24,12 +36,62 @@ export const OAuthLoginButton: FC<OAuthLoginButtonParams> = ({
...
+      web: 'Roboto, "Helvetica Neue", Helvetica, Arial, sans-serif',
+      ios: 'System',
+      default: 'sans-serif-medium',
+    }),
+
+    fontWeight: '500',
+    letterSpacing: 0.1,
+    lineHeight: 20,
+    fontSize: 14,
+  };
```
namick, line 87: Lots of this will be boilerplate for many components and should be moved to the theme.  A good example of how this can be done is found here:

https://github.com/callstack/react-native-paper/blob/main/src/styles/themes/v3/tokens.tsx

markdlv: 👍 Great offline discussion on RN paper too.  I'm all for considering adopting that to avoid having to write too much code for things like Button, Text, icons, what props should a theme have, etc.  Maybe their providers could be wrapped by ours if/where needed.  Let's maybe shoot for a 1 day prototype on this next week

markdlv (approved):
LGTM - thanks for making this incremental

src/components/OAuthLoginButton.tsx
```diff
@@ -24,12 +36,62 @@ export const OAuthLoginButton: FC<OAuthLoginButtonParams> = ({
...
+
+  return { button, touchableOpacity, content, label };
+};
+
+declare module './BrandConfigProvider/types' {
```
sternetj, line 92: I wonder if we could use TypeScript paths here to make this a bit easier. i.e. `declare module '@style' {`

src/components/OAuthLoginButton.tsx
```diff
@@ -24,12 +36,62 @@ export const OAuthLoginButton: FC<OAuthLoginButtonParams> = ({
...
+      default: 'sans-serif-medium',
+    }),
+
+    fontWeight: '500',
+    letterSpacing: 0.1,
+    lineHeight: 20,
+    fontSize: 14,
+  };
+
+  return { button, touchableOpacity, content, label };
```
sternetj, line 89: Instead of declaring variables for each named style we could do something like:

```
return {
  button: {
    // ...
  } as ViewStyle,
  label: {
    // ...
  } as TextStyle,
```

sternetj (approved):

src/components/OAuthLoginButton.tsx
namick: I spent some time trying to figure this out, but it sound like you already know the way!  Awesome, will update in next PR

src/components/OAuthLoginButton.tsx
namick: By casting, wouldn't we lose the type checking in the object itself?  My understanding was to not prefer casting.  For instance, this compiles just fine:
```
content = {
    flexDirection: 'row',
    alignItems: 'center',
    justifyContent: 'center',
    notValid: 'garbage',
  } as ViewStyle;
```

sternetj: Great point! We shouldn't do that then.

PR #65, opened 2023-03-03
PHC-4151 Cleanup TODOs in useActiveAccount

bruce-glazier:
## Changes

  - Allow passing in accountId to provider to force an account selection. This takes priority over what is in async storage and also still honors the list of accounts returned from `/accounts`, i.e., if the specified account is not present an error is thrown. 
  - Reads and writes the selected accountId from async storage so selections persist between sessions
  - Add react-query wrapped get async storage hook to simplify usage


bruce-glazier:
Still need to do manual testing on these changes

sternetj (approved):

namick (approved):

PR #68, opened 2023-03-03
PHC-4151 Update useActiveProject to read/write to async storage

bruce-glazier:
## Changes

  - Read/write the selectedProjectId to async storage
  - Contains https://github.com/lifeomic/react-native-sdk/pull/65 so review that one first

### DIFF
--- a/src/hooks/useActiveProject.test.tsx
+++ b/src/hooks/useActiveProject.test.tsx
@@ -1,11 +1,16 @@
 import React from 'react';
-import { act, renderHook } from '@testing-library/react-native';
+import { act, renderHook, waitFor } from '@testing-library/react-native';
 import { useSubjectProjects } from './useSubjectProjects';
 import { useMe } from './useMe';
 import {
   ActiveProjectContextProvider,
   useActiveProject,
 } from './useActiveProject';
+import * as useAsyncStorage from './useAsyncStorage';
+import { QueryClient, QueryClientProvider, UseQueryResult } from 'react-query';
+import AsyncStorage from '@react-native-async-storage/async-storage';
+import AsyncStorageMock from '@react-native-async-storage/async-storage/jest/async-storage-mock';
+import { mockDeep } from 'jest-mock-extended';
 
 jest.mock('./useSubjectProjects', () => ({
   useSubjectProjects: jest.fn(),
@@ -16,6 +21,7 @@ jest.mock('./useMe', () => ({
 
 const useSubjectProjectsMock = useSubjectProjects as jest.Mock;
 const useMeMock = useMe as jest.Mock;
+let useAsyncStorageSpy = jest.spyOn(useAsyncStorage, 'useAsyncStorage');
 
 const mockSubjectProjects = [
   {
@@ -41,7 +47,9 @@ const mockMe = [
 const renderHookInContext = async () => {
   return renderHook(() => useActiveProject(), {
     wrapper: ({ children }) => (
-      <ActiveProjectContextProvider>{children}</ActiveProjectContextProvider>
+      <QueryClientProvider client={new QueryClient()}>
+        <ActiveProjectContextProvider>{children}</ActiveProjectContextProvider>
+      </QueryClientProvider>
     ),
   });
 };
@@ -53,6 +61,13 @@ beforeEach(() => {
   useMeMock.mockReturnValue({
     data: mockMe,
   });
+
+  useAsyncStorageSpy.mockReturnValue([
+    {
+      ...mockDeep<UseQueryResult<string | null>>(),
+    },
+    (value: string) => AsyncStorage.setItem('selectedProjectIdKey', value),
+  ]);
 });
 
 test('without provider, methods fail', async () => {
@@ -93,6 +108,7 @@ test('exposes some props from useSubjectProjects and useMe', async () => {
     isFetched: true,
     error,
   });
+
   await rerender({});
 
   expect(result.current).toMatchObject({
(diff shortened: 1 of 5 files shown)


sternetj (dismissed):

src/hooks/useActiveProject.tsx
```diff
@@ -50,35 +59,54 @@ export const ActiveProjectContextProvider = ({
         return;
       }
 
-      // TODO: Save for previously-selected project
+      storeProjectId(selectedProject.id);
```
markdlv, line 62: We need to `await` this, I'd think?

src/hooks/useActiveProject.tsx
```diff
@@ -50,35 +59,54 @@ export const ActiveProjectContextProvider = ({
...
+        ),
+        isFetched: !!(
+          projectsResult.isFetched &&
+          useMeResult.isFetched &&
+          storedProjectIdResult.isFetched
+        ),
+        error:
+          projectsResult.error ||
+          useMeResult.error ||
+          storedProjectIdResult.error,
```
markdlv, line 109: I'm not sure we'd necessarily want to return the `storedProjectIdResult.error` here.  It feels like an error that could be recovered from (e.g. the user just needs to re-select the project) whereas the others are errors consumers will need to handle.

src/hooks/useActiveProject.tsx
```diff
@@ -36,6 +40,11 @@ export const ActiveProjectContextProvider = ({
   const [hookReturnValue, setHookReturnValue] = useState<ActiveProjectProps>(
     {},
   );
+  const storedProjectIdResult = useGetAsyncStorage(selectedProjectIdKey);
```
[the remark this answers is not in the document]
bruce-glazier: The motivation behind was to "hookify" the async storage to keep the useEffect cleaner without having to embed an async function everywhere we need to access it. The hope was that the code would be cleaner.

src/hooks/useActiveProject.tsx
bruce-glazier: We may want to catch any errors here but I'm not sure why we would need to wait for this to complete before proceeding. At this point, we've already set the activeAccountId in state so we really only need to store this for the next session. This goes to the below explanation about why I wrapped the get async storage with `react-query`, mostly to keep the effect cleaner without having to convert it to be async.

src/hooks/useActiveProject.tsx
bruce-glazier: True enough, we could always just fall back to use the default in that event.

src/hooks/useActiveProject.tsx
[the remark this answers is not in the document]
bruce-glazier: I've opened [a PR](https://github.com/lifeomic/react-native-sdk/pull/69) to include setItem within the hook to make usage a bit more consistent. With those changes do you feel I've properly addressed or replied to your concerns?

src/hooks/useActiveProject.tsx
```diff
@@ -50,35 +59,54 @@ export const ActiveProjectContextProvider = ({
...
   );
 
   useEffect(() => {
-    if (projectsResult.data?.length) {
-      // TODO: Load previously-selected project
-
-      const firstProjectId = projectsResult.data?.[0].id;
-      if (firstProjectId) {
-        setActiveProjectId(firstProjectId);
+    if (projectsResult.data?.length || storedProjectIdResult.isFetched) {
```
bruce-glazier, line 73: Should be `&&` here, will push an update.

bruce-glazier: Done

src/hooks/useActiveProject.tsx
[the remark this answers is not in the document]
markdlv: Yes- that consistency makes a big diff for me, thanks!  One can always use `@react-native-async-storage/async-storage` directly, or wherever getting a react-query interface for it is helpful one can use `useAsyncStorage` - dig it.

src/hooks/useActiveProject.tsx
markdlv: But in this case, the callback is already `async`.  And agreed- if you still feel we shouldn't wait/verify it's stored before moving on, we need to (whether in this PR or down the road) log the error or something so it doesn't just go unnoticed.

src/hooks/useActiveProject.tsx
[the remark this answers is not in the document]
bruce-glazier: In that case, I'd like to get https://github.com/lifeomic/react-native-sdk/pull/69 merged then rebase this one with master to get it updated.

src/hooks/useActiveProject.tsx
```diff
@@ -50,26 +59,34 @@ export const ActiveProjectContextProvider = ({
...
   );
 
   useEffect(() => {
-    if (projectsResult.data?.length) {
-      // TODO: Load previously-selected project
-
-      const firstProjectId = projectsResult.data?.[0].id;
-      if (firstProjectId) {
-        setActiveProjectId(firstProjectId);
+    if (projectsResult.data?.length && !storedProjectIdResult.isLoading) {
```
markdlv, line 73: `!storedProjectIdResult.isLoading` doesn't necessarily mean that `storedProjectIdResult.data` is populated.  e.g. it could be that `storedProjectIdResult.error` is populated instead?  I think you have the code within this if statement defensive enough to handle `storedProjectIdResult.data` being undefined, but there's probably an edge case bug in here (e.g. first project fetch succeeded, but subsequent one failed).  Not worried about this atm, just pointing it out as another factor in the react-query approach

markdlv (approved):

src/hooks/useActiveProject.tsx
bruce-glazier: Resolved by refactor

src/hooks/useActiveProject.tsx
```diff
@@ -26,6 +29,31 @@ const ActiveProjectContext = createContext({
...
+    const defaultProject = projects?.[0];
+    const defaultSubject = subjects?.find(
+      (s) => s.projectId === defaultProject?.id,
+    );
+    if (!defaultProject || !defaultSubject) {
+      console.warn('Failed attempt to select default projectId', projectId);
+      return { selectedProject: undefined, selectedSubject: undefined };
+    }
+    return { selectedProject: defaultProject, selectedSubject: defaultSubject };
+  }
```
namick, line 53: General question about how we want to use `console.warn` in this project.  Are these warnings for us, or for the end consumer/developer who might be building with this SDK in their app?  Obviously these are only shown in development mode.  Also, isn't it a normal, supported case that a user may not have any projects, and thus a console warning might be noise?

The PR looks good though, not wanting to hold it up with my questions.  👍

namick (approved):

src/hooks/useActiveProject.tsx
sternetj: Should add something like `[LifeOmicSDK]` prefixes to logs and errors so that user's can quickly identify where the error is coming from?

src/hooks/useActiveProject.tsx
```diff
@@ -36,40 +65,63 @@ export const ActiveProjectContextProvider = ({
   const [hookReturnValue, setHookReturnValue] = useState<ActiveProjectProps>(
     {},
   );
+  const { useGetItem, setItem } = useAsyncStorage();
```
sternetj, line 68: Why does this hook return a hook? why cant we do something like

```
const [storedProjectIdResult, setStoredProjectId] = useAsyncStorage(selectedProjectIdKey);
```

src/hooks/useActiveProject.test.tsx
```diff
@@ -142,3 +170,56 @@ test('setActiveProjectId ignores invalid projectId', async () => {
...
+  nextRerender({});
+  expect(nextResult.current).toMatchObject({
+    activeProject: mockSubjectProjects[1],
+    activeSubjectId: mockMe[1].subjectId,
+  });
+
+  useAsyncStorageSpy = jest.spyOn(useAsyncStorage, 'useAsyncStorage');
+});
+
+test('initial render writes selected account to async storage', async () => {
```
sternetj, line 204: Is this test title correct?

src/hooks/useActiveProject.tsx
bruce-glazier: We can and we should, I like that even better.

src/hooks/useAsyncStorage.ts
```diff
@@ -1,17 +1,15 @@
...
+export function useAsyncStorage(key: string) {
+  const itemResult = useQuery<string | null>('async-storage-get', () =>
+    AsyncStorage.getItem(key),
+  );
 
-  const setItem = (key: string, value: string) =>
-    AsyncStorage.setItem(key, value);
+  const setItem = (value: string) =>
+    AsyncStorage.setItem(key, value).catch((error) =>
+      console.error(`[LifeOmicSDK]:${error}`),
```
bruce-glazier, line 11: I think in a future PR I will purpose a wrapper or something around console to include that prefix.

sternetj (approved):

PR #71, opened 2023-03-07
PHC-4075 - Add style to settings screen

namick:
## Changes

  - Add style to settings screen

## Screenshots

# before

Screenshot 2023-03-07 at 12 24 35 PM

# after

Screenshot 2023-03-07 at 12 23 20 PM

### DIFF
--- a/src/navigators/TabNavigator.tsx
+++ b/src/navigators/TabNavigator.tsx
@@ -26,7 +26,8 @@ export function TabNavigator() {
         component={SettingsStack}
         options={{
           tabBarLabel: t('tabs-settings', 'Settings'),
-          headerShown: false,
+          headerShown: true,
+          headerTitle: 'Settings',
         }}
       />
     </Tab.Navigator>
--- a/src/screens/SettingsScreen.tsx
+++ b/src/screens/SettingsScreen.tsx
@@ -3,8 +3,10 @@ import {
   ScrollView,
   StyleSheet,
   Text,
+  TextStyle,
   TouchableOpacity,
   View,
+  ViewStyle,
 } from 'react-native';
 import { t } from 'i18next';
 import DeviceInfo from 'react-native-device-info';
@@ -15,29 +17,12 @@ import { OAuthLogoutButton } from '../components/OAuthLogoutButton';
 import { tID } from '../common/testID';
 import { useActiveAccount } from '../hooks/useActiveAccount';
 import { SettingsStackParamList } from '../navigators/SettingsStack';
+import { useStyles } from '../hooks/useStyles';
+import { Theme } from '../components/BrandConfigProvider/theme/Theme';
 
 const versionNumber = DeviceInfo.getVersion();
 const buildNumber = DeviceInfo.getBuildNumber();
 const fullVersion = versionNumber + ' (' + buildNumber + ')';
-
-const MainMenuItem = ({
-  title,
-  action,
-  badge,
-}: {
-  title: string;
-  action?: () => void;
-  badge?: React.ReactNode;
-}) => {
-  return (
-    <TouchableOpacity onPress={action} accessibilityRole="button">
-      <View style={styles.mainMenuItem}>
-        <Text>{title}</Text>
-        {badge}
-      </View>
-    </TouchableOpacity>
-  );
-};
 
 type NavigationParams = NativeStackNavigationProp<
   SettingsStackParamList,
@@ -47,6 +32,7 @@ type NavigationParams = NativeStackNavigationProp<
 export const SettingsScreen = () => {
   const { account } = useActiveAccount();
   const { navigate } = useNavigation<NavigationParams>();
+  const { styles } = useStyles('SettingsScreen', defaultStyles);
 
   return (
     <View style={styles.container}>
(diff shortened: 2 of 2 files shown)


src/screens/SettingsScreen.tsx
```diff
@@ -56,47 +42,114 @@ export const SettingsScreen = () => {
...
+
+function MainMenuItem({ title, action, badge }: Props) {
+  const { styles } = useStyles('SettingsScreen', defaultStyles);
+
+  return (
+    <TouchableOpacity onPress={action} accessibilityRole="button">
+      <View style={styles.mainMenuItem}>
+        <Text style={styles.mainMenuItemText}>{title}</Text>
+        {badge}
+        <Text style={styles.arrow}>{'>'}</Text>
```
namick, line 81: TODO: Replace this with default actual icons. Possibly from react native paper.  Also maybe make an optional icon prop.

src/screens/SettingsScreen.tsx
```diff
@@ -15,30 +17,13 @@ import { OAuthLogoutButton } from '../components/OAuthLogoutButton';
...
-}) => {
-  return (
-    <TouchableOpacity onPress={action} accessibilityRole="button">
-      <View style={styles.mainMenuItem}>
-        <Text>{title}</Text>
-        {badge}
-      </View>
-    </TouchableOpacity>
-  );
-};
```
markdlv, line 40: Dig the simplification- thanks

src/screens/SettingsScreen.tsx
markdlv: It might be better to either put that TODO in code OR just leave this out for now.

src/screens/SettingsScreen.tsx
```diff
@@ -56,47 +42,114 @@ export const SettingsScreen = () => {
...
-const styles = StyleSheet.create({
-  mainMenuItem: {
+interface Props {
+  title: string;
+  action?: () => void;
+  badge?: React.ReactNode;
+}
+
+function MainMenuItem({ title, action, badge }: Props) {
+  const { styles } = useStyles('SettingsScreen', defaultStyles);
```
markdlv, line 74: Interesting- are there pros/cons to reusing a name & defaultStyles like this vs. us trying to be strict about each component getting its own name/defaultStyles?  I can't think of any major cons offhand looking at the `useStyles` impl, and dig the simplicity of this approach.

markdlv (approved):

src/screens/SettingsScreen.tsx
namick: Yes, I believe this is the correct approach.  

The purpose of our `useStyles` hook and the type-building we are doing is to create a single namespace of customizable style objects for every component that we export out of this library.  This also becomes part of the public API that can't be changed without major version bumps.  

As is, these sub-components can be refactored without breaking changes.  This also enables users to be able to customize a component by choosing from a set of style objects under a single namespace that is also the name of the component they are trying to style.

markdlv: Nice!  That's a huge pro I wasn't even thinking about - good deal.

markdlv (approved):

src/screens/SettingsScreen.tsx
sternetj: We want to make sure that we do this carefully though and may even want to consider a different approach. If we have a lot of these subcomponents that attach their style definition to a single namespace, it is likely there will be an overlap which can result in some less flexible styling abilities. For example, if `MainMenuItem` and `SettingsScreen` both defined a `container` style property, the consumer would only be able to provide one style override for this shared definition.

In your current use-case everything looks good though since the styles are only defined once in `defaultStyles` but it is definitely something to keep an eye on if we start splitting these shared definitions across multiple files.

sternetj: Additionally, we might consider allowing namespaces to use a dot notation to provide further granularity and grouping of component definitions, i.e. `SettingsScreen.MainMenuItem`

src/screens/SettingsScreen.tsx
```diff
@@ -56,47 +42,117 @@ export const SettingsScreen = () => {
...
+  return {
+    mainMenuItem,
+    mainMenuItemText,
+    arrow,
+    subMenuContainer,
+    container,
+    versionContainer,
+    scroll,
+    divider,
+  };
```
sternetj, line 150: I'm still not in love with this pattern because it forces us to make new variables for every named style, produces lengthy styles for even small style definitions, and is frequently redundant because of definitions like `const fooText: TextStyle = {`

I propose we create a utility function that infers the style types based on property suffixes and allow for providing the namespace in a single place that can be used to extract the namespace in both `ComponentNamedStyles` and the `useStyles` hook.

```typescript
type StringLiteral<T> = T extends `${string & T}` ? T : never;
type IgnoreCapitals<T> = T extends string ? Uncapitalize<T> | Capitalize<T> : never;
type TextSuffixes = 'Text' | 'Label';
type ImageSuffixes = 'Image';

type NamedStyles<T> = {
    [P in keyof T]:
    P extends `${string}${IgnoreCapitals<TextSuffixes>}` ? TextStyle :
    P extends `${string}${IgnoreCapitals<ImageSuffixes>}` ? ImageStyle :
    ViewStyle
}

type CreateStyles = <N extends string, T extends NamedStyles<T>>(
    name: N,
    builder: T | ((theme: Theme) => T)) => { namespace: StringLiteral<N> } & NamedStyles<T>;

const createStyles: CreateStyles = (namespace, builder) => {
    const styles = typeof builder === 'function' ? builder({ theme: 'theme' }) : builder;
    return Object.assign({ namespace }, styles) as any;
}
```

[TS Playground Demo](https://www.typescriptlang.org/play?ts=4.8.4#code/C4TwDgpgBAKgFhAttAvFA3lYDkC4oDk2SEBUAvgNwBQoksEAHsAMqgA2qGWTrHEAZ3wDgAJwCWAOwDmFWuGgA1cRADubEJyhpMANxXr+QqCIky5daAElEAQ2kQNWnVHF2HTwcLFTZ5avL0bGbSADLiwBCituwAPDAAfNqwULwQkgAmAlAABgAk6Ka+UABksOQ5UAD8KfiSELpRNJZQVtKSAPaiEADCtmARMQLxSWgwqczpWSY+5jUAqpIAxv2D7OIAXhAjUAA+UH0DwDGb24lQdQ1NAS0wvCwArgBmT+KMgskEd8xk+wShtgARhB2ARmgpWu5HM9Xu9smgCDZ7KQaIFoAA5WzIDKeYbnHTUKBEqAAbQACq5JFAANYQEAdJ6wAC6uEJxIpaUy2XyhVm0nIBTanW6hzWePuMLeggSFWqDGYngubKJHMmXNyBSKMgF6CFXV6q2O7GGSI8krhMsqNVNjn4SuJUGUak81H8aIO3VskVxyVi6ImkXVWukABoUpzppjsbiRgkABTKqCSLEQfDokOJwEPcTsDJRfDjfZxuPEPCwHAQACU2iSMEr1ZQSUwyeQAjAtiWqagwV84Ui0Ti6KS5FKUCjEBxRhGqKWHUkIigS093qM+B6y9tmg+aDjLcE7c7YazObzogbTcTs-nwBmW-hWAUDKgx9zUW0KARTwey2A4jnZBqF9TzjTBSy7IgKzIchq3wIDrgdbpgAeUQqQAeUBAArCAlmAAA6WwBAEcR2hApMUzbDtoHIMMRDvasCKgWxJBAGg3U4G9IhEAAmZIlwgL1N04AQ4wIJ4Og6Agwzjc8oBAxMxI6b5gHwUD7lXQhIh+MNqPk8SAWBdgVJ4BV1KIXgoIzB12CBEEjM0vg73wMyfgoSziSvY4pHzbh9GdUzfNUCzE0QEAbSMtxkVxJyIocIKHXslTE3itTHMIBSCETfxoMrAJ2J4EQABZeI3XERMBWxREk2T0F0xTeDslKhKc+yqp0h0FP02zuHsqKNPM1zE2sgyGpM1LnOAOL3LnTz6lEIyAt6ggAsmokQrC7gYsErxCE2oLspuQRgC43C9woztaEO46FKU3CeqMZpLtwjqbPYW7GsEB7uNwjzbC80RcIW+6Lq+taoVwzbcVRTjgAKk7yIPCBgZhp7xJuu670+5HnoMt7RqEzHYZ+v6AYMSGkdh0HkXBqFIaAA)

This allows us to create styles in a concise way and allow the name we use to strongly type the property values and produce a return type that can be easily understood by a library consumer.

```typescript
// Styles would now look like this
const defaultStyles = createStyles('Example', (theme) => ({
  label: {
    fontSize: theme.font.body
  },
  image: {
    resizeMode: 'cover'
  },
  headerText: {
    fontSize: 20
  }
});

// as opposed to this
const defaultStyles = (theme) => {
  const label: TextStyle = {
    fontSize: theme.font.body
  };

  const image: ImageStyle = {
    resizeMode: 'cover'
  };

  const headerText: TextStyle = {
    fontSize: 20
  }

  return {
    label,
    image,
    headerText
  };
};
```

namick: Love this, I'll create a PR for it.

src/screens/SettingsScreen.tsx
namick: Yea, for large components that are composed of many smaller components with their own files, I would think we would structure it in a way that made this very clear.  Likely using the filesystem to collect a component with it's subcomponents.

 But from the user's perspective, it is important that they don't have to dig around in the internals of the component to figure out how to style it.  We want to be able to produce a simple, straightforward API documentation for each component we export.

src/navigators/TabNavigator.tsx
```diff
@@ -26,7 +26,8 @@ export function TabNavigator() {
         component={SettingsStack}
         options={{
           tabBarLabel: t('tabs-settings', 'Settings'),
-          headerShown: false,
+          headerShown: true,
+          headerTitle: 'Settings',
```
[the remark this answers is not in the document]
namick: I'll get this in the next PR.

src/screens/SettingsScreen.tsx
sternetj: It might be tricky for things like this where we can't control the name though: https://github.com/lifeomic/react-native-sdk/pull/75/files/8f716ccb43959444b3524ab136695f8b8dac2b09#diff-461a3dcd13f11b9d2ca8ac026e59da09377b9f41a4cf34c69c3831ac88e82490R47-R52

I bet there is a way we can make it work though.

PR #72, opened 2023-03-07
PHC-4075 - Suggest using npx pod-install

namick:
I've been using this command from expo instead of `cd ios && pod install && cd ..`.  Should be easier on end users.

## Changes

  - Add `pod-install` to dev dependencies
  - Add  `yarn pod-install` to readme (from https://github.com/expo/expo-cli/tree/main/packages/pod-install)

## Screenshots

Screenshot 2023-03-07 at 10 44 33 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ starting point.
 
 - `yarn add @lifeomic/react-native-sdk`
 - Install peer dependencies listed in `package.json`
-- `npx pod-install`
+- `yarn pod-install`
 
 See `example/AppDemo.tsx` for simple usage, which could be as simple as:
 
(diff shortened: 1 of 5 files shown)


markdlv:
Not opposed to its usage, but what do you think about adding it as a package dev dependency instead of relying on the first run to prompt the user with `Need to install the following packages: pod-install@...` ?

namick:
> Not opposed to its usage, but what do you think about adding it as a package dev dependency instead of relying on the first run to prompt the user with `Need to install the following packages: pod-install@...` ?

Interesting idea.  Would you still call it with npx and thus run the latest package?

namick:
> > Not opposed to its usage, but what do you think about adding it as a package dev dependency instead of relying on the first run to prompt the user with `Need to install the following packages: pod-install@...` ?

> Interesting idea.  Would you still call it with npx and thus run the latest package?

Looking at this again, I think I know what you mean.  Then we would ask the user to use `yarn pod-install`.  

I'll give this a try.

namick:
Verified and updated.  `yarn pod-install` seems to work well.

PR #74, opened 2023-03-07
PHC-4075 - Update theming to use react native paper

namick:
## Changes

  - Update theming to use react native paper

## Screenshots

No visual changes


src/common/RootProviders.tsx
```diff
@@ -29,12 +30,14 @@ export function RootProviders({
             <OAuthContextProvider authConfig={authConfig}>
               <ActiveAccountContextProvider>
                 <ActiveProjectContextProvider>
-                  <NoInternetToastProvider>
-                    <SafeAreaProvider>
-                      <NavigationContainer>{children}</NavigationContainer>
-                      <Toast />
-                    </SafeAreaProvider>
-                  </NoInternetToastProvider>
+                  <BrandConfigProvider>
```
namick, line 33: The Paper Provider which is wrapped in the BrandConfig provider needs to be outside any UI elements, but inside most other networking or store providers because `Modal` components are portaled here, at the top level.

src/components/BrandConfigProvider/styles/StylesProvider.tsx
```diff
@@ -1,26 +1,25 @@
 import * as React from 'react';
-import { merge } from 'lodash';
+import merge from 'lodash/merge';
```
namick, line 2: So we don't import all of lodash

src/components/BrandConfigProvider/theme/ThemeProvider.tsx
```diff
@@ -1,17 +1,28 @@
...
-  );
-}
+export function ThemeProvider({ theme: customTheme, children }: Props) {
+  const theme = merge({}, defaultTheme, customTheme);
 
-export const ThemeContext = React.createContext<{ theme: Theme }>({
-  theme: new Theme(),
-});
+  return <PaperProvider theme={theme}>{children}</PaperProvider>;
+}
```
namick, line 28: Currently, this doesn't leave room for a typescript user to add a custom theme property that is typed properly because the type is defined here.  I wanted to keep it simple for this PR but we can add a generic here in the future.

src/components/BrandConfigProvider/theme/base/index.ts
```diff
@@ -1,5 +1,4 @@
 export * from './colors';
 export * from './spacing';
 
-export const roundness = 4;
 export const buttonRoundness = 20;
```
namick, line 4: `roundness` is already in Paper.  `buttonRoundness` is our custom property.

src/hooks/useTheme.test.tsx
```diff
@@ -14,18 +16,18 @@ const renderHookInContext = async (theme?: Theme) => {
...
 test('returns theme', async () => {
-  const { result } = await renderHookInContext();
+  const {
+    result: { current: theme },
+  } = await renderHookInContext();
 
-  expect(result.current.theme).toBeInstanceOf(Theme);
-  expect(result.current.theme).toEqual(baseTheme);
+  expect(theme.colors.onBackground).toBe(baseTheme.colors.onBackground);
+  expect(theme.colors.background).toBe(baseTheme.colors.background);
```
namick, line 24: Paper theme has other defaults that our baseTheme doesn't worry about.  So I'm just testing a couple properties rather than asserting that the whole thing is the same.

src/components/BrandConfigProvider/styles/StylesProvider.tsx
```diff
@@ -1,26 +1,25 @@
...
   NamedStyles,
   BrandConfigProviderStyles,
 } from './types';
-import { Theme } from '../theme/Theme';
-import { useTheme } from '../../../hooks/useTheme';
+import { useTheme } from '../theme/ThemeProvider';
 
 export const StylesContext = React.createContext({
-  componentStyles: {} as RecursivePartial<ComponentStyles>,
+  styles: {} as RecursivePartial<ComponentStyles>,
```
markdlv, line 13: Why this rename?

namick: It was part of keeping the external API of the StylesProvider clean:

```TSX
  <ThemeProvider theme={theme || {}}>
    <StylesProvider styles={styles || {}}>{children}</StylesProvider>
  </ThemeProvider>
```

From that perspective, the styles being passed in are not necessarily `componentStyles`, but a potentially other styles as well.  This leaves the door open to other style use and making the system easier to change in the future.

Getting specific is only needed on line 22 below.

The type on line 13 above might need to change if we add other types of styles, but that is easy to do without changing the external API.

example/storybook/stories/BrandConfigProvider/BrandConfigProvider.stories.tsx
```diff
@@ -16,7 +16,7 @@ storiesOf('BrandConfigProvider', module)
     const customColors = object('theme.colors', baseTheme.colors);
     const customSpacing = object('theme.spacing', baseTheme.spacing);
```
bruce-glazier, line 17: I could be wrong about this but having only spacing defined in our theme feels like it is not granular enough to be useful. For example, if we are using `spacing.medium` globally for things like margin, padding, etc, won't it be total mayhem if that value is tweaked? Perhaps splitting these out by usage would be helpful, i.e., `margin.small` or `padding.medium` to get more predicable results could help? Even still I am having a hard time imaging a scenario where tweaking these global spacing values would have the desired outcome, almost feels like these should just be constants excluded from our theming.

namick: Yes, I agree 💯 that it isn't granular enough to be useful.  I would be in favor of removing it.  It isn't a part of MD either, and Paper seems to just have sensible and opinionated defaults in density-independent pixels that work for most apps and allow styling to override it.

This is interesting: https://m3.material.io/foundations/layout/understanding-layout/spacing

But yea, I'm in favor of removing spacing.

sternetj: Great point @bruce-glazier. Probably best to just bake those into the default styles and let a consumer tweak them if they want.

src/components/BrandConfigProvider/styles/StylesProvider.tsx
sternetj, line 13: I feel like eventually we might want to allow styles to be an object or a function that accepts the theme as an argument. This would allow a consumer to override some parts of the theme, and then construct the component style from the built theme. That way if they want to change the background and foreground colors of a component it would be easy to switch from primary/onPrimary (default) to secondary/onSecondary (custom) if they wanted and wouldn't require that they set that up ahead of time to override that color via the `StylesProvider` prop.

package.json
```diff
@@ -87,6 +87,8 @@
   },
   "dependencies": {
     "lodash": "^4.17.21",
+    "react-native-paper": "^5.3.1",
+    "react-native-vector-icons": "^9.2.0",
```
sternetj, line 91: I know loading fonts can be tricky sometimes, is this handled by autolinking?

sternetj (approved):

package.json
namick: No, in fact Paper links here: https://github.com/oblador/react-native-vector-icons#installation and those instructions have you disable autolinking for react-native-vector-icons.   It works, [PR is here](https://github.com/lifeomic/react-native-sdk/pull/77), but I'm not 100% that this is the best way.

sternetj: Interesting, this would mean a consumer would have to do this themselves if they set up their own repo or we would have to include it in the starter repo. I wonder if we should just ship with chroma icons instead? I am not sure we have chroma as font though.

namick: I think for the moment we should just include it with the starter repo.  It's actually not that complicated or problematic, I'm just surprised it is not more automated.

For the most part, I'm guessing consumers are going to want to use their own icons anyway.

PR #75, opened 2023-03-08
PHC-4075 - Update OAuthLoginButton to use Paper

namick:
## Changes

  - Update OAuthLoginButton to use Paper

## Screenshots

No visual changes

### DIFF
--- a/src/components/OAuthLoginButton.test.tsx
+++ b/src/components/OAuthLoginButton.test.tsx
@@ -22,7 +22,7 @@ test('renders', () => {
   const screen = render(
     <OAuthLoginButton onSuccess={onSuccess} onFail={onFail} label="login" />,
   );
-  expect(screen.getByTestId('oauth-login-button')).toBeDefined();
+  expect(screen.getByTestId('button')).toBeDefined();
 });
 
 test('login utilizes useOAuthFlow', async () => {
@@ -30,7 +30,7 @@ test('login utilizes useOAuthFlow', async () => {
     <OAuthLoginButton onSuccess={onSuccess} onFail={onFail} label="login" />,
   );
   await waitFor(async () => {
-    const loginButton = screen.getByTestId('oauth-login-button');
+    const loginButton = screen.getByTestId('button');
     await fireEvent.press(loginButton);
   });
   expect(loginMock).toHaveBeenCalledWith({
--- a/src/components/OAuthLoginButton.tsx
+++ b/src/components/OAuthLoginButton.tsx
@@ -1,31 +1,23 @@
-import React, { FC, useCallback } from 'react';
-import {
-  Platform,
-  TouchableOpacity,
-  TouchableOpacityProps,
-  View,
-  ViewStyle,
-  Text,
-  TextStyle,
-} from 'react-native';
+import React, { useCallback } from 'react';
+import { TextStyle, TouchableOpacityProps, ViewStyle } from 'react-native';
 import { useStyles } from '../../src/hooks/useStyles';
 import { LoginParams, useOAuthFlow } from '../../src/hooks/useOAuthFlow';
-import { tID } from '../common/testID';
 import { Theme } from './BrandConfigProvider';
+import { Button } from 'react-native-paper';
 
-type OAuthLoginButtonParams = Omit<TouchableOpacityProps, 'onPress'> &
+type Props = Omit<TouchableOpacityProps, 'onPress'> &
   LoginParams & {
     label: string;
     styles?: OAuthLoginButtonStyles;
   };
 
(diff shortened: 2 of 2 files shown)


src/components/OAuthLoginButton.test.tsx
```diff
@@ -22,15 +22,15 @@ test('renders', () => {
   const screen = render(
     <OAuthLoginButton onSuccess={onSuccess} onFail={onFail} label="login" />,
   );
-  expect(screen.getByTestId('oauth-login-button')).toBeDefined();
+  expect(screen.getByTestId('button')).toBeDefined();
```
namick, line 25: Paper has default `testID` values.

src/components/OAuthLoginButton.tsx
```diff
@@ -1,97 +1,55 @@
...
-    lineHeight: 20,
-    fontSize: 14,
-  };
+const defaultStyles = (_theme: Theme) => {
+  const style: ViewStyle = {};
+  const contentStyle: ViewStyle = {};
+  const labelStyle: TextStyle = {};
 
-  return { button, touchableOpacity, content, label };
+  return { style, contentStyle, labelStyle };
```
namick, line 52: These three style objects are available on Paper's `Button`.  Including them here makes `OAuthLoginButton` styleable globally and eventually in the app config service.

src/components/OAuthLoginButton.tsx
```diff
@@ -1,97 +1,55 @@
...
-      </TouchableOpacity>
-    </View>
+    <Button
+      mode="contained"
+      onPress={handleOnPress}
+      {...styles}
+      {...buttonProps}
+    >
+      {label}
+    </Button>
```
namick, line 43: All of Paper's Button [props](https://callstack.github.io/react-native-paper/docs/components/Button/) like `buttonColor`, `textColor`, `loading`, `icon`, `disabled`, etc are still available on an instance level.

src/components/OAuthLoginButton.test.tsx
markdlv: 🤔 gut reaction: how do broad testIDs like `button` work with multiple buttons on screen?

src/components/OAuthLoginButton.tsx
```diff
@@ -1,97 +1,55 @@
...
 
-type OAuthLoginButtonParams = Omit<TouchableOpacityProps, 'onPress'> &
+type Props = Omit<TouchableOpacityProps, 'onPress'> &
   LoginParams & {
     label: string;
     styles?: OAuthLoginButtonStyles;
   };
 
-export const OAuthLoginButton: FC<OAuthLoginButtonParams> = ({
+export function OAuthLoginButton({
```
bruce-glazier, line 14: Just curious, what was your motivation for converting this from an arrow function?

src/components/OAuthLoginButton.tsx
```diff
@@ -1,97 +1,55 @@
...
-        style={styles.touchableOpacity}
-      >
-        <View style={styles.content}>
-          <Text style={styles.label}>{label}</Text>
-        </View>
-      </TouchableOpacity>
-    </View>
+    <Button
+      mode="contained"
+      onPress={handleOnPress}
```
bruce-glazier, line 38: IMO we should keep `testID={tID('oauth-login-button')}` here. A consumer of this library would want a more specific default name for writing tests, especially E2E tests.

namick: OK, I'm okay with that.  But couldn't (shouldn't) they just add their own testID in their test?

src/components/OAuthLoginButton.test.tsx
namick: That is just the default, but in a more complicated test, you probably want to add your own:

```TSX
const screen = render(<OAuthLoginButton testID="button-47" />)

expect(screen.getByTestId('button-47')).toBeDefined();
```

namick: I'm happy to add it back though if we want a different "default" for our OAuth buttons

sternetj (dismissed):

src/components/OAuthLoginButton.tsx
namick: I added the more specific testId back here.  👍

src/components/OAuthLoginButton.tsx
bruce-glazier: Thanks for the detailed explanation, I like it and agree. We should see if we can get a lint rule in place to ensure we follow this pattern moving forward (provided we all agree).

bruce-glazier (approved):

sternetj (approved):

markdlv (approved):

PR #77, opened 2023-03-08
PHC-4075 - Add font icons to Android and iOS

namick:
## Changes

  - Added font icons for android and ios

## Screenshots

Screenshot 2023-03-08 at 4 40 24 PM


sternetj (approved):
What happens if the font is not correctly loaded into the app? Do things look wonky? I wonder if it would we better to disable icons if we don't think we will need them.

Since this is a LifeOmic sdk I would expect it to ship with LifeOmic icons so we present a consistent look for customers.

namick:
> What happens if the font is not correctly loaded into the app? Do things look wonky? I wonder if it would we better to disable icons if we don't think we will need them.
> 
> Since this is a LifeOmic sdk I would expect it to ship with LifeOmic icons so we present a consistent look for customers.

If the font is not correctly loaded into the app the developer gets a warning and a default question mark icon is shown (when `__DEV__` is true).

This PR is adding the icons to the example app.  This will allow (potential) customers to play with the example app and see icons working.  This doesn't add the icons to the sdk, although our sdk, by proxy of Paper, does support them.  We may also want to add LifeOmic icons in the future, but I don't think we should disable icon functionality.

My understanding is that the example app in this repo and `lifeomic/react-native-starter` should decidedly not look like the LifeOmic app, but instead have a generic, "wireframe" theme.   I don't know for sure, but as an open source project built on another open source project, shipping with lifeomic icons doesn't feel to me like a requirement.  🤔

markdlv (approved):
Yea, it's tricky to decide what should go into the example/starter apps vs. what takes the "example" too far.  Our LifeOmic icons/fonts are also open-source, so an argument is definitely there that example/default icons should be those.  I think what we have in this PR is fine for now - we can iterate if needed.

PR #85, opened 2023-03-16
PHC-4075 - Use unique key for Async Storage

namick:
## Changes

  - Use unique key for Async Storage
  - Update Reactotron config to display Async Storage

## Screenshots

Screenshot 2023-03-16 at 7 42 14 PM

Screenshot 2023-03-16 at 7 41 17 PM

### DIFF
--- a/src/hooks/useAsyncStorage.ts
+++ b/src/hooks/useAsyncStorage.ts
@@ -2,7 +2,7 @@ import AsyncStorage from '@react-native-async-storage/async-storage';
 import { useQuery } from 'react-query';
 
 export function useAsyncStorage(key: string) {
-  const itemResult = useQuery<string | null>('async-storage-get', () =>
+  const itemResult = useQuery<string | null>(['async-storage-get', key], () =>
     AsyncStorage.getItem(key),
   );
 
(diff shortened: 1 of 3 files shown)


src/hooks/useAsyncStorage.ts
```diff
@@ -2,7 +2,7 @@ import AsyncStorage from '@react-native-async-storage/async-storage';
 import { useQuery } from 'react-query';
 
 export function useAsyncStorage(key: string) {
-  const itemResult = useQuery<string | null>('async-storage-get', () =>
+  const itemResult = useQuery<string | null>(['async-storage-get', key], () =>
     AsyncStorage.getItem(key),
   );
```
namick, line 8: This was a bug where `useQuery` global cache key was being set to `async-storage-get` for all Async Storage key/value pairs.  Adding the `key` to make it unique fixes it.  

I spent some time writing tests for this but eventually bailed on that because I'm not sure using `useQuery` here is a good solution anyway.

I would be in favor of replacing it with a light wrapper that abstracts away serialization/deserialization, maybe something [like this](https://github.com/infinitered/ignite/blob/master/boilerplate/app/utils/storage/storage.ts).  That way we could easily replace Async Storage with something much faster in the future if we want... [like this](https://github.com/mrousavy/react-native-mmkv).

sternetj (approved):

bruce-glazier (approved):
Thanks for catching this. You are probably right about useQuery not being the best solution here. At the time it seemed like it would maybe improve our code cleanliness but it is probably not a common use case of react-query.

markdlv (approved):

PR #91, opened 2023-03-22
PHC-4329 - Migrate TrackTile

namick:
## Changes

  - Imported TrackTile src code, name-spacing it under `src/components/TrackTile`
  - Updated and normalized packages, fixed paths and other configuration
  - Disabled TrackTile tests, and excluded it from the build for the moment
  - Added a single set of stories and got StoryBook to run existing (non TrackTile) stories

The [first commit ](https://github.com/lifeomic/react-native-sdk/pull/91/commits/27ce0f4cb7d826f6b05fc51f8a1351b5bb2c02be)is all the newly copied files, you'll probably want to skip that.  The [last commit ](https://github.com/lifeomic/react-native-sdk/pull/91/commits/539f295c5b0da154e80ac69b89c70170efcfa97f)is fixing prettier and linting, also a huge, automated changeset.

You might be interested in the [package changes](https://github.com/lifeomic/react-native-sdk/pull/91/commits/cd9da66b8a46867f56def14519057acd7f086d78) or [other commits](https://github.com/lifeomic/react-native-sdk/pull/91/commits) that are more digestible.


example/storybook/stories/TrackTile/util/MockEnvironmentDecorator.tsx
```diff
@@ -0,0 +1,76 @@
...
+          account: 'stub',
+          project: 'stub',
+        },
+        axiosInstance: axiosInstance as any,
+      })}
+    >
+      {children}
+    </TrackTileServiceProvider>
+  );
+}
```
namick, line 76: I had to pull this Provider out because we have linting rules against inline components

example/storybook/stories/TrackTile/util/MockEnvironmentDecorator.tsx
```diff
@@ -0,0 +1,76 @@
...
+        },
+        proceduresConnection: {
+          edges: [],
+        },
+      },
+    },
+  } as FetchTrackerResponse,
+} = {}) => {
+  // const axiosInstance = axios.create();
+  // const mock = new MockAdapter(axiosInstance);
```
namick, line 36: Oops, I'll delete these next PR

example/package.json
```diff
@@ -20,7 +20,10 @@
     "react-native-app-auth": "^6.4.3",
     "react-native-config": "^1.5.0",
     "react-native-device-info": "^10.3.0",
+    "react-native-gesture-handler": "^2.9.0",
     "react-native-keychain": "^8.1.1",
+    "react-native-reanimated": "^3.0.2",
+    "react-native-svg": "^13.8.0",
```
namick, line 26: These three required being added to the example app as well because they have native parts requiring pod-install.  Because of this I added them also to the peer-dependencies of the sdk

markdlv: 💯

package.json
```diff
@@ -52,9 +57,13 @@
     "react-native": "0.68.2",
     "react-native-app-auth": "^6.4.3",
     "react-native-device-info": "^10.3.0",
+    "react-native-draggable-flatlist": "^4.0.1",
```
markdlv, line 60: It doesn't seem like adding the deps above this line to `devDependencies` is right.  If, according to our [peer dependencies strategy](https://github.com/lifeomic/react-native-sdk/blob/master/CONTRIBUTING.md#peer-dependencies-strategy), you think they shouldn't be peer dependencies - I think they need to go into `dependencies`.  Otherwise these won't be included in a way the npm-published lib can use them.

markdlv (dismissed):
Goodness - the commit focus separation was **perfect**.  Thanks so much for doing that!  121-files-changed PR doesn't look fun, but reviewing this was a downright pleasant experience.

I've got the one note about `devDependencies` to look into, but approving in case a) I'm wrong or b) you want to iterate & fix in next PR.

package.json
namick: Yes, fixed.

PR #92, opened 2023-03-23
PHC-4329 - Update packages

namick:
## Changes

  - Downgrade react-native-reanimated to v2.8.0 to prevent crash
  - Add material-bottom-tabs and draggable-flatlist to peer dependencies
  - Add missing peer dependencies to example app


example/ios/Podfile.lock
```diff
@@ -408,7 +408,7 @@ PODS:
     - React-Core
   - RNKeychain (8.1.1):
     - React-Core
-  - RNReanimated (3.0.2):
+  - RNReanimated (2.8.0):
```
bruce-glazier, line 411: Do you plan to follow up on this later? It would be great if we could ship with the latest since updating these libraries can be difficult as the repo and it's usage grows.

bruce-glazier (dismissed):

example/package.json
```diff
@@ -53,8 +61,6 @@
     "metro-react-native-babel-preset": "0.73.7",
     "pod-install": "^0.1.38",
     "react-dom": "18.2.0",
-    "react-native-safe-area-context": "^4.5.0",
-    "react-native-screens": "^3.19.0",
```
markdlv, line 57: Good catch- thanks

markdlv (approved):
Good thoughts on reanimated version in comments, but also don't get too hung up on that or anything.  This is good as-is if you want to merge/iterate.

example/ios/Podfile.lock
namick: Yes, I'll plan to follow up on this later.  Upgrading to 3.0.2 was crashing without a lot of useful information.

markdlv (approved):

sternetj (approved):

PR #96, opened 2023-03-25
PHC-4329 - Fix TypeScript build errors

namick:
## Changes

  - Added @testing-library/react-hooks
  - Added types for d3-scale
  - Switched to @react-native-picker/picker
  - Updated to newer api in newer @testing-library
  - Fixed various typing issues

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,9 @@
     "@react-navigation/native": "^6.1.3",
     "@react-navigation/native-stack": "^6.9.9",
     "@testing-library/jest-native": "^5.4.1",
+    "@testing-library/react-hooks": "^8.0.1",
     "@testing-library/react-native": "^11.5.1",
+    "@types/d3-scale": "^3.3.0",
     "@types/jest": "^29.4.0",
     "@types/lodash": "^4.14.191",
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -95,7 +97,7 @@
   },
   "dependencies": {
     "@lifeomic/chromicons-native": "^2.14.0",
-    "@react-native-community/picker": "^1.8.1",
+    "@react-native-picker/picker": "^2.4.9",
```
markdlv, line 100: Can you say more about this switch and why it is needed?  The typing additions make sense, but I'm not clear on this part.

namick: Yes, track-tile-ui used `Picker` from `react-native` for android but also used `PickerIOS` from `@react-native-community/picker`. 

The newer React Native used in this SDK no longer ships with `Picker` because it was moved to `@react-native-community/picker`.  Upgrading to that community version caused tests to break, leading me to discover that the library has been renamed to `@react-native-picker/picker`.

Screenshot 2023-03-27 at 9 24 34 PM

Now both `Picker` and `PickerIOS` are imported from this same lib.

markdlv (approved):

PR #98, opened 2023-03-28
PHC-4329 - Update UnitPicker tests and story

namick:
## Changes

  - Make Picker a peer dependency
  - Add story and fix tests for UnitPicker


bruce-glazier (approved):

src/components/TrackTile/TrackerDetails/UnitPicker.tsx
```diff
@@ -87,10 +87,7 @@ export const UnitPicker: FC<UnitPickerProps> = (props) => {
                   variant="semibold"
                   style={styles.unitPickerPopupAccessoryDoneTextIOS}
                 >
-                  {i18n.t('f92965e2c8a7afb3c1b9a5c09a263636', {
```
bruce-glazier, line 90: I would expect to see this changes across all the language JSON files too. Usually I do a Replace All when updating the key.

markdlv: I really appreciate/prefer PRs focused on one thing.  If we take the "update all i18next keys" (whether soon or as a final wrap-up PR), it should be especially focused on that one thing because it'll be so large.

example/package.json
```diff
@@ -12,6 +12,7 @@
   "dependencies": {
     "@react-native-async-storage/async-storage": "^1.17.11",
     "@react-native-community/netinfo": "^9.3.7",
+    "@react-native-picker/picker": "^2.4.9",
```
markdlv, line 15: Just want to call it out to verify - did the previous `@react-native-community/picker` require a Pods install as well, or is that new to this newer picker?  No concern either way, just curious

markdlv (approved):

src/components/TrackTile/TrackerDetails/UnitPicker.tsx
namick: Correct.  I have a singular focused PR upcoming with all fixed.  This one was required for the test to pass.

example/package.json
namick: Yes, the previous did require it as well.  In fact it is largely the same codebase - just a renamed repo and then continued development.

src/components/TrackTile/TrackerDetails/UnitPicker.tsx
bruce-glazier: Updating the key name in these files would only change a single string across 7 files and usually it is easier to update these in the JSON files while you still have the original key. Also the lookups to the translations for this key are broken until the updates are made but that too probably isn't a big deal given we will just fallback to the default text.

markdlv: Thanks for clarifying Bruce - I misunderstood.  I thought you were suggesting we update **all** the i18n keys at once, and I got on a soapbox about PR focus 😆.  You're spot on- @namick since we imported track-tile-ui's translation (JSON) files, the string `f92965e2c8a7afb3c1b9a5c09a263636` is throughout those files and orphaned once this PR merges.  Assuming you already know that, and this is all just me catching up after butting in, all good 👍.  i.e. It's fine whether we update `f92965e2c8a7afb3c1b9a5c09a263636` -> `track-tile-done` in this PR or a subsequent one, as long as we know we're getting to it soon.

bruce-glazier (approved):

PR #99, opened 2023-03-29
PHC-4320 - Get more tests working

namick:
## Changes

  - Add some mocking and unmocking
  - Downgrade d3-scale

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
   },
   "dependencies": {
     "@lifeomic/chromicons-native": "^2.14.0",
-    "d3-scale": "^4.0.2",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -101,7 +101,7 @@
   },
   "dependencies": {
     "@lifeomic/chromicons-native": "^2.14.0",
-    "d3-scale": "^4.0.2",
+    "d3-scale": "3.2.1",
```
namick, line 104: This is the version working in track-tile-ui.  I should have left it.

markdlv (approved):

PR #107, opened 2023-04-03
Update Wearables to use the current themes/styles mechanism

bruce-glazier:
## Changes

The Wearables library had a theme/styles mechanism that wasn't the same as the one in this repo. This PR only focuses on updating wearables to use this libraries theme/style mechanism.

## Screenshots

image


src/components/Wearables/SelectorRow.tsx
```diff
@@ -56,22 +54,28 @@ export const SelectorRow: FC<SelectorRowProps> = (props) => {
...
+const defaultStyles = createStyles('SelectorRow', (theme) => ({
   container: {
     minHeight: 40,
   },
   choiceText: {
     textAlign: 'right',
   },
   textWrapper: {
-    padding: Padding.medium,
+    padding: 12,
```
markdlv, line 65: Feel free to set this to something in the theme if something is close, even if it changes the UI slightly.  Or, if you deem this is best, that's fine too.

bruce-glazier: I think it's generally better to set the margin and padding values for each element individually. I don't think setting these values globally at the theme level will work well everywhere.

namick (approved):

src/components/Wearables/SelectorRow.tsx
markdlv: I've definitely seen "everything is based on the same theme value" go wrong / not scalable.  But also worth noting: manually entered values can't be affected when branding unless one does "component style overrides" which is a pretty advanced topic.  On the "general" side, I'd lean more towards using theme values in our default styles - and use manually entered values only where we know it's an exception that shouldn't change with a theme.  But, for example, text padding being controlled via 1 theme prop across an app seems like a good thing.  Can you explain more why you're thinking having more manual values in here is better?

src/components/Wearables/SelectorView.tsx
```diff
@@ -31,15 +32,22 @@ export const SelectorView: FC<SelectorViewProps> = (props) => {
...
+          styles={styles.selectorRow as SelectorRowStyles}
           title={item.title}
         />
       ))}
     </View>
   );
 };
 
-export const SelectorViewDefaultStyles = {
+const defaultStyles = createStyles('SelectorView', () => ({
```
markdlv, line 43: I wonder if we'll someday need namespacing of some sort on these style names.  Many classes in this lib are pretty generically named.

src/components/Wearables/WearablesView.tsx
```diff
@@ -219,20 +218,28 @@ export const WearablesViewDefaultStyles = {
...
   },
   sectionHeader: {
-    marginHorizontal: Margin.standard,
-    marginBottom: Margin.small,
+    marginHorizontal: 16,
+    marginBottom: 8,
     fontSize: 28,
     fontWeight: '500',
-    marginTop: Margin.standard,
+    marginTop: 16,
```
markdlv, line 234: Yea the more manual numbers in here the more I ask myself, why wouldn't we use `theme` for these?  It could be that these are all sort of non-standard and converting some of these components to `react-native-paper` will standardize them in better prep for theme numbers?

markdlv (approved):

src/components/Wearables/SelectorRow.tsx
bruce-glazier: Good offline discussion about all of this, merging as is for now, but will be doing a follow up PR to address centralized spacing via the theme.

PR #108, opened 2023-04-03
Default Tile to fullLength

bruce-glazier:
## Changes

- Switch default display mode to full length tiles


markdlv (approved):
Sweet - great to see it's this easy

PR #113, opened 2023-04-04
PHC-4328 Add Wearables hooks and screen

bruce-glazier:
## Changes

  - Adds the necessary hooks for configuring wearables
  - Adds the Wearables Screen
  
## Next up: 
- Add unit tests/style TODOs around spacing

## Screenshots
Populated with mock data:


src/hooks/useWearables.ts
```diff
@@ -0,0 +1,114 @@
...
+  return setLastSync;
+};
+
+export const useSetWearableState = () => {
+  const { httpClient } = useHttpClient();
+  const { accountHeaders } = useActiveAccount();
+  const setWearableState = async (
+    ehrId: string,
+    enabled: boolean,
+    meta?: any,
```
sternetj, line 91: It be nice if we could add a type here instead of `any`

src/screens/WearablesScreen.tsx
```diff
@@ -0,0 +1,62 @@
...
+    [appId, setWearableState],
+  );
+
+  const updateSyncTypeSettings = async (settings: SyncTypeSettings) => {
+    await setSyncTypes(settings);
+    refetch();
+  };
+  return (
+    <View style={[styles.container]}>
+      <SafeAreaView style={[styles.container]}>
```
sternetj, line 39: Do we need this `SafeAreaView`? I thought react-navigation handled this for us.

sternetj (approved):

src/hooks/useWearables.ts
bruce-glazier: I can follow up on this. In general, I'm trying to limit the changes to a simple copy/paste where possible.

src/screens/WearablesScreen.tsx
bruce-glazier: You are probably right about that, I can follow up.

src/hooks/useWearables.ts
```diff
@@ -0,0 +1,114 @@
...
+  );
+};
+
+export const useWearableIntegration = (ehrId: string) => {
+  const { httpClient } = useHttpClient();
+  const { accountHeaders } = useActiveAccount();
+  return useQuery('/v1/wearables/{ehrId}', () => {
+    httpClient.get(`/v1/wearables/${ehrId}`, { headers: accountHeaders });
+  });
+};
```
markdlv, line 49: Cool - I've been thinking about how to go about multiple `useQuery`s grouped together.  There are a lot of repeated httpClient & accountHeaders LOC.  In the FHIR client first pass, I'm trying out having the file expose one hook that wraps several `useQuery`s as callbacks.  That might be the wrong approach and this is the right one - just wanted to mention as we're going to probably hit this in several places.

src/hooks/useWearables.ts
```diff
@@ -0,0 +1,114 @@
...
+export const useWearableIntegrations = (appId: string) => {
+  const { httpClient } = useHttpClient();
+  const { accountHeaders } = useActiveAccount();
+  return useQuery('/v1/wearables', () =>
+    httpClient
+      .get<WearablesSyncState>('/v1/wearables', {
+        params: { appId },
+        headers: accountHeaders,
+      })
+      .then((res) => res.data),
```
markdlv, line 39: Can we use `enabled` and check the presence of `accountHeaders` as a defense mechanism?

src/hooks/useWearables.ts
```diff
@@ -0,0 +1,114 @@
...
+};
+
+export const useSetSyncTypes = () => {
+  const { httpClient } = useHttpClient();
+  const { accountHeaders } = useActiveAccount();
+  const setSyncTypes = (settings: SyncTypeSettings) =>
+    httpClient.put('/v1/wearables/sync-types', settings, {
+      headers: accountHeaders,
+    });
+  return setSyncTypes;
```
markdlv, line 58: Did you consider using `useMutation` from `react-query` and decide against it?  While `useQuery` can be a little tougher to get right, I think it might be interesting to refactor this and the 2 methods below it into a single hook which exposes 3 async callbacks.  That'd allow for `httpClient` and `accountHeaders` sharing too.

markdlv (approved):

src/hooks/useWearables.ts
bruce-glazier: That would be interesting, I'd like to see how that would look too. I will try it out in my follow up PR.

src/hooks/useWearables.ts
markdlv: Just to close the loop: as mentioned [here](https://github.com/lifeomic/react-native-sdk/pull/121#discussion_r1160836483), this idea of mine _is_ a bad idea (to have useQuery underneath a useCallback - that'll hit [this issue](https://react.dev/warnings/invalid-hook-call-warning)).  Looks like you're doing it in a [subsequent PR](https://github.com/lifeomic/react-native-sdk/pull/120/files#diff-881ac6549e4fd0710b6837d26da387f05de27be84a226355c28832c3dcab9573R103) just like I did it in [useFhirClient first pass](https://github.com/lifeomic/react-native-sdk/pull/121).  Seems like next iteration is to try out this [lazy query pattern](https://tanstack.com/query/v4/docs/react/guides/disabling-queries#lazy-queries).

PR #116, opened 2023-04-05
Remote notifications support

aymanenadi:
## Changes
 - Remote notifications support

## Screenshots
https://user-images.githubusercontent.com/19395435/230129019-a34f9909-a778-43b9-aa1a-ff5c64980569.mov


example/ios/Podfile.lock
```diff
@@ -703,4 +709,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 446419b91f3d523996e9ca2fa51242b50bd76552
 
-COCOAPODS: 1.11.3
+COCOAPODS: 1.12.0
```
markdlv, line 712: I can get behind this, but we all kind of need to switch together.  Alternatively, we could back this and the Gemfile.lock out into a separate PR for clarity.  Both good ideas - just need to do this across all devs & repos.

example/ios/example/AppDelegate.mm
```diff
@@ -41,4 +46,15 @@ - (BOOL)concurrentRootEnabled
...
+// Push notifications registration
+- (void)application:(UIApplication *)application didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)deviceToken {
+  [RNNotifications didRegisterForRemoteNotificationsWithDeviceToken:deviceToken];
+}
+- (void)application:(UIApplication *)application didFailToRegisterForRemoteNotificationsWithError:(NSError *)error {
+  [RNNotifications didFailToRegisterForRemoteNotificationsWithError:error];
+}
+- (void)application:(UIApplication *)application didReceiveRemoteNotification:(NSDictionary *)userInfo fetchCompletionHandler:(void (^)(UIBackgroundFetchResult result))completionHandler {
+  [RNNotifications didReceiveBackgroundNotification:userInfo withCompletionHandler:completionHandler];
+}
```
markdlv, line 58: Cool - and similar to the native changes needed for react-native-app-auth, we can point folks to https://wix.github.io/react-native-notifications/docs/getting-started/.  Super long-term, we might encapsulate all of these kinds of things in our own native module, but I like this approach as our first full pass using these libs.

markdlv: Speaking of ☝️ should the Android changes be in this PR?  If/when we get there, can you provide a video of what the Android side looks like alongside iOS?

src/hooks/index.ts
```diff
@@ -13,3 +13,4 @@ export * from './useSubjectProjects';
 export * from './useStyles';
 export * from './useTheme';
 export * from './useUser';
+export * from './useNotifications';
```
markdlv, line 16: Nit-pick - can you put this in alphabetical?

src/common/index.ts
```diff
@@ -3,3 +3,4 @@ export * from './LoggedInProviders';
 export * from './RootProviders';
 export * from './SecureStore';
 export * from './testID';
+export * from './Notifications';
```
markdlv, line 6: Nit- can you put this in case-insensitive alphabetical?

src/common/Notifications.ts
```diff
@@ -0,0 +1,107 @@
...
+  callback: ({
+    deviceToken,
+    denied,
+    error,
+  }: {
+    deviceToken?: string;
+    denied?: boolean;
+    error?: RegistrationError;
+  }) => void,
+) => {
```
markdlv, line 48: Did you consider using a Promise ctor instead of callback?

src/common/Notifications.ts
```diff
@@ -0,0 +1,107 @@
...
+  Notifications.events().registerNotificationReceivedBackground(
+    (
+      backgroundNotification: Notification,
+      completion: (response: NotificationBackgroundFetchResult) => void,
+    ) => {
+      callback(backgroundNotification);
+      completion(NotificationBackgroundFetchResult.NEW_DATA);
+    },
+  );
+};
```
markdlv, line 93: 👍 I like the single callback for both foreground & background.  We can always add a prop to indicate which it is if a) it's not already there and b) we need it someday.

markdlv (dismissed):
LGTM - I've got a bunch of nit-pick comments but don't let them hold you up.  All worth reviewing, but feel free to iterate as needed.

Question: the `Send Notifications` button demo shown in the desc is not in this PR, right?  Can that go into an `example/` storybook in a subsequent PR?

example/ios/Podfile.lock
aymanenadi: I don't remember explicitly upgrading to `1.12.0`. I downgraded back to `1.11.3` until we get the chance to discuss as a team.

example/ios/example/AppDelegate.mm
aymanenadi: Yes, I will have a separate PR for Android along with a video.

package.json
```diff
@@ -91,6 +92,7 @@
     "react-native-draggable-flatlist": ">=3.0.6",
     "react-native-gesture-handler": ">=2.9.0",
     "react-native-keychain": ">=8.1.1",
+    "react-native-notifications": "^4.3.3",
```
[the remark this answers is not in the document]
aymanenadi: Good call. I'll update it.

src/common/Notifications.ts
aymanenadi: I did. I went with callbacks because they're the closest to the two most popular RN push notifications libraries APIs, [react-native-notifications](https://wix.github.io/react-native-notifications/api/general-events#registerremotenotificationsregistered) and [notifee](https://notifee.app/react-native/docs/events).

example/ios/Podfile.lock
```diff
@@ -437,10 +439,10 @@ PODS:
     - React-RCTText
     - ReactCommon/turbomodule/core
     - Yoga
-  - RNScreens (3.19.0):
+  - RNScreens (3.20.0):
     - React-Core
     - React-RCTImage
-  - RNSVG (13.8.0):
+  - RNSVG (13.9.0):
```
markdlv, line 445: Just to make sure - are these other pod updates intentional?  Not opposed to them, but from a purity perspective it'd be sweet to see this file be nothing but `react-native-notifications` related.

markdlv (approved):

example/ios/Podfile.lock
aymanenadi: No, they are not intentional. I reverted the cocoapods version to `1.11.3`, and did a full re-install. We're not pinning the corresponding libraries to specific versions, which is I believe these pods got bumped up.

PR #119, opened 2023-04-06
PHC-4529 - Integrate track-tile-ui into TilesList

namick:
## Changes

  - Your changes...

## Screenshots

Screenshot 2023-04-06 at 1 43 06 PM

### DIFF
--- a/src/components/TrackTile/TrackTile.tsx
+++ b/src/components/TrackTile/TrackTile.tsx
@@ -18,7 +18,8 @@ export interface Styles extends NamedStyles, StylesProp<typeof defaultStyles> {}
 export type TrackTileProps = {
   title?: string;
   background?: ReactNode;
-  onOpenSettings: (valuesContext: TrackerValuesContext) => void;
+  // TODO: Change this back to non-optional when we have a settings screen
+  onOpenSettings?: (valuesContext: TrackerValuesContext) => void;
   onOpenTracker: (metric: Tracker, valuesContext: TrackerValuesContext) => void;
   hideSettingsButton?: boolean;
   icons?: Record<string, React.ComponentType<SvgProps>>;
@@ -51,7 +52,7 @@ export const TrackTile: FC<TrackTileProps> = ({
             <Text style={styles.trackTileHeaderTitle}>{title}</Text>
             {!hideSettingsButton && (
               <OpenSettingsButton
-                onPress={() => onOpenSettings(valuesContext)}
+                onPress={() => onOpenSettings?.(valuesContext)}
               />
             )}
           </View>
--- a/src/components/TrackTile/services/TrackTileService.ts
+++ b/src/components/TrackTile/services/TrackTileService.ts
@@ -139,23 +139,11 @@ export type TrackTileService = {
   readonly patientId?: string;
 
   /**
-   * @description account and project that the patient data will be saved to
-   */
-  readonly datastoreSettings: {
-    readonly account: string;
-    readonly project: string;
-  };
-
-  /**
    * @description account and project that is used for fetching tracker settings. Defaults to **datastoreSettings**
    */
   readonly accountSettings?: {
     readonly account: string;
     readonly project: string;
-    /**
-     * @description used to decide if public trackers should be fetched alongside account trackers
-     */
-    readonly includePublicTrackers: boolean;
   };
 
   fetchTrackers: (includePublic?: boolean) => Promise<Tracker[]>;
--- a/src/components/tiles/TilesList.tsx
+++ b/src/components/tiles/TilesList.tsx
@@ -9,6 +9,7 @@ import { useStyles, useDeveloperConfig } from '../../hooks';
 import { getCustomAppTileComponent } from '../../common/DeveloperConfig';
 import { spacing } from '../BrandConfigProvider/theme/base';
 import { createStyles } from '../BrandConfigProvider';
+import TrackTile from './TrackTile';
 
 export interface TilesListProps {
   tiles?: AppTile[];
@@ -43,6 +44,7 @@ export const TilesList = ({
   return (
     <ScrollView testID={tID('tiles-list')} style={styles.scrollView}>
       <View style={styles.view}>
+        <TrackTile />
         {tiles?.map((appTile) => (
           <Tile
             key={appTile.id}
--- a/src/navigators/HomeStack.tsx
+++ b/src/navigators/HomeStack.tsx
@@ -4,12 +4,14 @@ import { HomeScreen } from '../screens/HomeScreen';
 import { AppTile } from '../hooks/useAppConfig';
 import { AppTileScreen } from '../screens/AppTileScreen';
 import { CustomAppTileScreen } from '../screens/CustomAppTileScreen';
+import { TrackTileTrackerScreen } from '../screens/TrackTileTrackerScreen';
 import { AppNavHeader } from '../components/AppNavHeader';
 
 export type HomeStackParamList = {
   Home: undefined;
   'tiles/AppTile': { appTile: AppTile };
   'tiles/CustomAppTile': { appTile: AppTile };
+  'tiles/TrackTile': { tracker: any; valuesContext: any };
 };
 
 const Stack = createNativeStackNavigator<HomeStackParamList>();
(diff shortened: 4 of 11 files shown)


src/components/tiles/TilesList.tsx
```diff
@@ -43,6 +44,7 @@ export const TilesList = ({
   return (
     <ScrollView testID={tID('tiles-list')} style={styles.scrollView}>
       <View style={styles.view}>
+        <TrackTile />
```
bruce-glazier, line 47: Thinking we should pass the TrackTile into the TileList via the children prop so this list remains fully configurable.

namick: Done.

src/screens/HomeScreen.tsx
```diff
@@ -54,3 +59,28 @@ const styles = StyleSheet.create({
     flex: 1,
   },
 });
+
+function TrackTileWrapper() {
```
bruce-glazier, line 63: Nit: I think you could move this inside HomeScreen since it is already calling the `useAppConfig` hook. Something like:

```
const navigation = useNavigation<HomeScreenNavigation>();
const trackTile = appConfig?.homeTab?.tiles?.includes?.('trackTile') ?  <TrackTile
       onOpenTracker={(tracker, valuesContext) =>
         navigation.navigate('tiles/TrackTile', {
           tracker,
           valuesContext,
         })
       }
       title={appConfig?.homeTab?.trackTileSettings?.title}
     /> : null;
```
and then use it in TilesList:
```
          <TilesList
             TrackTile={trackTile}
             tiles={appConfig?.homeTab?.appTiles}
           />
```

bruce-glazier (approved):

src/components/TrackTile/TrackerDetails/DatePicker.tsx
```diff
@@ -11,10 +11,6 @@ declare module './TrackerDetails' {
   interface Styles extends StylesProp<typeof defaultStyles> {}
 }
 
-declare module './AdvancedTrackerDetails' {
-  interface Styles extends StylesProp<typeof defaultStyles> {}
-}
```
markdlv, line 16: I'm reading through each file and have seen removal of this kind of thing a time or two, but not sure I understand it.  Is creating this module done elsewhere now, or is this more of a general shift to instead be tied to the theme/style provider in the SDK (as evident from the nice simplification in the `TrackTileProvider` file above).

src/components/TrackTile/TrackerDetails/to-observation-resource.ts
```diff
@@ -44,7 +44,7 @@ const toFhirObservationResource = (
     meta: {
       tag: [
         {
-          code: datastoreSettings.project,
+          code: accountSettings?.project || '',
```
markdlv, line 47: What's the idea with this change?  Assuming `to-observation-resource` is always called right before sending to the cloud, it doesn't seem like we'd want to allow for an empty string here.  We know that'd cause a 400, so seems like we should just not send the request / get to this method in those cases.

src/screens/HomeScreen.tsx
```diff
@@ -54,3 +59,28 @@ const styles = StyleSheet.create({
...
+  const navigation = useNavigation<HomeScreenNavigation>();
+  const { data: appConfig } = useAppConfig();
+
+  // TODO: Push this logic to the useAppConfig hook
```
markdlv, line 67: RE: this TODO: I'd like the `useAppConfig` hook to stay dumb / API-driven if possible.  If there are props like these we need on top of the API response, we should consider creating a hook that's layered on top of `useAppConfig`.  In this case, it could be something like `useTrackTileConfig()`.  If the props are only used in one place, IMO the hook isn't needed and it's fine for the props to just be created where they're used.

src/screens/HomeScreen.tsx
markdlv: What might be nice (for a subsequent PR) is to shift all the messiness into TilesList, whereas right now both HomeScreen and TilesList feel a bit awkward. For example, maybe refactor `TilesList` to take in all of `appConfig?.homeTab` and destructure the bits it needs (e.g. `{ appTiles, tiles, ... }` etc.).  In my mind, this would leave `HomeScreen` purely showing `TilesList` and `TilesList` containing all the temporary messiness we can refactor over time.  We can eventually have TilesList host a cleaner factory for different tile types, add tile-ordering, etc. etc.  If HomeScreen is reasoning about tile types, refactoring for cleanliness will be more difficult.

markdlv (approved):
Let's get this in and iterate on it

src/screens/HomeScreen.tsx
namick: I normally don't like defining a component inside another component, but in this case, we are really just configuring it and it does make sense 👍

src/components/TrackTile/TrackTile.tsx
```diff
@@ -18,7 +18,8 @@ export interface Styles extends NamedStyles, StylesProp<typeof defaultStyles> {}
 export type TrackTileProps = {
   title?: string;
   background?: ReactNode;
-  onOpenSettings: (valuesContext: TrackerValuesContext) => void;
+  // TODO: Change this back to non-optional when we have a settings screen
```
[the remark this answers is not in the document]
namick: Correct.

src/screens/HomeScreen.tsx
namick: I like this idea a lot!   Will do.

src/components/TrackTile/TrackerDetails/to-observation-resource.ts
namick: Good catch.  This change was added last minute when I was trying to get through some typing errors. I meant to get back to it.  Didn't realize an empty string would cause an observation to fail either though... 😨

src/components/TrackTile/TrackerDetails/DatePicker.tsx
namick: I removed it because it was causing type errors and isn't needed.  It is part of the old system of generating types for custom user styles and is being replaced by our updated system in the SDK

src/components/TrackTile/TrackerDetails/to-observation-resource.test.ts
```diff
@@ -9,10 +9,10 @@ const createDate = new Date('2021-09-21T12:00:00.000');
 const defaultResource = (resource: Partial<ResourceSettings> = {}) => ({
   ...resource,
   createDate,
-  datastoreSettings: {
+  accountSettings: {
     account: 'datastore-account',
     project: 'id',
-    ...resource.datastoreSettings,
+    ...resource.accountSettings,
```
sternetj, line 15: Probably could just remove this.

src/components/TrackTile/TrackerDetails/to-procedure-resource.test.ts
```diff
@@ -13,10 +13,10 @@ const defaultResource = (
   ...resource,
   createDate,
   id: '1',
-  datastoreSettings: {
+  accountSettings: {
     account: 'datastore-account',
     project: 'id',
-    ...resource.datastoreSettings,
+    ...resource.accountSettings,
```
sternetj, line 19: Probably could just remove this.

src/components/TrackTile/services/TrackTileService.ts
```diff
@@ -139,23 +139,11 @@ export type TrackTileService = {
...
-  readonly datastoreSettings: {
-    readonly account: string;
-    readonly project: string;
-  };
-
-  /**
-   * @description account and project that is used for fetching tracker settings. Defaults to **datastoreSettings**
+   * @description account and project that is used for fetching tracker settings.
    */
   readonly accountSettings?: {
```
sternetj, line 144: Might be nice to remove this and hoist account and project up to the service level.

src/components/tiles/TilesList.tsx
```diff
@@ -53,6 +55,7 @@ export const TilesList = ({
 
   return (
     <ScrollView testID={tID('tiles-list')} style={styles.scrollView}>
+      {TrackTile && TrackTile}
```
sternetj, line 58: Don't need the `&& TrackTile`

src/navigators/HomeStack.tsx
```diff
@@ -27,6 +29,7 @@ export function HomeStack() {
         name="tiles/CustomAppTile"
         component={CustomAppTileScreen}
       />
+      <Stack.Screen name="tiles/TrackTile" component={TrackTileTrackerScreen} />
```
sternetj, line 32: Don't forget to handle loading the the `AdvancedTrackerScreen` in a later PR

sternetj (approved):

PR #120, opened 2023-04-06
PHC-4328 Wearables PR feedback

bruce-glazier:
## Changes

  - Consolidate hooks
  - Wait for account headers
  - Add types for meta
  - Copy over translations 
  - Using theme spacing


src/hooks/useWearables.ts
```diff
@@ -110,5 +72,54 @@ export const useSetWearableState = () => {
...
+            headers: accountHeaders,
+          })
+          .then((res) => res.data),
+      {
+        enabled: !!accountHeaders,
+      },
+    );
+
+  return {
+    setWearableState,
```
bruce-glazier, line 119: I attempted to utilize useMutation here but it would lead to quite a bit of refactor in the Wearables components so I decided to punt on that for now.

src/hooks/useWearables.ts
```diff
@@ -110,5 +72,54 @@ export const useSetWearableState = () => {
...
+    useQuery(
+      'get-wearable',
+      () =>
+        httpClient.get(`/v1/wearables/${ehrId}`, { headers: accountHeaders }),
+      {
+        enabled: !!accountHeaders,
+      },
+    );
+
+  const useWearableIntegrationsQuery = (appId: string) =>
```
sternetj, line 103: I feel like we should just use the bundleId as the `appId` and remove this argument. This should make things consistent and avoid some confusion on what value should be provided.

sternetj (dismissed):

src/hooks/useWearables.ts
bruce-glazier: Ahh, yes, I will do that in a follow up PR. 👍

src/components/BrandConfigProvider/theme/ThemeProvider.tsx
```diff
@@ -26,7 +26,7 @@ export type Theme = typeof combinedDefaultTheme & MD3Theme;
 
 export type ThemeProp = RecursivePartial<Theme>;
 
-export const useTheme = () => usePaperTheme<Theme>();
+export const useTheme = () => usePaperTheme<Theme>(combinedDefaultTheme);
```
bruce-glazier, line 29: This is why spacing wasn't available by default.

src/hooks/useWearables.ts
bruce-glazier: Done

sternetj (dismissed):

src/hooks/useWearables.ts
markdlv: 👍 In the fhir client where I'm utilizing mutation, I'm still figuring out how to make error handling actually work.  I also find myself consistently utilizing `mutateAsync`.  So the async func approach you have in place might be altogether better for now.

src/components/BrandConfigProvider/theme/ThemeProvider.tsx
```diff
@@ -21,6 +21,7 @@ const combinedDefaultTheme = merge(
   MD3DefaultTheme,
   baseDefaultTheme,
 );
+Object.assign(DefaultTheme, combinedDefaultTheme); // Set a new "default" theme
```
markdlv, line 24: Hopefully @namick sees / comments on this before merging, to make sure we're not going in the wrong direction here.  Makes sense to me.

markdlv (dismissed):

src/components/BrandConfigProvider/theme/ThemeProvider.tsx
namick: Good offline conversation with @bruce-glazier, with one minor change given that DefaultTheme and MD3LightTheme are[ actually the same](https://github.com/callstack/react-native-paper/blob/main/src/core/theming.tsx#L19) theme - we're all on the same page and this is the right direction!

namick (approved):
LGTM

src/components/BrandConfigProvider/theme/ThemeProvider.tsx
```diff
@@ -1,28 +1,25 @@
...
 const combinedDefaultTheme = merge(
-  {},
-  reactNavigationDefault,
   MD3DefaultTheme,
+  reactNavigationDefault,
   baseDefaultTheme,
 );
 
-export type Theme = typeof combinedDefaultTheme & MD3Theme;
+export type Theme = typeof combinedDefaultTheme;
```
bruce-glazier, line 22: The type of MD3DefaultTheme is MD3Theme which is already included in combinedDefaultTheme so this was redundant.

namick (approved):

PR #121, opened 2023-04-06
Add useFhirClient and example app custom native screen demoing it

markdlv:
https://user-images.githubusercontent.com/1445395/230505593-7df635e1-25ad-494e-9212-45d42366dd80.mp4


bruce-glazier (approved):

src/hooks/useFhirClient.tsx
```diff
@@ -0,0 +1,130 @@
...
+  id: string;
+  resourceType: keyof ResourceTypes;
+};
+
+export function useFhirClient() {
+  const { httpClient } = useHttpClient();
+  const { accountHeaders } = useActiveAccount();
+  const { activeProject, activeSubjectId } = useActiveProject();
+
+  const useSearchResourcesQuery = (queryParams: QueryParams) => {
```
sternetj, line 29: Should we use the lazy query [pattern](https://tanstack.com/query/v4/docs/react/guides/disabling-queries#lazy-queries) instead of returning a new function every time? Then we could wrap these in callbacks and not make them their own hooks.

sternetj (approved):

src/hooks/useFhirClient.tsx
markdlv: Yes, absolutely.  As mentioned [here](https://github.com/lifeomic/react-native-sdk/pull/113#discussion_r1158936177), I started with these as callbacks as-is.  Everything worked fine except lint, which pointed out that's a fundamental [violation](https://reactjs.org/link/invalid-hook-call) of rules of hooks.  But yea, let's definitely iterate on this over time to find the right balance of where the `useQuery` goes, how its consumed upstream (directly or indirectly), etc.

PR #127, opened 2023-04-12
Add support for a simpleTheme prop that takes in 2 colors

markdlv:
This really opens the door to seeing which theme color props we're using and which we aren't (e.g. `primaryContainer`), and which components are utilizing theme props and which aren't yet.

This is (obviously) not fully baked in regards to components utilizing the simpleTheme as documented in @JayMoore75's doc.  This PR focuses primarily on the interface for providing a simple theme, and how it's consumed in `RootProviders`.  I considered consuming `useDeveloperConfig` from areas like `ThemeProvider` - but I think I like keeping those components dumb and not tied to DeveloperConfig.  The disadvantage is if a dev re-creates RootProviders, they'll need to also re-create `useDeveloperConfig` consumption.  I'm ok with this and think it matches how `DeveloperConfig` primarily targets devs/apps that _want_ to utilize `RootProviders` for its simplicity.  If we continue down that path, we can get better about documenting where each dev config prop is consumed.

https://user-images.githubusercontent.com/1445395/231596864-42466e20-76f1-4424-82bf-677f13f69945.mp4

### DIFF
--- a/src/common/RootProviders.tsx
+++ b/src/common/RootProviders.tsx
@@ -13,6 +13,7 @@ import Toast from 'react-native-toast-message';
 import { NoInternetToastProvider } from '../hooks/NoInternetToastProvider';
 import { BrandConfigProvider } from '../components/BrandConfigProvider';
 import { TrackTileProvider } from '../components/TrackTile/TrackTileProvider';
+import { useDeveloperConfig } from '../hooks/useDeveloperConfig';
 
 const queryClient = new QueryClient();
 
@@ -23,6 +24,8 @@ export function RootProviders({
   authConfig: AuthConfiguration;
   children?: React.ReactNode;
 }) {
+  const { theme } = useDeveloperConfig();
+
   return (
     <QueryClientProvider client={queryClient}>
       <AuthContextProvider>
@@ -32,7 +35,7 @@ export function RootProviders({
               <ActiveAccountContextProvider>
                 <ActiveProjectContextProvider>
                   <TrackTileProvider>
-                    <BrandConfigProvider>
+                    <BrandConfigProvider theme={theme}>
                       <NoInternetToastProvider>
                         <SafeAreaProvider>
                           <ThemedNavigationContainer>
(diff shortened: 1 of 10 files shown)


bruce-glazier (dismissed):

src/components/DeveloperConfigProvider.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+    };
+    const { result } = await renderHookInContext({
+      simpleTheme,
+    });
+    expect(result.current.theme).toEqual({
+      colors: {
+        primary: simpleTheme.primaryColor,
+        primaryContainer: simpleTheme.accentColor,
+      },
+    });
```
namick, line 49: What is the thinking in mapping `simpleTheme.accentColor` to `primaryContainer` rather than to the `secondary` accent color?.  I believe primaryContainer is derived from primary "for less prominent components" or elements that need "less emphasis".

I could see it the way you have it as well though, because simpleTheme's accent is derived from simpleTheme's primary.  However, if we want simpleTheme's second color to be used as an accent, then we might want it on secondary...

src/common/RootProviders.tsx
```diff
@@ -22,6 +23,9 @@ export function RootProviders({
   authConfig: AuthConfiguration;
   children?: React.ReactNode;
 }) {
+  const { theme: customTheme } = useDeveloperConfig();
+  const theme = customTheme ?? {};
```
sternetj, line 27: The `theme` provided to `BrandConfigProvider` can be undefined so this isn't necessary.

sternetj (dismissed):

src/components/DeveloperConfigProvider.test.tsx
namick: Sorry, meant to approve as well.  Don't want my query here to hold up this PR.

namick (dismissed):

sternetj (approved):

PR #130, opened 2023-04-13
chore: Use individual account attributes

namick:
## Changes

  - Use individual account and project attributes

### DIFF
--- a/src/components/TrackTile/TrackerDetails/to-observation-resource.ts
+++ b/src/components/TrackTile/TrackerDetails/to-observation-resource.ts
@@ -11,7 +11,7 @@ const toFhirObservationResource = (
   code?: Code,
 ): TrackerResource => {
   const { tracker, value, createDate } = from;
-  const { id, accountSettings, patientId } = from;
+  const { id, projectId, patientId } = from;
   const { system, name: metricName } = tracker;
   const metricId = tracker.metricId ?? tracker.id;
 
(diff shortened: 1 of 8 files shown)


src/components/TrackTile/TrackerDetails/to-observation-resource.ts
```diff
@@ -44,7 +44,7 @@ const toFhirObservationResource = (
     meta: {
       tag: [
         {
-          code: accountSettings?.project || '',
```
namick, line 47: This addresses feedback about a possible 400 when submitting an observation with an empty string.

src/components/TrackTile/hooks/useAxiosTrackTileService.ts
```diff
@@ -41,10 +41,8 @@ export const useAxiosTrackTileService = (): TrackTileService => {
   const { activeSubjectId: patientId, activeProject } = useActiveProject();
   const { account } = useActiveAccount();
 
-  const accountSettings = {
-    account: account?.id ?? '',
-    project: activeProject?.id ?? '',
-  };
```
namick, line 47: However, I'm realizing the issue is pushed back up to here.  Not sure if we need to do anything though because we should be protected by isLoading at this point.

This is also in track-tile-ui [here](https://github.com/lifeomic/life-research-mobile/blob/master/src/hooks/useTrackTileService.tsx#L17).

PR #131, opened 2023-04-13
fix: Add the account header when registering a device

aymanenadi:


markdlv (approved):
Do we not have a consumer calling `registerDeviceToken` in example yet?  I expected the addl param to cause TS/lint issues for any consumers.

aymanenadi:
> Do we not have a consumer calling `registerDeviceToken` in example yet? I expected the addl param to cause TS/lint issues for any consumers.

Not yet, but I'm working on an example BMA app to test the notifications E2E. The plan is to re-use some of it as examples in the sdk and the starter packages.

PR #133, opened 2023-04-14
Fix i18n imports

bruce-glazier:
## Changes

Imports from `@i18n` are not working in react-native-starter


bruce-glazier:
Nope, this isn't a good idea, we want to be able to import from @i18n. Closing.

PR #135, opened 2023-04-14
fix: Use relative paths for i18n imports

bruce-glazier:
## Changes

  - Make it so we no longer need to add an alias to the metro configuration

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -18,8 +18,6 @@
     "typeRoots": ["./node_modules/@types", "src/typings"],
     "paths": {
       "@styles": ["./src/components/BrandConfigProvider/styles/types"],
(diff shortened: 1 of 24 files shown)


sternetj (dismissed):
Should we remove it from the `tsconfig.json` as well?

tsconfig.json
```diff
@@ -18,8 +18,6 @@
     "typeRoots": ["./node_modules/@types", "src/typings"],
     "paths": {
       "@styles": ["./src/components/BrandConfigProvider/styles/types"],
```
bruce-glazier, line 20: This is required when creating the defaultStyles types in a component.

sternetj (approved):

PR #136, opened 2023-04-14
chore: Add TrackTile settings

namick:
## Changes

  - Add TrackTile settings

## Screenshots

Screenshot 2023-04-14 at 5 35 52 PM

### DIFF
--- a/src/navigators/HomeStack.tsx
+++ b/src/navigators/HomeStack.tsx
@@ -6,12 +6,18 @@ import { AppTileScreen } from '../screens/AppTileScreen';
 import { CustomAppTileScreen } from '../screens/CustomAppTileScreen';
 import { TrackTileTrackerScreen } from '../screens/TrackTileTrackerScreen';
 import { AppNavHeader } from '../components/AppNavHeader';
+import { TrackerValuesContext } from '../components/TrackTile/main';
+import TrackTileSettingsScreen from '../screens/TrackTileSettingsScreen';
+import { t } from '@i18n';
 
 export type HomeStackParamList = {
   Home: undefined;
   'tiles/AppTile': { appTile: AppTile };
   'tiles/CustomAppTile': { appTile: AppTile };
   'tiles/TrackTile': { tracker: any; valuesContext: any };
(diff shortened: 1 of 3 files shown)


src/navigators/HomeStack.tsx
```diff
@@ -6,12 +6,18 @@ import { AppTileScreen } from '../screens/AppTileScreen';
...
+import { TrackerValuesContext } from '../components/TrackTile/main';
+import TrackTileSettingsScreen from '../screens/TrackTileSettingsScreen';
+import { t } from '@i18n';
 
 export type HomeStackParamList = {
   Home: undefined;
   'tiles/AppTile': { appTile: AppTile };
   'tiles/CustomAppTile': { appTile: AppTile };
   'tiles/TrackTile': { tracker: any; valuesContext: any };
+  'tiles/trackTileSettings': {
```
sternetj, line 18: The casing is different than the others. Do we want to make that consistent or are we switching to snack-case?

src/screens/TrackTileSettingsScreen.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import { NativeStackScreenProps } from '@react-navigation/native-stack';
+import { HomeStackParamList } from '../navigators/HomeStack';
+import { ManageTrackers } from '../components/TrackTile/ManageTrackers/ManageTrackers';
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+  },
+});
```
sternetj, line 12: Should this use are style system?

sternetj (approved):

src/screens/TrackTileSettingsScreen.tsx
namick: Yep, coming up in the next PR

src/navigators/HomeStack.tsx
namick: Good catch.  Looking at this now, are we attempting to imbue semantic meaning into these route names?  Maybe we should have them align with their stack location.

PR #137, opened 2023-04-17
feat: Allow providing a custom base URL

aymanenadi:
## Changes
  - Allow providing a custom base URL to enable working against non-production environments.

### DIFF
--- a/src/common/DeveloperConfig.ts
+++ b/src/common/DeveloperConfig.ts
@@ -14,10 +14,13 @@ import { AppTile } from '../hooks/useAppConfig';
  * when an app tile with a matching URL is tapped.
  *
  * @param simpleTheme Allows for configuring a theme via a primary and accent color.
+ *
+ * @param baseURL Allows for configuring a custom base API URL
  */
 export type DeveloperConfig = {
   appTileScreens?: AppTileScreens;
   simpleTheme?: SimpleTheme;
+  baseURL?: string;
 };
 
 export type SimpleTheme = {
--- a/src/common/RootProviders.tsx
+++ b/src/common/RootProviders.tsx
@@ -24,13 +24,13 @@ export function RootProviders({
   authConfig: AuthConfiguration;
   children?: React.ReactNode;
 }) {
-  const { theme } = useDeveloperConfig();
+  const { theme, baseURL } = useDeveloperConfig();
 
   return (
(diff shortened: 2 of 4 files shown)


namick (approved):

aymanenadi:
It appears we're still having an issue with `i18`: `Error: src/navigators/HomeStack.tsx(11,19): error TS2307: Cannot find module '@i18n' or its corresponding type declarations.`

namick:
> It appears we still have an issue with `i18`: `Error: src/navigators/HomeStack.tsx(11,19): error TS2307: Cannot find module '@i18n' or its corresponding type declarations.`

I ran into that as well.  I think it is from a faulty merge.  Fixed it in this PR: https://github.com/lifeomic/react-native-sdk/pull/138

But you can make that change as well:
```
- import { t } from '@i18n';
+ import { t } from '../../lib/i18n';
```

aymanenadi:
> > It appears we still have an issue with `i18`: `Error: src/navigators/HomeStack.tsx(11,19): error TS2307: Cannot find module '@i18n' or its corresponding type declarations.`
> 
> I ran into that as well. I think it is from a faulty merge. Fixed it in this PR: #138
> 
> But you can make that change as well:
> 
> ```
> - import { t } from '@i18n';
> + import { t } from '../../lib/i18n';
> ```

Thank you. I'll add that.

src/common/DeveloperConfig.ts
```diff
@@ -14,10 +14,13 @@ import { AppTile } from '../hooks/useAppConfig';
  * when an app tile with a matching URL is tapped.
  *
  * @param simpleTheme Allows for configuring a theme via a primary and accent color.
+ *
+ * @param baseURL Allows for configuring a custom base API URL
  */
 export type DeveloperConfig = {
   appTileScreens?: AppTileScreens;
   simpleTheme?: SimpleTheme;
+  baseURL?: string;
```
markdlv, line 23: Big picture, this looks great.

A nit-pick - in this very-high-level interface context, the name is a little vague.  Base URL for what, the web apps that launch from app tiles?  I could see this being `apiBaseURL` instead, even though it's indeed called `baseURL` for the two providers that take it.  

I could also see more hinting in the description about how this is probably not needed unless performing advanced debugging involving a dev mock server, or something to that effect.

markdlv (approved):

PR #138, opened 2023-04-17
chore: Standardize route names based on stack

namick:
This addresses some feedback and fixes an incorrect import resulting (I think) from a merge error

## Changes

  - Fix import
  - Update route names to reflect their stack location

### DIFF
--- a/src/navigators/HomeStack.tsx
+++ b/src/navigators/HomeStack.tsx
@@ -8,14 +8,14 @@ import { TrackTileTrackerScreen } from '../screens/TrackTileTrackerScreen';
 import { AppNavHeader } from '../components/AppNavHeader';
 import { TrackerValuesContext } from '../components/TrackTile/main';
 import TrackTileSettingsScreen from '../screens/TrackTileSettingsScreen';
-import { t } from '@i18n';
+import { t } from '../../lib/i18n';
 
 export type HomeStackParamList = {
   Home: undefined;
-  'tiles/AppTile': { appTile: AppTile };
-  'tiles/CustomAppTile': { appTile: AppTile };
-  'tiles/TrackTile': { tracker: any; valuesContext: any };
-  'tiles/trackTileSettings': {
+  'Home/AppTile': { appTile: AppTile };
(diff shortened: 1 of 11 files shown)


src/navigators/HomeStack.tsx
```diff
@@ -8,14 +8,14 @@ import { TrackTileTrackerScreen } from '../screens/TrackTileTrackerScreen';
 import { AppNavHeader } from '../components/AppNavHeader';
 import { TrackerValuesContext } from '../components/TrackTile/main';
 import TrackTileSettingsScreen from '../screens/TrackTileSettingsScreen';
-import { t } from '@i18n';
+import { t } from '../../lib/i18n';
 
 export type HomeStackParamList = {
```
namick, line 13: Fixed import

bruce-glazier (approved):
👍

src/navigators/HomeStack.tsx
```diff
@@ -8,14 +8,14 @@ import { TrackTileTrackerScreen } from '../screens/TrackTileTrackerScreen';
...
   Home: undefined;
-  'tiles/AppTile': { appTile: AppTile };
-  'tiles/CustomAppTile': { appTile: AppTile };
-  'tiles/TrackTile': { tracker: any; valuesContext: any };
-  'tiles/trackTileSettings': {
+  'Home/AppTile': { appTile: AppTile };
+  'Home/CustomAppTile': { appTile: AppTile };
+  'Home/TrackTile': { tracker: any; valuesContext: any };
+  'Home/TrackTileSettings': {
```
markdlv, line 18: Initial gut reaction is I didn't think I liked this - thought these belong "to the TilesList".  But indeed, the `TilesList` could be used across several stacks.. so the prefix should represent (if anything) the stack it's in.  I could see an argument for not having the prefix at all, but if `/TrackTile` were also in the Settings stack, for example, these prefixes allow us to be specific about which one we want to navigate to.  I dig it!

markdlv (approved):

PR #139, opened 2023-04-17
PHC-5429 - Update TrackTile style

namick:
## Changes

  - Updated TrackTile style

## Screenshots

Screenshot 2023-04-17 at 6 56 52 PM
Screenshot 2023-04-17 at 6 57 04 PM
Screenshot 2023-04-17 at 6 57 16 PM
Screenshot 2023-04-17 at 6 57 27 PM
Screenshot 2023-04-17 at 6 57 38 PM

# before
Screenshot 2023-04-17 at 6 59 35 PM

### DIFF
--- a/example/storybook/stories/TrackTile/TrackTiles.stories.tsx
+++ b/example/storybook/stories/TrackTile/TrackTiles.stories.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 
 import { storiesOf } from '@storybook/react-native';
 import { MockEnvironmentDecorator } from './util/MockEnvironmentDecorator';
-import { View } from 'react-native';
 import { action } from '@storybook/addon-actions';
 import { TrackTile } from 'src/components/TrackTile';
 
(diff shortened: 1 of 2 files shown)


markdlv (approved):
Some of the screenshots look improved, but at the same time not great.  Like how for some only the title shows, and others there's a black background.  Is this a checkpoint PR - more coming?

namick:
> Some of the screenshots look improved, but at the same time not great. Like how for some only the title shows, and others there's a black background. Is this a checkpoint PR - more coming?

I have the (sub)components separated in different PRs here, yes, I should have mentioned that this isn't complete.

namick:
> I wonder if we should consider ditching the settings button for the time being. I know when it was first added we wanted it to avoid the extra row adding a lot of vertical spacing which is what the card layout ends up adding. I have seen cases where the settings button is added at the end of the scrollable list. Maybe we could do that or move it to the settings tab.

I'm not sure if we want the settings button there by default or not, but I do know that the vertical spacing, all the spacing really, isn't needed. We can have different defaults.  

The goal here is to have as little default settings as possible.  This reduces about 25 lines of default style to just three.

PR #141, opened 2023-04-18
fix: PHC-4644 Fix Track Tile Tracker NaNs

sternetj:
## Changes

  - Fixed issue where missing value was converted to NaN
  - Added safe guards in convert function to assume 0 if value is undefined
  - Updated `MockEnvironmentDecorator` to use the mocked axios instance for stories


markdlv (dismissed):
Whether in this PR or another, might consider proving the NaN is resolved via a unit test.  I know you've got the storybook in place, but nice to have something protect against regression at the PR/build level

markdlv (approved):

PR #145, opened 2023-04-19
Export NotificationsStack

aymanenadi:


cmjarreau (approved):

evanheisler (approved):

src/navigators/index.ts
```diff
@@ -2,3 +2,4 @@ export * from './HomeStack';
 export * from './RootStack';
 export * from './SettingsStack';
 export * from './TabNavigator';
+export * from './NotificationsStack';
```
markdlv, line 5: Ah darn didn't catch it in time - please try and keep these alphabetical if possible.  I know these index files are annoying, but we should try to keep them as clean as possible until we have a better solution in place.

PR #149, opened 2023-04-20
PHC-5429 - feat: Final TrackTile in TileList tweaks

namick:
## Changes

  - Adjusted TrackTile spacing to tighten it up to match PillarsTile
  - Removed SafeAreaView as it is causing issues and we already have safe area protection in react-navigation header and tab navigation

### before

https://user-images.githubusercontent.com/220893/233506469-6a55130d-bb88-4656-bd57-5c3b1dc33b37.mov

### after

https://user-images.githubusercontent.com/220893/233506493-727dd0ae-5aef-46b0-8b6c-36efb9f73593.mov

### DIFF
--- a/src/components/TrackTile/TrackTile.tsx
+++ b/src/components/TrackTile/TrackTile.tsx
@@ -58,6 +58,7 @@ export function TrackTile({
         titleStyle={styles.titleText}
         right={settingsButton}
         rightStyle={styles.settingsButton}
+        style={styles.title}
       />
       <Card.Content style={styles.content}>
         <TrackerRow
@@ -73,14 +74,16 @@ export function TrackTile({
 }
 
 const defaultStyles = createStyles('TrackTile', (theme) => ({
-  titleText: {
-    margin: theme.spacing.medium,
+  title: {
+    marginVertical: -8,
   },
+  titleText: {},
   card: {
     margin: theme.spacing.large,
   },
   settingsButton: {
     marginRight: theme.spacing.medium,
+    marginTop: -8,
   },
   content: {},
 }));
--- a/src/components/TrackTile/TrackerRow/Tracker.tsx
+++ b/src/components/TrackTile/TrackerRow/Tracker.tsx
@@ -101,7 +101,7 @@ const defaultStyles = createStyles('Tracker', (theme) => ({
     textAlign: 'center',
   },
   tracker: {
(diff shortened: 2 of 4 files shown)


src/components/TrackTile/TrackTile.tsx
```diff
@@ -73,14 +74,16 @@ export function TrackTile({
 }
 
 const defaultStyles = createStyles('TrackTile', (theme) => ({
-  titleText: {
-    margin: theme.spacing.medium,
+  title: {
+    marginVertical: -8,
   },
+  titleText: {},
```
sternetj, line 80: I feel like I have seen this in a couple of places. Not in this PR but I feel like it would be good to add `title` to our list of `TextStyle` type prefixes and then switch any `title` styles to be `titleContainer` or something like that. `titleText` just reads kind of clunky to me.

sternetj (approved):
Looks good. In the video the trackers looked centered. Is that how they appear until more trackers are added and then it becomes left aligned?

namick:
> Looks good. In the video the trackers looked centered. Is that how they appear until more trackers are added and then it becomes left aligned?

Correct, left aligned with side scroll.


https://user-images.githubusercontent.com/220893/233645527-c5c33e9f-3af2-4c54-a410-15a87aa9344a.mov

src/components/TrackTile/TrackTile.tsx
namick: That is a good thought. I wonder how intuitive `title` would be that it automatically becomes a `TextStyle` type though.  I like that we have sufixes where `somethingText` becomes a `TextStyle` type and `somethingView` becomes a `ViewStyle` type.  Same with `somethingImage` and `ImageStyle`.  These map directly.  The default being that if it doesn't end with any of those three it is a `ViewStyle`.  Adding more rules than that for common use cases like `title` might get confusing, what about `header` or `name` for instance.  Could we accomplish the same thing by getting more specific in our naming when there is ambiguity?  I could just change the above to be `titleView` and `titleText` for example.

markdlv (approved):

src/components/TrackTile/TrackTile.tsx
sternetj: Good points, let's just stick with it for now then 👍

PR #150, opened 2023-04-21
feat: PHC-4574 Add Wearable LifeCycle Hooks and HealthKit UI elements

sternetj:
## Changes

  - Adds a Wearable Lifecycle Provider for creating lifecycle handlers to handle native device functionality
  - Adds UI elements for Apple Health


src/components/Wearables/WearableLifecycleProvider.tsx
```diff
@@ -0,0 +1,152 @@
...
+  onPreToggle: (
+    wearable: WearableIntegration,
+    enabled: boolean,
+  ) => Promise<void>;
+  onPostToggle: (wearable: WearableIntegration) => Promise<void>;
+  sanitizeEHRs: (
+    ehrs: WearableIntegration[],
+    enableMultiWearable?: boolean,
+    legacySort?: boolean,
+  ) => Promise<WearableIntegration[]>;
```
sternetj, line 29: These lifecycle methods were the only ones needed by HealthKit but we could add to these to support other native functionality or customers as needed.

sternetj:
I need to make some tweaks to this so don't review it just yet.

src/common/init.ts
```diff
@@ -9,5 +9,10 @@ export function initializeI18Next() {
 }
 
 export const init = () => {
+  try {
+    require('../../../../../plugins');
```
sternetj, line 13: There is probably a better way to do this so please let me know if you can think of one. We want to load plugins if they exist at `<projectRoot>/plugins`. When this file in the module is installed it is _likely_ located at `<projectRoot>/node_modules/@lifeomic/react-native-sdk/src/common/init.js`

markdlv (approved):
I left some offline comments for the plugins route, but big picture LGTM

bruce-glazier (approved):

PR #151, opened 2023-04-24
feat: Update primary color

namick:
## Changes

  - Update primary color for Tracker History Chart (aka Pillar Detail Graph Color)
  - Replace gradients with primary color for Quick Navigate Tile Chevron Background
  - Update primary color for Tracker Tile Activity Bar

## Screenshots

Screenshot 2023-04-24 at 12 18 20 PM

Screenshot 2023-04-24 at 12 18 30 PM


src/components/TrackTile/TrackerRow/Tracker.tsx
```diff
@@ -43,7 +45,7 @@ export function Tracker(tracker: TrackerProps) {
         <View style={styles.trackerProgressBarContainer}>
           <RadialProgress
             disabled={disabled}
-            color={tracker.color}
+            color={theme.colors.primary}
```
sternetj, line 48: Just want to double check and make sure that we want all tracker progress indicators to use the primary theme color.

src/components/tiles/Tile.tsx
```diff
@@ -5,43 +5,22 @@ import { useStyles } from '../../hooks/useStyles';
...
-  chevronGradient?: LinearGradientProps;
   children?: React.ReactNode;
   Icon?: React.FC<SvgProps>;
   id?: string;
   onPress?: () => void;
   testID?: string;
-  tileGradient?: LinearGradientProps;
 }
 
-const defaultTileGradient: LinearGradientProps = {
```
sternetj, line 24: @bruce-glazier recently added these. Are we sure we want to remove them?

sternetj (dismissed):

src/components/tiles/Tile.tsx
bruce-glazier: We had gradients where this was ported over from so seemed like something we wanted here too but I get that it doesn't fit the "simple theme" concept we are going for here.

markdlv: Long-term seems like we want to enable either, but I dig this as a short-term change to enable simpleTheme.  One question I have is whether `LinearGradient` is used elsewhere?  AFAIK it's not, so `react-native-linear-gradient` should be removed so that there's one less native iOS pod being packaged with the app.  It also looks like `styles.gradient` could be removed?  Those should ideally be removed in this PR.  

Or, alternatively, I wonder about undoing all the changes in this file and instead making it so `defaultChevronGradient` gets colors of `[primaryColor, primaryColor]`?  This would require moving defaultChevronGradient to be set within the component, because it'd need the theme.

namick: Oh, yea, I removed them because the new PMA doc specifically stated that gradients would not be an option.

I'll remove `react-native-linear-gradient` as well.  We can do without the bloat for now and it will be easy to add back whenever we need it.

bruce-glazier (approved):

PR #154, opened 2023-04-26
Add developer toggle to switch environments at runtime

bruce-glazier:
## Changes

  - Add developer toggle to switch target environment at runtime
  - Only appears when `__dev__` is set, i.e., running in metro. 
  - Stores result in async storage and persists between runs.
  - **Note**: Requires a manual refresh to clear the cache after switching environments

## Screenshots

https://user-images.githubusercontent.com/76954025/234664534-06502d59-9649-42bf-ba5f-2b627e5ff434.mp4


example/.env.example
```diff
@@ -3,4 +3,13 @@ OAUTH_REDIRECT_URL='com.lifeomic.mobile-example://loginCallback'
...
+API_BASE_URL='https://api.us.lifeomic.com'
 DEMO_APP='false'
+
+# In dev mode there is a toggle on the settings screen to switch back and forth from the above to the below
+# OAUTH_CLIENT_ID_SECONDARY='36snajbcbt8tmg2ir5r5k1rdpp'
+# OAUTH_REDIRECT_URL_SECONDARY='com.lifeomic.mobile-example://loginCallback'
+# OAUTH_AUTHORIZATION_ENDPOINT_SECONDARY='https://lifeomic-prod-us.auth.us-east-2.amazoncognito.com/logout'
+# OAUTH_TOKEN_ENDPOINT_SECONDARY='https://lifeomic-prod-us.auth.us-east-2.amazoncognito.com/oauth2/token'
+# OAUTH_REVOKE_ENDPOINT_SECONDARY='https://lifeomic-prod-us.auth.us-east-2.amazoncognito.com/oauth2/revoke'
+# API_BASE_URL_SECONDARY='https://api.us.lifeomic.com'
```
markdlv, line 15: Want to grab this newline?

src/hooks/index.ts
```diff
@@ -16,3 +16,4 @@ export * from './useStyles';
 export * from './useTheme';
 export * from './useUser';
 export * from './useWearables';
+export * from './useEnvironmentToggle';
```
markdlv, line 19: Would you mind alphabetizing this?

src/screens/SettingsScreen.tsx
```diff
@@ -46,6 +47,12 @@ export const SettingsScreen = () => {
             action={() => navigate('Settings/Wearables')}
           />
           <Divider />
+          {__DEV__ && process.env.NODE_ENV !== 'test' && (
+            <>
+              <ToggleEnvironmentMenuItem />
+              <Divider />
+            </>
+          )}
```
markdlv, line 55: Whether instead of this or in addition to it, might consider putting this on the LoginScreen so devs can switch envs without having to login.

markdlv (dismissed):

example/.env.example
bruce-glazier: Done

src/screens/SettingsScreen.tsx
bruce-glazier: It is a good idea but I'm going to kick that down the road for now since I already built the component for the settings screen.

markdlv (approved):

src/hooks/useEnvironmentToggle.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { NativeEventEmitter } from 'react-native';
+
+export const useEnvironmentToggle = () => {
+  const [usePrimaryEnv, setUsePrimaryEnv] = useState(true);
+
+  useEffect(() => {
+    const eventEmitter = new NativeEventEmitter({
+      addListener: () => {},
+      removeListeners: () => {},
+    });
```
sternetj, line 12: Should this happen outside of the hook and have it be exported so that `ToggleEnvironmentMenuItem.tsx` can use that reference instead of making its own new instance?

sternetj (approved):

PR #158, opened 2023-04-26
feat: Generate color theme from a single color

namick:
Note:  I'm treating this as a feature branch and will be merging others into it.  We won't merge this until all others have been approved.

## Changes

  -  Added a function to generate all theme colors from a single color
  -  Updated Story to show this ability


src/components/BrandConfigProvider/theme/generateColors.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  elevationLevels.forEach((level, i) => {
+    elevations[`level${i + 1}`] = Color(surface)
+      .mix(Color(primary), Number(level))
+      .rgb()
+      .string();
+  });
+
+  return elevations;
+};
```
namick, line 56: Two parts of the theme that are required for RN Paper but are not a part of Material Design 3...

This is the first:

```TS
elevation: {
    level0: 'transparent',
    level1: 'rgb(238, 246, 246)',
    level2: 'rgb(230, 241, 242)',
    level3: 'rgb(223, 237, 238)',
    level4: 'rgb(220, 235, 237)',
    level5: 'rgb(215, 232, 234)',
}
```

src/components/BrandConfigProvider/theme/generateColors.ts
```diff
@@ -0,0 +1,82 @@
...
+    .alpha(0.4)
+    .rgb()
+    .string();
+
+  return {
+    surfaceDisabled,
+    onSurfaceDisabled,
+    backdrop,
+  };
+};
```
namick, line 82: And these tokens with alpha channels:

```TS
 surfaceDisabled: 'rgba(25, 28, 29, 0.12)',
 onSurfaceDisabled: 'rgba(25, 28, 29, 0.38)',
 backdrop: 'rgba(41, 50, 52, 0.4)',
```

jest.config.js
```diff
@@ -28,7 +28,9 @@ module.exports = {
     'src/common/RootProviders',
     'src/common/LoggedInProviders',
     'src/common/ThemedNavigationContainer',
-    'src/navigators', 'src/components/TrackTile'
+    'src/navigators', 'src/components/TrackTile',
+    'src/screens',
```
markdlv, line 32: Why are `screens` being added?  Feel free to drop coverage thresholds instead.  The tricky thing about this exclusion list is that it's easy to forget about; it'd be good to keep screens revealing it's coverage

namick: I didn't exclude `src/screens` temporarily.  I don't think there is generally much logic in screens and therefore don't require tests.  My thought was that this coverage metric could be more accurate if it reflected only areas that generally give good ROI with testing.

Happy to add it back however.

markdlv (dismissed):

markdlv (approved):

PR #160, opened 2023-04-27
chore: Move environment toggle to login screen

bruce-glazier:
## Changes

  - Rebuild environment toggle using SegmentedButtons
  - Move toggle to LoginScreen for easier use
  - Improve logic to ensure selection stays synchronized across component and hook

## Screenshots

https://user-images.githubusercontent.com/76954025/234974404-7d854aed-d224-4269-88e2-258b54ac98ac.mp4


src/components/EnvironmentSelection.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const [selectedButton, setSelectedButton] = useState('primary');
+  const { styles } = useStyles(defaultStyles);
+
+  const selectEnvironment = useCallback(
+    (value: string) => {
+      setStoredEnvironment(value);
+      setSelectedButton(value);
+      Alert.alert(
+        `Switched to ${value} environment`,
+        'To ensure new content is loaded perform a manual refresh of the app now.',
```
sternetj, line 27: Might just consider doing the refresh here if this is absolutely necessary: https://reactnative.dev/docs/devsettings#reload

src/hooks/useEnvironmentSelection.tsx
```diff
@@ -0,0 +1,31 @@
...
+  removeListeners: () => {},
+});
+
+const environmentToggledEvent = 'environmentToggledEvent';
+
+export const emitEnvironmentToggled = (value: string) =>
+  eventEmitter.emit(environmentToggledEvent, value);
+
+export const useEnvironmentSelection = () => {
+  const [usePrimaryEnvironment, setUsePrimaryEnvironment] = useState(true);
```
sternetj, line 15: Should this check async storage? If the app is using secondary, I think this would be true on load.

src/components/EnvironmentSelection.tsx
```diff
@@ -0,0 +1,87 @@
...
+  useEffect(() => {
+    if (storedEnvironment.data) {
+      setSelectedButton(storedEnvironment.data);
+    }
+  }, [storedEnvironment.data]);
+
+  // Communicate updated selection to listener
+  // in useEnvironmentSelection hook
+  useEffect(() => {
+    emitEnvironmentToggled(selectedButton);
```
sternetj, line 43: One downside of this is that if this component isn't loaded the environment toggle could be out of sync.

bruce-glazier: I think this is desired behavior, correct me if I am wrong. If the toggle component doesn't load, i.e, we are not running in development mode, we only want to use the primary environment.

src/hooks/useEnvironmentSelection.tsx
bruce-glazier: It will get automatically overwritten when the useEnvironmentSelection hook loads so it may appear in the primary position for a split second. It was originally reading from async storage but I decided that a single read from the hook would be sufficient for synchronizing the selection.

src/components/EnvironmentSelection.tsx
bruce-glazier: Added auto refresh (which I think is fine since this is a dev toggle).

sternetj (approved):

src/components/EnvironmentSelection.tsx
bruce-glazier: Ohhh, right. You've got an excellent point there. I have to look closer at that case.

bruce-glazier: Merging as-is for now and will update for this case in another PR.

PR #161, opened 2023-04-27
PHC-4621 - Add primarySource color

namick:
## Changes

  - Add `primarySource` color to be the fixed reference to the original brand source color

### DIFF
--- a/src/components/BrandConfigProvider/theme/base/default.ts
+++ b/src/components/BrandConfigProvider/theme/base/default.ts
@@ -3,12 +3,15 @@
  *
  * The base color scheme uses a popular design system to
  * dynamically generate all color tokens from the HTML color:
- * "gray" - #808080	- rgb(128,128,128)
+ * "gray" - #808080	- rgb(128,128,128) - This is available in the
+ * theme colors object as primarySource.
  *
  * We are also using this file as a test fixture to make sure
  * the generateColors function works perfectly.
  */
 export const colors = {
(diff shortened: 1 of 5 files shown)


markdlv (approved):

sternetj (approved):
Seems okay but I am a little unclear as to what `source` means in the context of the color. Could you explain what the distinction is?

namick:
> Seems okay but I am a little unclear as to what `source` means in the context of the color. Could you explain what the distinction is?

Yes, in the MD3 color system, the color palettes are generated from a key source color.    This one color becomes thirteen "tones", which are luminance from absolute black to absolute white.

Screenshot 2023-04-28 at 11 46 59 AM

So in `light` mode, the theme maps `colors.primary` to `primary40`.  This is often very close to the actual source brand color used to generate the theme, but sometimes not, depending on the luminance of the source color.

Screenshot 2023-04-28 at 12 01 12 PM

This allows us to have a reference to the actual source color used to generate the theme.  So for instance, we want to use this primary source color in many places to match exactly with the brand styleguide.  However, we don't want to use it in text for instance because we need to guarantee a certain contrast ratio for usability and accessibility.

I named it `primarySource` so that we can also have `secondarySource` and `tertiarySource` when/if we want to support a greater amount of theme generation control.

bruce-glazier (approved):

sternetj:
@namick Thanks for the explanation. Sounds great!

PR #162, opened 2023-04-28
PHC-4621 - Update style and surfaces toward new design

namick:
## Changes

  - Add elevated light style to the header and tab navigation
  - Smaller and more relaxed title typography
  - Fixed issue with nested ScrollView in both TileList and HomeScreen.  Replaced with new ScreenSurface.
  - New ScreenSurface component.   This is meant to be the base of any scrolling screen like notifications or settings, but just focused on the home screen in this PR
  - Tweaked Pillar and TrackTile styles to be more like PMA
  - Updated to our createStyles/useStyles system in a couple components where it made sense as I was adjusting default styles

## Screenshots

# before
https://user-images.githubusercontent.com/220893/235037731-03dc8419-3ecd-4845-b9a0-70a9c32d3db0.mov

# after

https://user-images.githubusercontent.com/220893/235037776-98c5d8e6-a8d1-409e-b75d-579542b7a3df.mov

### DIFF
--- a/src/components/AppNavHeader.tsx
+++ b/src/components/AppNavHeader.tsx
@@ -1,9 +1,10 @@
 import React from 'react';
 
-import { Appbar } from 'react-native-paper';
+import { Appbar, Text } from 'react-native-paper';
 import { NativeStackHeaderProps } from '@react-navigation/native-stack';
 import { createStyles } from './BrandConfigProvider';
 import { useStyles } from '../hooks/useStyles';
+import { TextStyle } from 'react-native';
 
 export function AppNavHeader({
   back,
@@ -15,21 +16,35 @@ export function AppNavHeader({
   const title = options.title || route.name;
 
   return (
-    <Appbar.Header style={styles.style}>
+    <Appbar.Header style={styles.style} elevated>
       {back ? (
         <Appbar.BackAction
           onPress={navigation.goBack}
           style={styles.backAction}
         />
       ) : null}
-      <Appbar.Content title={title} style={styles.content} />
+      <Appbar.Content
+        title={<Title text={title} style={styles.titleText} />}
+        style={styles.content}
+      />
     </Appbar.Header>
   );
 }
 
-const defaultStyles = createStyles('AppNavHeader', () => ({
-  style: {},
+const Title = ({ text, style }: { text: string; style?: TextStyle }) => (
+  <Text variant="titleMedium" style={style}>
+    {text}
+  </Text>
+);
+
+const defaultStyles = createStyles('AppNavHeader', (theme) => ({
+  style: {
+    backgroundColor: '#fff',
+  },
   content: {},
+  titleText: {
+    color: theme.colors.onSurfaceVariant,
+  },
   backAction: {},
 }));
 
--- a/src/components/ScreenSurface.tsx
+++ b/src/components/ScreenSurface.tsx
@@ -1,0 +1,46 @@
+import React from 'react';
+
+import { Surface } from 'react-native-paper';
+import { createStyles } from './BrandConfigProvider';
+import { useStyles } from '../hooks/useStyles';
+import { ScrollView, View } from 'react-native';
+
+type Props = {
+  children?: React.ReactNode;
+  styles?: ScreenSurfaceStyles;
+} & React.ComponentProps<typeof Surface>;
+
+export function ScreenSurface({
+  children,
+  styles: instanceStyles,
+  ...props
+}: Props) {
+  const { styles } = useStyles(defaultStyles, instanceStyles);
+
+  return (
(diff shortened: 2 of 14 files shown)


src/components/TrackTile/TrackTile.test.tsx
```diff
@@ -94,7 +94,11 @@ describe('Track Tile', () => {
 
     const { findByLabelText } = render(
       <TrackTileProvider>
-        <TrackTile onOpenSettings={onOpenSettings} onOpenTracker={jest.fn()} />
+        <TrackTile
+          onOpenSettings={onOpenSettings}
+          onOpenTracker={jest.fn()}
+          title="Test Title"
+        />
       </TrackTileProvider>,
```
namick, line 102: This test alerted me to the fact that the settings icon/button on the TrackTile is now dependent on their being a `title` prop.  Both of these features are not yet tested in this PR and is still an open question as to how they should look when enabled.  I think @sternetj might have had an idea about how to do this.  In any case it will get it's own subsequent PR.

src/components/tiles/Tile.tsx
```diff
@@ -80,16 +80,10 @@ const defaultStyles = createStyles('Tile', (theme) => ({
...
-      width: 0,
-    },
-    shadowRadius: 8,
-    marginHorizontal: theme.spacing.large,
-    marginBottom: theme.spacing.extraSmall,
-  },
+    marginHorizontal: theme.spacing.medium,
+    marginBottom: theme.spacing.medium,
+    ...shadow(2),
+  } as ViewStyle,
```
namick, line 86: Super cool `shadow` function from Paper that takes an elevation.   Have to look into enabling our Type system with it in a subsequent PR.

src/components/AppNavHeader.tsx
```diff
@@ -15,21 +16,35 @@ export function AppNavHeader({
...
-  style: {},
+const Title = ({ text, style }: { text: string; style?: TextStyle }) => (
+  <Text variant="titleMedium" style={style}>
+    {text}
+  </Text>
+);
+
+const defaultStyles = createStyles('AppNavHeader', (theme) => ({
+  style: {
+    backgroundColor: '#fff',
```
sternetj, line 42: This should maybe come from the theme?

src/components/ScreenSurface.tsx
```diff
@@ -0,0 +1,46 @@
...
+export function ScreenSurface({
+  children,
+  styles: instanceStyles,
+  ...props
+}: Props) {
+  const { styles } = useStyles(defaultStyles, instanceStyles);
+
+  return (
+    <Surface style={styles.surface} {...props}>
+      <ScrollView overScrollMode="always" showsVerticalScrollIndicator={false}>
```
sternetj, line 22: Do we want `overScrollMode` on `android` to be always?

src/navigators/TabNavigator.tsx
```diff
@@ -45,3 +59,18 @@ export function TabNavigator() {
     </Tab.Navigator>
   );
 }
+
+const defaultStyles = createStyles('TabNavigator', (_theme) => ({
```
sternetj, line 63: If `_theme` isn't used you could just do `()` instead

sternetj (approved):

src/components/AppNavHeader.tsx
namick: Yea it should.  However, the theme doesn't have a full white or full black specifically because each token needs to map to a different lightness (tone) of the same color depending on if it is in light or dark mode.  The closest it has is `background` which maps to `neutral99` in light mode - one step lightness away from pure white.  But I haven't set up translucent headers yet (another PR) and so you can see that band of slightly tinted white vs pure white in the notch area.

This will be changed to `theme.colors.background` in that PR.

src/components/ScreenSurface.tsx
namick: I think so?  Not really sure.

src/navigators/TabNavigator.tsx
```diff
@@ -45,3 +59,18 @@ export function TabNavigator() {
...
+  barStyle: {
+    backgroundColor: '#fff',
```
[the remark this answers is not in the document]
namick: Yes, see comment above.

src/components/ScreenSurface.tsx
```diff
@@ -0,0 +1,46 @@
...
+        <View style={styles.container}>{children}</View>
+      </ScrollView>
+    </Surface>
+  );
+}
+
+const defaultStyles = createStyles('ScreenView', (theme) => ({
```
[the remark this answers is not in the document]
namick: Good catch!

PR #163, opened 2023-04-28
Refactor navigation types

bruce-glazier:
### Up for consideration: Refactor navigation types and pull all types into a dedicated types file.
See https://reactnavigation.org/docs/typescript#organizing-types for inspiration.

This has the benefit of truly enforcing that navigators can only belong on the appropriate screens. This is not the case when using the useNavigation hook with a ParamList, i.e., any ParamList is allowed on any screen.

For example when setup correctly we specify the`HomeStackScreenProps` for the **HomeScreen**:
image

If we intentionally specify the wrong ScreenProps from the **HomeScreen** we will be met with a TS error:
image

We also get the benefit of reducing boilerplate on screens where the useNavigation hook was used, e.g,:
image


src/navigators/types.tsx
```diff
@@ -0,0 +1,78 @@
...
+    StackScreenProps<HomeStackParamList, T>,
+    RootStackScreenProps<keyof RootStackParamList>
+  >;
+
+export type HomeStackParamList = {
+  Home: undefined;
+  'Home/AppTile': { appTile: AppTile };
+  'Home/AuthedAppTile': { appTile: AppTile };
+  'Home/CustomAppTile': { appTile: AppTile };
+  'Home/TrackTile': { tracker: any; valuesContext: any };
```
sternetj, line 42: Can we use `Tracker` and `TrackerValuesContext` instead?

sternetj (approved):

sternetj (approved):

PR #165, opened 2023-04-28
test: Fix wearable story

namick:
## Changes

  -  Fix wearable story


sternetj (approved):
Thank you! I can't believe I missed that one 🤦‍♂️

sternetj (changes requested):
I think there are a couple other instances of enableMultiWearable in this file as well

namick:
> I think there are a couple other instances of enableMultiWearable in this file as well

Ha, well that is the only one that broke `yarn test`

sternetj (approved):

PR #171, opened 2023-05-02
PHC-4753 Circle Discussions cleanup and tests

bruce-glazier:
## Changes

  - Adds loading indicator when paging down
  - Moves circle tiles below app tiles
  - Uses button text for circle name if available otherwise defaults to circle name
  - Adds new `Any` style naming option
  - Adds unit tests
  - Adds default icon for circle tiles
  - Some renames and general cleanup

### Example app

<div style={{flex: 1, flex-direction: 'row' }}>

</div>

### Starter app

<div style={{flex: 1, flex-direction: 'row' }}>

</div>

### DIFF
--- a/src/components/BrandConfigProvider/styles/createStyles.ts
+++ b/src/components/BrandConfigProvider/styles/createStyles.ts
@@ -37,6 +37,8 @@ export type NamedStyles<T> = {
     ? TextStyle
     : P extends `${string}${IgnoreCapitals<ImageSuffixes>}`
     ? ImageStyle
+    : P extends `${string}${IgnoreCapitals<AnySuffix>}`
+    ? any
     : ViewStyle;
 };
 
@@ -51,6 +53,8 @@ type IgnoreCapitals<T> = T extends string
 type TextSuffixes = 'Text' | 'Label';
 
 type ImageSuffixes = 'Image';
+
+type AnySuffix = 'Any';
 
 type StylesBuilder<T> = (theme: Theme) => T;
 
--- a/src/hooks/useInfinitePosts.tsx
+++ b/src/hooks/useInfinitePosts.tsx
@@ -1,9 +1,13 @@
-import { useInfiniteQuery } from 'react-query';
+import { InfiniteData, useInfiniteQuery } from 'react-query';
 import { gql } from 'graphql-request';
 import { useGraphQLClient } from './useGraphQLClient';
 import { useActiveAccount } from './useActiveAccount';
 
-export function usePosts({ circleId }: { circleId?: string }) {
+interface useInfinitePostsProps {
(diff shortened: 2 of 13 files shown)


src/components/BrandConfigProvider/styles/createStyles.ts
```diff
@@ -52,6 +54,8 @@ type TextSuffixes = 'Text' | 'Label';
 
 type ImageSuffixes = 'Image';
 
+type AnySuffix = 'Any';
```
bruce-glazier, line 57: Cheat code for allowing a child components styles within a parents styles: https://github.com/lifeomic/react-native-sdk/pull/171/files#diff-83f4bf1929726afb337b4ad9f7c588a6a8165205ed8f177be6490e058638dfacR85

Ideally we'd have some magic to simply infer the type from the set property but I'm not sure that is doable (or worth doing).

src/hooks/useInfinitePosts.tsx
```diff
@@ -1,9 +1,13 @@
...
 import { gql } from 'graphql-request';
 import { useGraphQLClient } from './useGraphQLClient';
 import { useActiveAccount } from './useActiveAccount';
 
-export function usePosts({ circleId }: { circleId?: string }) {
+interface useInfinitePostsProps {
+  circleId?: string;
+}
+
+export function useInfinitePosts({ circleId }: useInfinitePostsProps) {
```
bruce-glazier, line 10: Renamed to follow the naming scheme established by react-query.

src/components/BrandConfigProvider/styles/createStyles.ts
namick: Interesting solution.  However, is it really needed?  I have passed styles to child components before without issue.

bruce-glazier: If you want to embed a child component's styles within a parent's default styles, you will need to redefine each style type (e.g., childView, childLabel, etc.) and pass them accordingly. This is because our style types only support basic component types, while the child components contain nested styles. Unless there's another approach I am not considering, this `Any` hack allows us to both define the styles in the parent's default styles and pass them in directly to the child component.

namick: Note that I'm not I'm not actually sure if it is needed or not.  Approving incase I'm wrong.  👍

namick (approved):

src/components/BrandConfigProvider/styles/createStyles.ts
bruce-glazier: I may revisit this to see if I can come up with a better idea in a future PR. Merging as-is for now.

PR #176, opened 2023-05-04
Phc 4775 add basic circle post page

sternetj:
## Changes

  - Adds a Thread View and associated components for viewing a single post thread

## TODO
  - Add tests
  - i18n the text
  - Add loading more
  - Add reactions
  - Add commenting
  - Add replying
  - Add reaction totals
  - Add screen and header component

## Screenshots

image


src/components/BrandConfigProvider/styles/createStyles.ts
```diff
@@ -56,7 +56,7 @@ type ImageSuffixes = 'Image';
 
 type AnySuffix = 'Any';
 
-type StylesBuilder<T> = (theme: Theme) => T;
+type StylesBuilder<T> = (theme: Theme) => NamedStyles<T>;
```
sternetj, line 59: Sorry I didn't catch this sooner but it looks like we weren't getting the style type intellisense when writing styles with `CreateStyles`. Now we should be able to see the following:
image

src/hooks/useCirclePost.ts
```diff
@@ -0,0 +1,139 @@
...
+  replies: {
+    edges: {
+      node: Post;
+    }[];
+    pageInfo: {
+      endCursor?: string;
+      hasNextPage?: boolean;
+    };
+  };
+  circle?: {
```
bruce-glazier, line 58: I _think_ it would be cleaner to only query for fields that have an immediate use and expand as we go to avoid unnecessary bloat but perhaps you already have a use for these fields mapped out.

src/hooks/useCirclePost.ts
```diff
@@ -0,0 +1,139 @@
+import { useCallback } from 'react';
+import { useQuery } from 'react-query';
+import { useActiveAccount } from './useActiveAccount';
+import { gql } from 'graphql-request';
+import { useGraphQLClient } from './useGraphQLClient';
+
+export const useCirclePost = (postId: string, disabled?: boolean) => {
```
bruce-glazier, line 7: Nit: My instinct says that this should just be `usePost` since the post is not necessarily the child of a circle and that the query itself is the `post` query.

src/hooks/useCirclePost.ts
sternetj: Yeah good call. That was an artifact from an earlier iteration. I'll remove it.

bruce-glazier (approved):

src/components/Circles/EmptyComments.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <View style={styles.container}>
+      <MessageSquare
+        stroke={styles?.iconTintColorImage?.tintColor}
+        style={styles.iconImage}
+      />
+
+      <Text style={styles.bodyText}>No comments yet.</Text>
+
+      <Text style={styles.helpText}>Be the first!</Text>
```
markdlv, line 21: Any reason not to put these in i18next from the get-go?

src/components/Circles/PostUnavailable.tsx
```diff
@@ -0,0 +1,58 @@
...
+        style={styles.icon}
+      />
+
+      <Text style={styles.headerText}>Oh No!</Text>
+
+      <Text style={styles.bodyText}>
+        Something went wrong loading this post.
+      </Text>
+
+      <Text style={styles.helpText}>Please try again later.</Text>
```
markdlv, line 24: Same here, should i18n these in this or next PR

src/components/Circles/ThreadPost.tsx
```diff
@@ -0,0 +1,74 @@
...
+          style={styles.avatar}
+          label={initials(post?.author?.profile?.displayName)}
+        />
+        <View>
+          <Text style={styles.usernameText}>
+            {post?.author?.profile?.displayName}
+          </Text>
+          {isValid(created) && (
+            <Text style={styles.responseTimeText}>
+              {formatDistanceToNow(new Date(post?.createdAt!))} ago
```
markdlv, line 34: Whether in this or subsequent - this parameterized phrase needs i18n, e.g. `{{distanceToNow}} ago`

markdlv (approved):

src/components/Circles/EmptyComments.tsx
sternetj: All of the i18n stuff was just laziness on my part. I'll address these right away.

PR #178, opened 2023-05-05
feat: Add useDynamicColorGroup

namick:
This PR adds a `useDynamicColorGroup` hook to encapsulate and memoize generating related colors when needed.

This is used to create a related dark but not black "backdrop" color behind the star rewards on each pillar and to create the hero container color in the TrackerDetails screen.

I did run into a bug and typing issues that I found hard to isolate into their own PRs so they are included, but it shouldn't be too bad.

## Changes

  - Add `useDynamicColorGroup` and test
  - More accurate mocking of material-color-utilities
  - Better theme typing
  - Fixed bug in theme where `merge` mutated imported theme

## Screenshots

| Before | After |
| - | - |
|  image |  image |

| Before | After |
| - | - |
|  image | image |

| Before | After |
| - | - |
|  image |  image |


src/components/BrandConfigProvider/theme/ThemeProvider.tsx
```diff
@@ -9,17 +9,8 @@ import {
...
 } from 'react-native-paper';
-import * as baseDefaultTheme from './base/default';
-
-// Merges baseDefaultTheme and reactNavigationDefault into MD3DefaultTheme
-// which is the default theme returned by usePaperTheme
-const combinedDefaultTheme = merge(
-  MD3DefaultTheme,
-  reactNavigationDefault,
-  baseDefaultTheme,
-);
```
namick, line 20: This was mutating MD3DefaultTheme in an effort to create a combined type.  Among other things, it caused our tests to accidentally work without mocking.  I added the useTheme mock and typing instead.

src/components/TrackTile/PillarsTile/Pillar.tsx
```diff
@@ -268,7 +270,12 @@ export const Pillar: FC<PillarProps> = (props) => {
                   ]}
                 />
                 {hasMetGoal && (
-                  <View style={styles.pillarStarBackground}>
+                  <View
+                    style={[
+                      styles.pillarStarBackground,
+                      { backgroundColor: backdrop },
```
sternetj, line 276: Should this go before `styles.pillarStartBackground` to allow overriding if desired?

src/components/TrackTile/TrackerDetails/TrackerDetails.tsx
```diff
@@ -185,7 +187,12 @@ export const TrackerDetails: FC<TrackerDetailsProps> = (props) => {
           tracker={tracker}
           unit={selectedUnit}
         />
-        <View style={styles.trackerDetailsHeaderContainer}>
+        <View
+          style={[
+            styles.trackerDetailsHeaderContainer,
+            { backgroundColor: colorContainer },
```
sternetj, line 193: Same question

src/hooks/useDynamicColorGroup.tsx
```diff
@@ -0,0 +1,28 @@
...
+    const colorPalette = generateColors(color);
+
+    setColorGroup({
+      color: colorPalette.primary,
+      onColor: colorPalette.onPrimary,
+      colorContainer: colorPalette.primaryContainer,
+      onColorContainer: colorPalette.onPrimaryContainer,
+      backdrop: colorPalette.backdrop,
+    });
+  }, [color]);
```
sternetj, line 25: It would probably be better to use a `useMemo` here unless `generateColors` is an expensive operation. I think this will cause an extra render

sternetj (approved):

src/hooks/useDynamicColorGroup.tsx
namick: Yep, you are correct.  The `useMemo` allowed me to remove the `useState` hook, saving an extra render.

sternetj (approved):

bruce-glazier (approved):

PR #184, opened 2023-05-08
Add additional (optional) navigation tab(s)

cmjarreau:
## Changes
  - add additionalNavTab param to DeveloperConfigProvider and map in TabNavigator

## Screenshots
Screen Shot 2023-05-08 at 2 29 32 PM


src/navigators/types.tsx
```diff
@@ -26,6 +26,8 @@ export type TabParamList = {
   HomeTab: undefined;
   NotificationsTab: undefined;
   SettingsTab: NavigatorScreenParams<SettingsStackParamList>;
+} & {
+  [keyof: string]: undefined;
```
bruce-glazier, line 30: I'm wondering if there is another way to do this where we could preserve these types without adding the equivalent of an any here. I'm not totally sure what that would be so I am approving as is.

bruce-glazier (approved):

bruce-glazier (approved):

PR #189, opened 2023-05-10
feat: PHC-4775 Add Open Thread Screen Logic

sternetj:
## Changes

  - Wires up the discussion screen to open the thread screen when pressed
  - Fixes a path issue
  - Fixes the back arrow for the header (no more X)

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/2295908/f2ec383a-6b53-42fe-8c78-629661ff9d5c

### DIFF
--- a/example/storybook/stories/TrackTile/util/MockEnvironmentDecorator.tsx
+++ b/example/storybook/stories/TrackTile/util/MockEnvironmentDecorator.tsx
@@ -6,13 +6,13 @@ import {
   Tracker,
   TrackTileServiceProvider,
   useAxiosTrackTileService,
-} from '../../../../src/components/TrackTile/main';
+} from '../../../../../src/components/TrackTile/main';
 import {
   FetchTrackerResponse,
   QueryOntologyResponse,
-} from '../../../../src/components/TrackTile/hooks/useAxiosTrackTileService';
+} from '../../../../../src/components/TrackTile/hooks/useAxiosTrackTileService';
 import { nutrition } from './ontologies';
-import { HttpClientContextProvider } from '../../../../src/hooks/useHttpClient';
+import { HttpClientContextProvider } from '../../../../../src/hooks/useHttpClient';
 
 const axiosInstance = axios.create();
 const mock = new MockAdapter(axiosInstance);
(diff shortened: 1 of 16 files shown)


src/components/AppNavHeader.tsx
```diff
@@ -18,7 +19,8 @@ export function AppNavHeader({
   return (
     <Appbar.Header style={styles.style} elevated>
       {back ? (
-        <Appbar.BackAction
+        <Appbar.Action
+          icon={ChevronLeft}
```
bruce-glazier, line 23: 🙌

example/storybook/stories/TrackTile/util/MockEnvironmentDecorator.tsx
```diff
@@ -6,13 +6,13 @@ import {
   Tracker,
   TrackTileServiceProvider,
   useAxiosTrackTileService,
-} from '../../../../src/components/TrackTile/main';
+} from '../../../../../src/components/TrackTile/main';
```
bruce-glazier, line 9: Opps. Thanks for fixing this, seems like there must be something we can do to allow an invalid path from being merged in.

bruce-glazier:
I'm thinking that we should move the activity indicator below the parent post on the thread screen so the UI doesn't jump around so much.

sternetj:
@bruce-glazier I updated the flat list to use two indicators. 1 for the the initial load which shows up beneath the post and another for the refresh control. Another approach might be to add a skeleton loader instead of the activity indicator but figured it would be better to wait on that.

https://github.com/lifeomic/react-native-sdk/assets/2295908/df63582e-6b06-489b-8273-579e5257e35f

src/screens/CircleThreadScreen.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { HomeStackScreenProps } from '../navigators/types';
+import { Thread } from '../components/Circles/Thread';
+
+export const CircleThreadScreen = ({
+  navigation,
+  route,
+}: HomeStackScreenProps<'Home/Circle/Thread'>) => {
+  useLayoutEffect(() => {
+    navigation.setOptions({
+      title: `${route.params.post.author?.profile.displayName}'s Post`,
```
bruce-glazier, line 11: Nit: Could use i18next for this title

bruce-glazier:
> @bruce-glazier I updated the flat list to use two indicators. 1 for the the initial load which shows up beneath the post and another for the refresh control. Another approach might be to add a skeleton loader instead of the activity indicator but figured it would be better to wait on that.
> 
>  thread-loading.indicators.mov

Looks good to me!

bruce-glazier (approved):

PR #190, opened 2023-05-10
PHC-4753 Cleanup discussions screen

bruce-glazier:
## Changes

Implements the following to match ThreadPost
  - Use initials for default avatar icon
  - Adds size styles for avatar icon
  - Formats timestamp the same

Other tweaks:
- Loads i18next keys / init with supported languages
- Stops mocking i18next and instead inits real instance
- Uses plurals on the number of comments indicator (COMMENT / 1 COMMENT / 2 COMMENTS)
- Tweak reaction emoji label as some emojis could be cut short 

## Screenshots

image


src/components/Circles/Post.tsx
```diff
@@ -48,7 +69,7 @@ export const Post = ({ post }: PostProps) => {
           mode={'outlined'}
           onPress={() => {}} // TODO: Navigate to post details/comments page
         >
-          {post.replyCount} COMMENTS
+          {t('post-comments', { count: post.replyCount })}
```
sternetj, line 72: In tests I imagine this would render as `post-comments` based on the jest setup file. Is this what we want/expect?

sternetj (approved):

src/components/Circles/Post.tsx
bruce-glazier: I don't think we have any test coverage aimed specifically at this so it won't cause any problems with testing but perhaps I should look into a way to fix this.

bruce-glazier: On second thought, you are using this in your next PR for testing. Once you merge that in I will rebase that and fix any broken tests.

bruce-glazier: Addressed by unmocking i18next

sternetj (approved):

PR #191, opened 2023-05-11
feat: PHC-4775 Add Thread Replying and Nested Thread Opening

sternetj:
## Changes

  - Add Replying to posts and comments from the thread screen
  - Adds opening a thread from inside a thread
  - Adds autofocus to the new comment/post

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/2295908/5e4f4cc8-8fee-45fa-8797-261ac5dd3fc1

Fix for the comment not loading which is visible in the previous video

https://github.com/lifeomic/react-native-sdk/assets/2295908/9aa089c4-4097-48c4-af44-4343e81d0761

### DIFF
--- a/src/components/Circles/Thread.tsx
+++ b/src/components/Circles/Thread.tsx
@@ -1,12 +1,7 @@
 import React, { useState, useCallback, useRef } from 'react';
-import {
-  FlatList,
-  View,
-  RefreshControl,
-  Text,
-  TouchableOpacity,
-} from 'react-native';
+import { FlatList, View, RefreshControl, TouchableOpacity } from 'react-native';
 import { t } from 'i18next';
+import { Text } from 'react-native-paper';
 import { usePost, Post, useStyles, useTheme, ParentType } from '../../hooks';
 import { ThreadComment } from './ThreadComment';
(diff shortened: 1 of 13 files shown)


bruce-glazier (dismissed):

src/components/Circles/Thread.tsx
```diff
@@ -1,52 +1,124 @@
...
+      />
+
+      <TouchableOpacity
+        style={styles.addCommentBox}
+        onPress={() => setIsCreateCommentOpen(true)}
+      >
+        <Text style={styles.addCommentText}>
+          {t('thread-screen-add-comment', 'Add a comment')}
+        </Text>
+      </TouchableOpacity>
```
namick, line 108: Do we want to use Buttons and Text from react-native-paper in places like this?

I think we want to use Paper components as much as possible to make our own code cleaner and more themeable.

sternetj: Good point. I switched to use RNPaper's Text but left it as touchable opacity since it is kind of a unique button:
image

bruce-glazier (approved):

PR #192, opened 2023-05-11
PHC-4616 Add Circle Announcement Indicator

sternetj:
- Adds circle announcement banner. Went with `Announcement` for the banner name because circle is implied and therefore redundant

https://github.com/lifeomic/react-native-sdk/assets/2295908/a36769c7-1baf-44a6-975e-2358001cf563


src/components/Circles/ThreadPost.tsx
```diff
@@ -20,29 +21,32 @@ export const ThreadPost = ({ post, style }: Props) => {
...
+      {post.priority === Priority.ANNOUNCEMENT && <AnnouncementBanner />}
+      <View style={styles.container}>
+        <View style={styles.header}>
+          <Avatar.Text
+            size={size}
+            style={styles.avatar}
+            label={initials(post?.author?.profile?.displayName)}
+          />
+          <View>
+            <Text style={styles.usernameText}>
```
namick, line 34: Obviously not to do with this PR, but it will eventually be nice to have these all switched over to Paper components so we get consistency across apps built with this SDK.

markdlv (approved):

src/components/Circles/ThreadPost.tsx
sternetj: Yeah for sure. I am trying to be better about that but we will probably need to do a sweep here soon once the circle work is closer to mvp.

PR #195, opened 2023-05-12
chore: PHC-4775 Add Comment Loading

sternetj:
## Changes

  - Adds ability to load in previous replies

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/2295908/298e4a9e-adaa-4db2-aabc-50a1e37ec789

### DIFF
--- a/lib/i18n/locales/en.json
+++ b/lib/i18n/locales/en.json
@@ -61,5 +61,7 @@
   "configure-data-to-sync-message": "Your data is not syncing because it is not configured as a Data Source above.",
   "post-comments_zero": "COMMENT",
   "post-comments_one": "1 COMMENT",
-  "post-comments_other": "{{count}} COMMENTS"
+  "post-comments_other": "{{count}} COMMENTS",
+  "thread-load-comments_one": "Read {{count}} earlier comment...",
+  "thread-load-comments_other": "Read {{count}} earlier comments..."
 }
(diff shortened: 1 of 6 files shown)


namick (approved):
This code looks great by the way.  I am wondering why `useLoadReplies` is added to the `usePosts` file.  Since we are already partitioning the Circles hooks into their own directory under `components/Circles/hooks` is there any reason to combine multiple hooks into a single file?  Personally, I prefer to separate these concepts by file and reduce the file length.  Shared code or gql fragments could be shared by module as well.

Approving in case you have valid reasons for doing it this way.  :-)

sternetj:
> This code looks great by the way. I am wondering why useLoadReplies is added to the usePosts file. Since we are already partitioning the Circles hooks into their own directory under components/Circles/hooks is there any reason to combine multiple hooks into a single file? Personally, I prefer to separate these concepts by file and reduce the file length. Shared code or gql fragments could be shared by module as well.

Yeah great point. Maybe @bruce-glazier can way in but I am open to splitting these up. I think they were initially combined because of the reused `PostDetail` fragment. I think splitting out the graphql fragments would be fine and then splitting them up.

It might be better to do this in a dedicated PR since @bruce-glazier and I are still working in the same area right now.

bruce-glazier (approved):

markdlv (approved):

PR #198, opened 2023-05-16
Allow and detect custom app tiles with bundleId

markdlv:
This makes it so that, if you configure an app tile with a bundle id like `com.lifeomic.components-app://device-config` (in this app's example), and that URL is NOT configured in `developerConfig`, then it'll show an error screen (below) instead of attempting to load the URL in a browser.

Simulator Screen Shot - iPhone 14 Pro - 2023-05-16 at 10 59 14

### DIFF
--- a/src/common/DeveloperConfig.ts
+++ b/src/common/DeveloperConfig.ts
@@ -1,4 +1,5 @@
 import { ComponentType } from 'react';
+import { getBundleId } from 'react-native-device-info';
 import { AppTile } from '../hooks/useAppConfig';
 
 /**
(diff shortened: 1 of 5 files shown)


bruce-glazier (approved):

src/common/DeveloperConfig.ts
```diff
@@ -36,5 +37,9 @@ export function getCustomAppTileComponent(
   appTileScreens?: AppTileScreens,
   appTile?: AppTile,
 ) {
-  return !!appTile?.source.url && appTileScreens?.[appTile.source.url];
+  return (
+    !!appTile?.source.url &&
+    (appTileScreens?.[appTile.source.url] ||
+      appTile.source.url.startsWith(getBundleId()))
+  );
```
sternetj, line 44: So if we have configured a screen for it or if it matches our bundle id, that seems like the later would only apply in a case where we have a misconfiguration. i.e. we have not provided the screen in our developer config but the url matches our bundleId.

sternetj (approved):

PR #200, opened 2023-05-16
chore: Add Edit/Delete replies button to Circles

bruce-glazier:
## Changes

  - Your changes...

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/76954025/409d508a-52f1-4628-bca2-bddc15f76e85


lib/i18n/index.ts
```diff
@@ -38,15 +38,6 @@ export function configure() {
     .init({
       fallbackLng: 'en',
       supportedLngs: ['en', 'es', 'fr', 'ar', 'tr', 'pt', 'de'],
-      resources: {
```
bruce-glazier, line 41: Not sure why but this seemed like it was needed for tests but now it just overwrites the resources loaded from the resource loader plugin.

src/components/Circles/ThreadPost.tsx
bruce-glazier, line 1: This was basically identical to PostItem `formerly Post` so I figured we should just consolidate these into one.

src/components/Circles/PostsList.tsx
```diff
@@ -29,6 +29,7 @@ export const PostsList = ({ circleTile, onOpenPost }: PostsListProps) => {
     useInfinitePosts({
       circleId: circleTile?.circleId,
     });
+  // useFocusEffect(() => {});
```
sternetj, line 32: Is this needed?

sternetj (approved):

src/components/Circles/PostsList.tsx
bruce-glazier: I have a fast follow that removes this almost ready for review.

PR #201, opened 2023-05-16
feat: Surveys + opening urls support

aymanenadi:
https://github.com/lifeomic/react-native-sdk/assets/19395435/872c656c-2690-4aa3-abba-ea5fdac19061

### DIFF
--- a/src/hooks/index.ts
+++ b/src/hooks/index.ts
@@ -19,3 +19,4 @@ export * from './useSubjectProjects';
 export * from './useTheme';
 export * from './useUser';
 export * from './useWearables';
+export * from './useHandleAppTileEvents';
--- a/src/hooks/useHandleAppTileEvents.ts
+++ b/src/hooks/useHandleAppTileEvents.ts
@@ -1,0 +1,108 @@
+import { Linking } from 'react-native';
+import type {
+  WebViewMessageEvent,
+  WebViewNavigation,
+} from 'react-native-webview';
+import { useAppConfig } from './useAppConfig';
+import { useNavigation } from '@react-navigation/native';
+import { HomeStackParamList } from '../navigators/types';
+import { StackNavigationProp } from '@react-navigation/stack';
+
+type NavigationParams = {
+  [x: string]: unknown | undefined;
+  referenceDate?: string | number | Date;
+};
+
+type DeepLinkRouteName = 'tiles/Today/Survey';
+
+export enum AppTileMessageType {
+  deepLink = 'deepLink',
+  openUrl = 'openUrl',
+  openConsent = 'openConsent',
+}
+
+export type DeepLinkDataType = {
+  routeName: DeepLinkRouteName;
+  params: NavigationParams;
+};
+
+export type OpenUrlDataType = {
+  url: string;
+  title?: string;
+  appendUtm?: boolean;
+};
+
+export type DeepLinkAppletMessage = {
+  type: AppTileMessageType.deepLink;
+  data: DeepLinkDataType;
+};
+
+export type OpenUrlAppletMessage = {
+  type: AppTileMessageType.openUrl;
+  data: OpenUrlDataType;
+};
+
+export type AppTileMessage = DeepLinkAppletMessage | OpenUrlAppletMessage;
+
+export const useHandleAppTileEvents = () => {
+  const { data } = useAppConfig();
+  const todayTileSettings = data?.homeTab?.todayTileSettings;
+  const navigation = useNavigation<StackNavigationProp<HomeStackParamList>>();
+
+  const openSurveyAppTile = (responseId: string | undefined) => {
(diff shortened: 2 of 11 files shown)


src/navigators/TabNavigator.tsx
```diff
@@ -31,6 +31,7 @@ export function TabNavigator() {
       <Tab.Screen
         name="HomeTab"
         component={HomeStack}
+        key="HomeTab"
```
aymanenadi, line 34: This is unrelated. I kept getting warnings about missing keys.

src/screens/AuthedAppTileScreen.tsx
```diff
@@ -60,8 +69,36 @@ export const AuthedAppTileScreen = ({
...
+    oauthCallbackUrl,
+    activeSubjectId,
+    searchParams,
+  ]);
+
+  const handlePageLoaded = () => {
+    setIsPageLoaded(true);
+  };
+
+  const dispatchLXFocusEvent = useCallback(() => {
```
sternetj, line 91: It would be nice at some point if we could change `lx` to something more meaningful.

src/hooks/useHandleAppTileEvents.ts
```diff
@@ -0,0 +1,108 @@
...
+      case AppTileMessageType.openUrl:
+        handleOpenUrlMessage(appTileMessage);
+        break;
+
+      // no default
+    }
+  };
+
+  const handleAppTileNavigationStateChange = (event: WebViewNavigation) => {
+    if (event.url.indexOf('pending-lx-action') !== -1) {
```
sternetj, line 100: It would be nice at some point if we could change `lx` to something more meaningful.

sternetj (approved):

evanheisler (approved):
This is pretty much all ported over from the applet, correct?

src/hooks/useHandleAppTileEvents.ts
aymanenadi: I agree. I didn't change it in this PR because of the tight deadlines. We'll need to update all the app tiles that use this route, while still keeping compatibility with the LIFE apps / PMA.

aymanenadi:
> This is pretty much all ported over from the applet, correct?

Do you mean LIFE Extend? For the most part, yes. The main difference is how we create a new code every time a new app tile is opened since the codes are for one-time use only.

evanheisler:
> > This is pretty much all ported over from the applet, correct?
> 
> Do you mean LIFE Extend? For the most part, yes. The main difference is how we create a new code every time a new app tile is opened since the codes are for one-time use only.

Ah, yes thank you. That answers my question then — I was curious what was new code vs. ported over.

src/hooks/useHandleAppTileEvents.ts
```diff
@@ -0,0 +1,108 @@
...
+    const { routeName, params = {} } = appletMessage.data;
+    if (routeName === 'tiles/Today/Survey') {
+      openSurveyAppTile(params.questionnaire as string | undefined);
+    } else {
+      console.warn('Unsupported route name. Ignoring...', { routeName });
+    }
+  };
+
+  const enforceWebURLProtocol = (url: string) => {
+    const allowedProtocols = ['https://', 'http://'];
```
markdlv, line 72: Any reason you know of why we need to support `http://`.  If you're not sure, it'd be great to start this initial migration with `https://` being the only option.

markdlv: Oh, I see - this is used by `handleOpenUrlMessage`, which I'm betting is for program actions that like a youtube URL, article to review, or journal/info website?  Got it if so - if we allow program configuration with either scheme, probably best to keep both allowed here as well.  It's forcibly non-authed URLs here, which is key.

src/hooks/useHandleAppTileEvents.ts
markdlv: +1 on both fronts

src/hooks/index.ts
```diff
@@ -19,3 +19,4 @@ export * from './useSubjectProjects';
 export * from './useTheme';
 export * from './useUser';
 export * from './useWearables';
+export * from './useHandleAppTileEvents';
```
markdlv, line 22: Sorry for the nit but can you alphabetize this?

markdlv (approved):

src/hooks/index.ts
aymanenadi: Of course. I'll do it in a follow-up PR.

src/hooks/useHandleAppTileEvents.ts
aymanenadi: That's correct. This is for URLs entered by the admin in the program builder pointing to articles to read, etc. We do not currently enforce a particular protocol.

PR #204, opened 2023-05-17
Recursively update all cached posts

bruce-glazier:
## Changes

- Fixes issues when modifying nested replies

### DIFF
--- a/src/hooks/Circles/utils/optimisticallyDeletePosts.ts
+++ b/src/hooks/Circles/utils/optimisticallyDeletePosts.ts
@@ -5,6 +5,7 @@ import type {
 } from '../useInfinitePosts';
 import forEach from 'lodash/forEach';
 import remove from 'lodash/remove';
+import { Post } from '../types';
 
 type DeletePostsConfig = {
   queryClient: QueryClient;
@@ -54,10 +55,28 @@ const removeEdgeById = (currentData: InfinitePostsData, id: string) =>
 const removeRepliesById = (
   currentData: PostDetailsPostQueryResponse,
   id: string,
+  depth: number = 7, // set some default max depth
 ) => {
-  const deleted = remove(
-    currentData?.post.replies.edges,
-    (edge) => edge.node.id === id,
-  );
-  currentData.post.replyCount -= deleted.length;
+  let count = 0;
+  const recursiveRemove = (post: Post, currentDepth: number) => {
+    if (currentDepth > depth) {
+      return;
+    }
+
+    const deleted = remove(post.replies.edges, (edge) => edge.node.id === id);
+    post.replyCount -= deleted.length;
+
+    // Recurse into the replies of the remaining edges
+    post.replies.edges.forEach((edge) => {
+      if (edge.node.replies && edge.node.replies.edges) {
+        recursiveRemove(edge.node, currentDepth + 1);
+      }
+    });
+  };
+
+  if (currentData?.post.replies.edges) {
+    recursiveRemove(currentData.post, 0);
+  }
+
+  currentData.post.replyCount -= count;
 };
--- a/src/hooks/Circles/utils/optimisticallyDeletePosts.ts
+++ b/src/hooks/Circles/utils/optimisticallyDeletePosts.ts
@@ -3,9 +3,9 @@ import type {
   InfinitePostsData,
   PostDetailsPostQueryResponse,
 } from '../useInfinitePosts';
-import forEach from 'lodash/forEach';
 import remove from 'lodash/remove';
 import { Post } from '../types';
+import { cloneDeep } from 'lodash';
 
 type DeletePostsConfig = {
   queryClient: QueryClient;
@@ -17,14 +17,24 @@ export const optimisticallyDeletePosts = ({
   postId,
 }: DeletePostsConfig) => {
   queryClient.setQueryData<InfinitePostsData>(['posts'], (currentData) => {
-    removeEdgeById(currentData!, postId);
-    return currentData!;
+    if (!currentData) {
+      return currentData!;
+    }
+    const newData = cloneDeep(currentData);
+    for (const page of newData?.pages) {
+      const result = deletePostById(page.postsV2.edges, postId);
+      if (result) {
+        break;
+      }
+    }
+
+    return newData!;
   });
 
   queryClient
     .getQueryCache()
     .findAll({ queryKey: ['postDetails'] })
-    .map((query) => {
+    .forEach((query) => {
       query?.setData((currentData: PostDetailsPostQueryResponse) => {
         if (!currentData) {
           return currentData!;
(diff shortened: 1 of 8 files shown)


src/hooks/Circles/utils/optimisticallyUpdatePosts.ts
```diff
@@ -4,65 +4,62 @@ import type {
...
-        if (node?.node && visited.indexOf(node.node) === -1) {
-          visited.push(node.node);
-          node.node = updatePost(node.node);
-        }
-
-        return currentData;
-      });
+      query?.setData(
+        (currentData: PostDetailsPostQueryResponse | undefined) => {
+          const newData = cloneDeep(currentData);
```
bruce-glazier, line 35: Note: It appears that mutating currentData directly can result in undesired behavior. In one instance I was seeing a reaction incrementing twice. I'd recommend we always copy the data before modifying it.

src/hooks/Circles/utils/optimisticallyUpdatePosts.ts
```diff
@@ -4,65 +4,62 @@ import type {
...
+) => {
+  if (!edges) {
+    return;
+  }
+  forEach(edges, (edge) => {
+    if (edge.node.id === id) {
+      edge.node = transformFn(edge.node);
+    }
+    if (edge.node.replies?.edges) {
+      transformEdgesById(edge.node.replies?.edges, id, transformFn);
```
bruce-glazier, line 62: Infinite recursion is usually dangerous but the dataset here should never contain a circular reference and be fairly thin.

namick (dismissed):

src/hooks/Circles/utils/optimisticallyUpdatePosts.ts
markdlv: Good call out - can we add some early exit logic just in case?  e.g. Is there a number that'd be obviously too large?

src/hooks/Circles/utils/optimisticallyDeletePosts.ts
```diff
@@ -54,10 +55,28 @@ const removeEdgeById = (currentData: InfinitePostsData, id: string) =>
...
+
+    const deleted = remove(post.replies.edges, (edge) => edge.node.id === id);
+    post.replyCount -= deleted.length;
+
+    // Recurse into the replies of the remaining edges
+    post.replies.edges.forEach((edge) => {
+      if (edge.node.replies && edge.node.replies.edges) {
+        recursiveRemove(edge.node, currentDepth + 1);
+      }
+    });
```
markdlv, line 74: In all for loops / recursion in this PR (including this one), there aren't any network requests inside of these, right?  If so, I think we should definitely be doing this on the backend instead.  Even if not, It feels like the client is having to work pretty hard to keep all of this straight - is there any opportunity (now or down the road) to make this client-side logic simpler (or removed altogether)?

bruce-glazier: This is purely to update the local cache when the user performs a post mutation, i.e., update a post message, delete a post, add a new reaction. This seemingly expensive logic shouldn't be _that_ heavy given it's only iterating a fairly small dataset (whatever is available in the cache locally) and only occurs when such a mutation is issued. That said we could probably make some gains by using store indexes to access the data more directly without having to search for it by id each time.

src/hooks/Circles/utils/optimisticallyUpdatePosts.ts
bruce-glazier: Yeah, we could add some bail out logic just in case. I'll update it.

src/hooks/Circles/utils/optimisticallyUpdatePosts.ts
sternetj: I was kind of wondering about that. Glad you caught it early. From the [docs](https://tanstack.com/query/v4/docs/react/guides/updates-from-mutation-responses#immutability):

> Updates via setQueryData must be performed in an immutable way. DO NOT attempt to write directly to the cache by mutating data (that you retrieved via from the cache) in place. It might work at first but can lead to subtle bugs along the way.

We may want to double check other areas where we do this to make sure we are covered.

src/hooks/Circles/utils/optimisticallyDeletePosts.ts
sternetj: Another optimization is that I don't think we need to recur if `deleted.length > 0` an reply should only exist at one point in the post/reply tree

src/hooks/Circles/utils/optimisticallyDeletePosts.ts
```diff
@@ -54,10 +55,28 @@ const removeEdgeById = (currentData: InfinitePostsData, id: string) =>
...
+  };
+
+  if (currentData?.post.replies.edges) {
+    recursiveRemove(currentData.post, 0);
+  }
+
+  currentData.post.replyCount -= count;
```
sternetj, line 81: How does `count` get incremented? I am not seeing it.

src/hooks/Circles/utils/optimisticallyUpdatePosts.ts
```diff
@@ -4,65 +4,62 @@ import type {
...
-];
-
-const findEdgeById = (currentData: InfinitePostsData, id: string) =>
-  currentData?.pages
-    .map((page) => page.postsV2.edges)
-    .flat()
-    .find(({ node }) => node.id === id) ?? {
-    node: undefined,
-  };
+const transformEdgesById = (
```
sternetj, line 49: It would be nice if we could make this more functional and avoid the mutation

src/hooks/Circles/utils/optimisticallyUpdatePosts.ts
```diff
@@ -4,65 +4,62 @@ import type {
...
+  transformFn: (post: Post) => Post,
+) => {
+  if (!edges) {
+    return;
+  }
+  forEach(edges, (edge) => {
+    if (edge.node.id === id) {
+      edge.node = transformFn(edge.node);
+    }
+    if (edge.node.replies?.edges) {
```
sternetj, line 61: This could be an else if since the id will only exist at one point in the hierarchy.

bruce-glazier: My refactor has addressed this, we now return as soon as we've updated a post.

src/hooks/Circles/utils/optimisticallyDeletePosts.ts
bruce-glazier: That was an artifact of something I had tried earlier, no longer present in this latest commit/refactor.

src/hooks/Circles/utils/optimisticallyUpdatePosts.ts
bruce-glazier: Given this function is not exported and we have made sure to pass a clone of the data into it this should be safe but I do not disagree. However I am going to punt on that for now if that's okay with you.

src/hooks/Circles/utils/optimisticallyDeletePosts.ts
bruce-glazier: Handled by the refactor

src/hooks/Circles/utils/optimisticallyUpdatePosts.ts
bruce-glazier: Done

src/hooks/Circles/utils/optimisticallyUpdatePosts.ts
bruce-glazier: Found one more place in createPost and updated it accordingly.

src/hooks/Circles/utils/optimisticallyDeletePosts.ts
```diff
@@ -34,49 +44,59 @@ export const optimisticallyDeletePosts = ({
...
+  depth: number = 0,
 ) => {
-  let count = 0;
-  const recursiveRemove = (post: Post, currentDepth: number) => {
-    if (currentDepth > depth) {
-      return;
-    }
+  // limit the depth of recursion
+  if (depth > maxDepth) {
+    return true;
```
sternetj, line 62: Should this be `false` so line 94 doesn't get a false positive?

sternetj (approved):

src/hooks/Circles/utils/optimisticallyDeletePosts.ts
bruce-glazier: A true here simply signals that it is time to exit, not that we've succeeded or failed.

src/hooks/Circles/utils/optimisticallyUpdatePosts.ts
sternetj: Makes perfect sense!

sternetj (approved):

PR #210, opened 2023-05-19
feat: Add a consent screen as part of the onboarding process

aymanenadi:
This PR adds logic to render a consent screen if the project has a consent form and if the user did not accept it yet. Declining the consent logs the user out.

https://github.com/lifeomic/react-native-sdk/assets/19395435/1023220a-9ff5-46b5-ac89-50a30e994612


example/package.json
```diff
@@ -33,6 +33,7 @@
     "react-native-draggable-flatlist": "^3.0.6",
     "react-native-gesture-handler": "^2.9.0",
     "react-native-keychain": "^8.1.1",
+    "react-native-markdown-display": "^7.0.0-alpha.2",
```
sternetj, line 36: Should we use 6.x instead of an alpha version?

src/hooks/useConsent.ts
```diff
@@ -0,0 +1,109 @@
...
+      !loadingDirectives &&
+      directivesData &&
+      directivesData.items.length > 0 &&
+      directivesData.items[0].status === 'active';
+
+    const hasConsent = !loadingDefaultConsents && !!defaultConsentData;
+
+    return {
+      isLoading: loadingDirectives || loadingDefaultConsents,
+      shouldRenderConsentScreen: hasConsent && !didAcceptConsent,
```
sternetj, line 99: Should this be
```suggestion
      shouldRenderConsentScreen: hasConsent || !didAcceptConsent,
```

src/hooks/useConsent.ts
```diff
@@ -0,0 +1,109 @@
...
+  const useShouldRenderConsentScreen = () => {
+    const { data: directivesData, isLoading: loadingDirectives } =
+      useConsentDirectives();
+    const { data: defaultConsentData, isLoading: loadingDefaultConsents } =
+      useDefaultConsent();
+
+    const didAcceptConsent =
+      !loadingDirectives &&
+      directivesData &&
+      directivesData.items.length > 0 &&
```
markdlv, line 92: Somewhat nit but also nice for defensiveness on `items` - might simplify 91-92 to something like `!!directivesData?.items?.length`

src/screens/ConsentScreen.tsx
```diff
@@ -0,0 +1,111 @@
...
+          text: t('logout', 'Logout'),
+          style: 'destructive',
+          onPress: async () => {
+            await updateConsentDirective(false);
+            await logout({});
+          },
+        },
+      ],
+    );
+  };
```
markdlv, line 51: Might want to make all three of these methods above utilize `useCallback`

markdlv (approved):
Approving in case all of my & Teddy's comments end up being answered and/or handled in subsequent PRs

src/hooks/useConsent.ts
aymanenadi: No, I should've named `hasConsent` better, but it should be something like `defaultConsentExists`. We only want to render the consent screen if the project has a consent, and the user did not accept it. I'll rename the variable to be more descriptive.

PR #217, opened 2023-05-23
feat: Add pagination support to useFhirClient

aymanenadi:


src/hooks/useFhirClient.tsx
```diff
@@ -40,8 +44,14 @@ export function useFhirClient() {
     // TODO: add code, date, & other query param capabilities
     // TODO: consider using fhir-search across the board (documenting delay)
 
-    return useQuery(
-      [`search-${resourceType}`, params],
+    const fetchNextPage = () => {
+      if (hasMoreData) {
+        setNext((prev) => prev + 10);
```
aymanenadi, line 49: I just realized `pageSize` is an optional parameter. I'll update this to increment `next` by `pageSize`.

src/hooks/useFhirClient.tsx
```diff
@@ -40,8 +44,15 @@ export function useFhirClient() {
     // TODO: add code, date, & other query param capabilities
     // TODO: consider using fhir-search across the board (documenting delay)
 
-    return useQuery(
-      [`search-${resourceType}`, params],
+    const fetchNextPage = () => {
+      if (hasMoreData) {
+        const pageSize = queryParams.pageSize || 10;
+        setNext((prev) => prev + pageSize);
```
sternetj, line 50: Is there a way we could use the url? I think the scroll api gives us a token instead of a page size.

sternetj (dismissed):

src/hooks/useFhirClient.tsx
aymanenadi: I'm not sure I fully understand. `patient-service` returns a url containing a `next` query param which is what dictates which set of elements to return. In the latest commit, `next` is exposed via the hook, and it needs to be passed to `fetchNext` to fetch the next set of data.

src/hooks/useFhirClient.tsx
```diff
@@ -40,8 +47,14 @@ export function useFhirClient() {
     // TODO: add code, date, & other query param capabilities
     // TODO: consider using fhir-search across the board (documenting delay)
 
-    return useQuery(
-      [`search-${resourceType}`, params],
+    const fetchNext = (next: number) => {
+      if (hasMoreData) {
+        setNext(next);
+      }
+    };
```
markdlv, line 54: I'm not following the use-case for this method.  We've got `next` exposed, and we've got it being automatically updated via the useEffect.  But as a consumer, I'd use this to perhaps reset `next`?  That'd definitely come in handy if I was re-using one `useSearchResourcesQuery` across multiple queries (e.g. different resource types or codings).  But shouldn't it be called something like `resetFetch = (next: number = 0) => ...` instead of `fetchNext`?  i.e. I think it's just the naming that's throwing me off.

aymanenadi: > We've got next exposed, and we've got it being automatically updated via the useEffect.

The `useEffect` is an implementation detail so we can expose `next`. It does not effectively sets it. Setting `next` is done when calling `fetchNext`, which triggers the API call.


> But shouldn't it be called something like resetFetch = (next: number = 0) => ... instead of fetchNext? i.e. I think it's just the naming that's throwing me off.

I believe `fetchNext` is more descriptive in the sense that it gives an indication that a new API call is going to be made. Something like `resetFetch` or `resetNext` does not necessarily do that, unless the developer is familiar with the hook's internals and `useQuery`.
I don't really have strong preferences. I can go both ways at this point.

When changing the resource types or the codings, I believe we should automatically reset `next` to 0, and I believe this is something the hook should do. I'll create a separate PR for that.

sternetj: I probably overcomplicated this with my previous comment. I was thinking that `_search` was the `fhir-search` api and in my mind I thought we might get `next.url` links that looked like `url?next=10` or `url?scrollId=asdfasdf` in which case the next parsing logic wouldn't work and it would be better to just use the `next.url` provided to us when making the next query so we could do something like

```
// psuedo code:
const page1 = await axios.post(firstUrl)
const page2 = await axios.get(page1.links.next.url)
const page3 = await axios.get(page2.links.next.url)
```

Sorry for all the confusion on my part.

aymanenadi: @markdlv that's exactly right.

@sternetj no worries.

markdlv (approved):

sternetj (approved):

sternetj (approved):

PR #218, opened 2023-05-23
feat: Add filtering by FHIR codes support to useFhirClient

aymanenadi:


src/hooks/useFhirClient.tsx
```diff
@@ -12,9 +12,15 @@ type ResourceTypes = {
 };
 type ResourceType = Observation;
 
+type FhirCodes = {
```
markdlv, line 15: Can we use the `Coding` type from `fhir/r3` instead?

src/hooks/useFhirClient.tsx
```diff
@@ -12,9 +12,15 @@ type ResourceTypes = {
...
+  system: string;
+  code: string;
+};
+
 type QueryParams = {
   resourceType: keyof ResourceTypes;
   pageSize?: number;
+  coding?: FhirCodes[];
```
markdlv, line 23: Let's use `Coding[]` here (from `fhir/r3`) instead, so that we don't need to create the `FhirCodes` type.

markdlv (approved):

PR #219, opened 2023-05-23
Add Wearable Backfill

sternetj:
## Changes

  - Adds a backfill button to the wearable row when the feature is enabled
  - Adds an onBackfill lifecycle handler hook
  - Fixes typo in fitbit auth text
  - Adds a loading indicator to the ToggleButton

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/2295908/5769e23e-c7f8-4f93-8438-39e36eff9ff0

### DIFF
--- a/src/hooks/useFeature.ts
+++ b/src/hooks/useFeature.ts
@@ -1,0 +1,24 @@
+import { useQuery } from 'react-query';
+import { useActiveAccount } from './useActiveAccount';
+import { useHttpClient } from './useHttpClient';
+
+type Response<T extends string> = Record<T, boolean>;
+
+export function useFeature<T extends string>(feature: T) {
+  const { accountHeaders } = useActiveAccount();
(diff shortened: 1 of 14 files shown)


src/hooks/useFeature.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  return useQuery(
+    ['Features', feature],
+    () =>
+      httpClient.get<Response<T>>(`/v1/features/${feature}`, {
+        headers: accountHeaders,
+      }),
+    {
+      enabled: !!accountHeaders,
+      select(data) {
```
markdlv, line 19: Oo, nice - dig the usage of this method instead of `.then` in the `queryFn`, thereby keeping the original `data` intact in cache.  We should use this more often / throughout - I'm guilty of a lot of `.then`s in here.

PR #222, opened 2023-05-24
PHC-4837 Redesign tracker details

bruce-glazier:
## Changes

  - Tracker details design updates
  - We used to allow for free entry of the target but now we a defined list to selected from
  - Possible values depend on the defined stepAmount for the unit with an upper limit of stepAmount x 50.

## Note
- Looks like I recently introduced some undesired behavior with the counter incrementing/decrementing that will be address in a follow up

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/76954025/d8e16b80-db1c-4b43-9b48-806ea5de862a

Screenshot 2023-05-24 at 2 23 00 PM

Screenshot 2023-05-24 at 2 23 12 PM

### DIFF
--- a/src/components/TrackTile/TrackerDetails/AdvancedTrackerDetails/AdvancedTrackerDetails.test.tsx
+++ b/src/components/TrackTile/TrackerDetails/AdvancedTrackerDetails/AdvancedTrackerDetails.test.tsx
@@ -326,7 +326,13 @@ describe('Tracker Advanced Details', () => {
             id: 'tracker-id',
             metricId: 'metric-id',
             resourceType: 'Observation',
-            units: [{ display: 'Serving of Fruit', target: 5, unit: 'unit' }],
+            units: [
+              {
+                display: 'Servings of Fruit',
+                target: 5,
+                unit: 'Serving of Fruit',
+              },
+            ],
           } as any
         }
         valuesContext={valuesContext}
--- a/src/components/TrackTile/TrackerDetails/NumberPicker.tsx
+++ b/src/components/TrackTile/TrackerDetails/NumberPicker.tsx
@@ -41,9 +41,10 @@ export const NumberPicker = (props: NumberPickerProps) => {
   }, [onChange, tempValue]);
 
   const numbers = [];
-  const stepAmount = selectedUnit.stepAmount ?? 1;
-  const minValue = stepAmount * 1;
-  const maxValue = stepAmount * 50;
+  const stepAmount =
+    selectedUnit.targetStepAmount ?? selectedUnit.stepAmount ?? 1;
+  const minValue = selectedUnit.targetMin ?? 0;
+  const maxValue = selectedUnit.targetStepAmount ?? stepAmount * 50;
 
   for (let i = minValue; i <= maxValue; i += stepAmount) {
     numbers.push(i);
--- a/src/components/TrackTile/TrackerDetails/NumberPicker.tsx
+++ b/src/components/TrackTile/TrackerDetails/NumberPicker.tsx
@@ -1,0 +1,187 @@
+import { Modal, Platform, View, TouchableOpacity } from 'react-native';
+import React, { useCallback, useState } from 'react';
+import { Picker } from '@react-native-picker/picker';
+import { t } from '../../../../lib/i18n';
+import { tID } from '../common/testID';
+import {
+  createStyles,
+  useIcons,
+} from '../../../components/BrandConfigProvider';
+import { Button, Text } from 'react-native-paper';
+import { UnitType } from '../services/TrackTileService';
+import { useStyles } from '../../../hooks';
+
+export type NumberPickerProps = {
+  onChange: (value: string) => void;
+  selectedUnit: UnitType;
+  placeholder?: string;
+  value?: string;
+};
+
+export const NumberPicker = (props: NumberPickerProps) => {
+  const {
+    value = '0',
+    selectedUnit,
+    onChange,
+    placeholder = t('track-tile.set-your-target', 'Set Your Target'),
+  } = props;
+  const { styles } = useStyles(defaultStyles);
+  const [isOpen, setIsOpen] = useState(false);
+  const [tempValue, setTempValue] = useState(value);
+  const { ChevronDown } = useIcons();
+
+  const openInput = useCallback(() => {
+    setIsOpen(true);
+    setTempValue(value);
+  }, [value]);
+
+  const commitNewValue = useCallback(() => {
+    setIsOpen(false);
+    onChange(tempValue);
+  }, [onChange, tempValue]);
+
+  const numbers = [];
+  const stepAmount = selectedUnit.stepAmount ?? 1;
+  const minValue = stepAmount * 1;
+  const maxValue = stepAmount * 50;
+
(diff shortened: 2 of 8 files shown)


src/components/TrackTile/TrackerDetails/AdvancedTrackerDetails/AdvancedTrackerDetails.test.tsx
```diff
@@ -294,7 +294,7 @@ describe('Tracker Advanced Details', () => {
     ).toBeDefined();
   });
 
-  it('calls onEditValue when tapping on a value row', async () => {
+  it.only('calls onEditValue when tapping on a value row', async () => {
```
bruce-glazier, line 297: After much effort to fix this test I've given up for now for the sake of forward progress. In short, Jest cannot interact with the new NumberPicker even though the testId appears in the DOM. I will revisit this.

src/screens/TrackTileTrackerScreen.tsx
```diff
@@ -21,7 +21,7 @@ export const TrackTileTrackerScreen = ({
 
   React.useLayoutEffect(() => {
     navigation.setOptions({
-      headerTitle: t('Track {name}', { name: tracker?.name }),
+      title: t('Track {{name}}', { name: tracker?.name }),
```
bruce-glazier, line 24: The wasn't working at all before.

src/components/TrackTile/TrackerDetails/TrackerDetails.test.tsx
```diff
@@ -81,15 +82,15 @@ describe('Tracker Details', () => {
     });
   });
 
-  it('should upsert the tracker with a new target on unmount', async () => {
+  it.skip('should upsert the tracker with a new target on unmount', async () => {
```
bruce-glazier, line 85: After much effort to fix this test I've given up for now for the sake of forward progress. In short, Jest cannot interact with the new NumberPicker even though the testId appears in the DOM. I will revisit this.

src/components/TrackTile/TrackerDetails/NumberPicker.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  const commitNewValue = useCallback(() => {
+    setIsOpen(false);
+    onChange(tempValue);
+  }, [onChange, tempValue]);
+
+  const numbers = [];
+  const stepAmount = selectedUnit.stepAmount ?? 1;
+  const minValue = stepAmount * 1;
+  const maxValue = stepAmount * 50;
```
bruce-glazier, line 46: This relies on the stepAmount to be a large enough value that the range of selection is wide enough. For minutes, the step amount is 5 so the selection range is 5-250, which may need to be altered to fit the extreme. This allows a selection of just over 4 hours for trackers like self care and exercise but I could see exercise needing to have a higher upper limit.

src/components/TrackTile/TrackerDetails/AdvancedTrackerDetails/AdvancedTrackerDetails.test.tsx
sternetj: Woah this is awesome! I was highlighting them all and writing `it.skip` before

src/components/TrackTile/TrackerDetails/NumberPicker.tsx
markdlv: I'm a little skeptical this will work in all scenarios, and think we should probably take a simpler/dumber approach if possible.  One example that comes to mind is a habit you're trying to break, so we typically want the target to be 0.  The other thing that comes to mind is something like steps where the number is large enough (typical goal of 10k) that the default of 1-50 here will look silly.  Since `stepAmount` and all other props are set on a per-customer, per-metric basis - I'd recommend perhaps attempting these calculations _only_ if more explicit props aren't set (which we'd always recommend setting).  e.g. It could be we need new props (`targetStepAmount`, `targetMin`, `targetMax`), or maybe we have enough props to to reuse?  Either way - I think the preferred way would be 100% config-based, with calculated backups as an edge-case fallback.

bruce-glazier: Thanks, that sounds like a better approach for sure.

bruce-glazier: Done

src/components/TrackTile/TrackerDetails/NumberPicker.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+  const numbers = [];
+  const stepAmount =
+    selectedUnit.targetStepAmount ?? selectedUnit.stepAmount ?? 1;
+  const minValue = selectedUnit.targetMin ?? 0;
+  const maxValue = selectedUnit.targetStepAmount ?? stepAmount * 50;
+
+  for (let i = minValue; i <= maxValue; i += stepAmount) {
+    numbers.push(i);
+  }
```
sternetj, line 51: This might be worth memoizing

src/components/TrackTile/TrackerDetails/TrackerDetails.tsx
sternetj, line 1: I am seeing some inline styles here. Can we move them into defaultStyles?

src/components/TrackTile/TrackerDetails/TrackerDetails.tsx
```diff
@@ -324,3 +304,5 @@ declare module '@styles' {
   interface ComponentStyles
     extends ComponentNamedStyles<typeof defaultStyles> {}
 }
+
+export type TrackTileDetails = NamedStylesProp<typeof defaultStyles>;
```
sternetj, line 308: Should this be `TrackTileDetailsStyles`?

src/components/TrackTile/TrackerDetails/unit-display.ts
```diff
@@ -1,12 +1,12 @@
...
-import { convertToPreferredUnit } from '../util/convert-value';
-import { t } from '../../../../lib/i18n';
+import i18n, { t } from '../../../../lib/i18n';
 
 type UnitDisplayConfig = {
   value: number;
   unit: UnitType;
   tracker: Tracker;
   skipInterpolation?: boolean;
+  preserveCasing?: boolean;
```
sternetj, line 9: This doesn't appear to be used.

src/components/TrackTile/TrackerDetails/unit-display.ts
```diff
@@ -15,22 +15,28 @@ export const unitDisplay = ({
   tracker,
   skipInterpolation = false,
 }: UnitDisplayConfig) => {
-  // This should never happen. If it does it means we are missing translations.
-  const defaultDisplay = `{{count}} ${unit.display.replace(/^(.)/, (match) =>
+  const defaultOne = `{{count}} ${unit.unit.replace(/^(.)/, (match) =>
```
sternetj, line 18: I realize you didn't write this but we could use upperFirst from lodash here which might make it more readable.

src/screens/TrackTileTrackerScreen.tsx
sternetj: Nice fix! That was definitely my bad

sternetj (approved):
Approving incase you want to iterate

PR #224, opened 2023-05-24
feat: Open the circle discussion screen when pressing a POST program activity

aymanenadi:


src/hooks/useHandleAppTileEvents.ts
```diff
@@ -59,12 +59,32 @@ export const useHandleAppTileEvents = () => {
     });
   };
 
+  const openCircleDiscussionScreen = (circleId: string, circleName: string) => {
+    const circleTile: CircleTile = {
+      circleId,
+      circleName,
+      buttonText: circleName,
+      isMember: true,
```
aymanenadi, line 67: I don't believe the `isMember` field is used anywhere. Using `true` for now as the program users should be enrolled in the circle linked to the program activity.

aymanenadi: Happy to delete this field altogether if I can get confirmation that we don't need it anymore.

sternetj: Yeah I don't think we need this

sternetj (approved):

PR #225, opened 2023-05-25
PHC-4837 Tracker fixes

bruce-glazier:
## Changes

  - Fixes bouncing of tracker amounts when adding/subtracting
  - Restores original logic in ManageTrackers where useTrackers is conditionally rendered depending on if trackerRequestMeta is provided.

### DIFF
--- a/src/components/TrackTile/TrackerDetails/TrackAmountControl.tsx
+++ b/src/components/TrackTile/TrackerDetails/TrackAmountControl.tsx
@@ -8,19 +8,32 @@ import { convertToISONumber } from '../util/convert-value';
 import { numberFormatters } from '../formatters';
 import { createStyles } from '../../BrandConfigProvider';
 import { useStyles } from '../../../hooks';
+import { debounce } from 'lodash';
 
 type Props = {
   color: string;
   value: number;
   onChange: (value: number) => void;
+  saveInProgress: boolean;
 };
 const { numberFormat } = numberFormatters;
-const TrackAmountControl: FC<Props> = ({ color, value, onChange }) => {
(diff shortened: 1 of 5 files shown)


src/components/TrackTile/TrackerDetails/TrackAmountControl.tsx
```diff
@@ -87,7 +107,12 @@ const TrackAmountControl: FC<Props> = ({ color, value, onChange }) => {
           'Increment tracker value',
         )}
         accessibilityRole="button"
-        onPress={() => onChange(value + 1)}
+        onPress={() => {
+          const asString = addToNumberString(currentValue, 1, false) as string;
```
bruce-glazier, line 111: Using currentValue rather than an unstable incoming value ensures that we are consistently incrementing the "saved" value rather than it getting trampled when the upsert query resolves.

src/components/TrackTile/TrackerDetails/TrackAmountControl.tsx
```diff
@@ -24,7 +31,13 @@ const TrackAmountControl: FC<Props> = ({ color, value, onChange }) => {
 
   const [currentValue, setCurrentValue] = useState(numberFormat(value));
 
-  useEffect(() => setCurrentValue(numberFormat(value)), [value]);
+  useEffect(() => {
+    if (!saveInProgress) {
+      debounce(() => {
```
bruce-glazier, line 36: I did away with the debounce in my next PR because it caused issues with the initial syncing.

sternetj (approved):

PR #226, opened 2023-05-26
feat: Add LogoHeader and update AppNavHeader

namick:
## Changes

  - Your changes...


bruce-glazier (approved):
This looks good to me!

src/components/AppNavHeader.tsx
```diff
@@ -48,6 +88,7 @@ const defaultStyles = createStyles('AppNavHeader', (theme) => ({
     color: theme.colors.onSurfaceVariant,
   },
   backAction: {},
+  backActionIconAny: {},
```
sternetj, line 91: Not in this PR but I think we need to look at a way to inject props into a component. This kind of style doesn't easily convey to a consumer that they can set the `color` property to modify the component behavior.

src/common/DeveloperConfig.ts
```diff
@@ -23,12 +24,20 @@ import { SvgProps } from 'react-native-svg';
  * @param additionalNavigationTab Allows for configuring a custom additional material
  * bottom navigation tab (in addition to Home, Notifications, and Settings.
  */
+export interface RouteColor {
+  route: string;
+  color: (theme: Theme) => string;
```
sternetj, line 29: Instead of individual colors, I think we would get more mileage out of being able to specify a viewStyle or textStyle per route

sternetj (approved):

src/components/LogoHeader.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+
+import { View, Image, ImageSourcePropType } from 'react-native';
+import { SafeAreaView } from 'react-native-safe-area-context';
+import { createStyles } from './BrandConfigProvider';
+import { useStyles } from '../hooks/useStyles';
+
+interface Props {
+  imageSource: ImageSourcePropType;
```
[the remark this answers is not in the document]
namick: Great idea.

src/components/AppNavHeader.tsx
```diff
@@ -15,25 +17,63 @@ export function AppNavHeader({
...
+const useColorMapping = (
+  property: keyof ViewStyle | keyof TextStyle,
+  routeColors: RouteColor[] | undefined,
+  route: Record<string, any>,
+  style: StyleProp<ViewStyle | TextStyle>,
+) => {
+  const theme = useTheme();
+  const title = route.params?.appTile?.title;
+  const routeColor = routeColors?.find(
+    (item) => item.route === route.name || item.route === title,
```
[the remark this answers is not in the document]
namick: This matches first on route name, then on the appTile title passed in the params.  This is because all of our app tiles have the same route name.  (React Nav 6 uses a ephemeral generated id to keep track of routes internally.)

Previously we matched on app tile id, but this is problematic as they will change when a new app tile is published.

src/components/AppNavHeader.tsx
namick: I agree.  This particular component is a bit tricky because it is passed itself as a prop into (and then rendered by) react navigation.  However, maybe an object of props could be passed into developerConfig and then accessed via the hook.

src/common/DeveloperConfig.ts
namick: Oh, nice idea.  I like.

PR #227, opened 2023-05-26
fix: Remove StyleSheet.create Usages in Favor of createStyles

sternetj:
## Changes

  - Converts all library references of `StyleSheet.create` to use `createStyles` instead
  - Renames the style names to no longer include a namespace identifier (i.e. `trackTileContainer` -> `container`)

## Screenshots

This is generally a noop so there isn't really anything to show. I went through storybook and played around with all the modified component stories and things seemed fine to me.


example/storybook/stories/TrackTile/AdvancedTrackerEditor.stories.tsx
```diff
@@ -43,6 +43,7 @@ storiesOf('AdvancedTrackerEditor', module)
 
     Object.assign(ontology, nutrition);
     let color = '#33C317';
+    let unitCode = "[U'arb]";
```
sternetj, line 46: During testing I noticed this resulted in fractional values for procedures if not defined because of unit conversions

bruce-glazier (approved):
Nice! May as well get this merged in first so I can begin to rebase against it.

PR #228, opened 2023-05-26
PHC-4837 Tracker Bar Graph and Image

bruce-glazier:
## Changes

  - Load image in tracker details screen if URI exists in tracker
  - Change bar graph orientation and other design updates


src/components/TrackTile/TrackerDetails/TrackerHistoryChart/Bar.tsx
```diff
@@ -35,34 +36,38 @@ const Bar: FC<BarProps> = (props) => {
...
-      ]}
-    />
+    <View style={containerStyle}>
+      <Animated.View
+        testID={testID}
+        style={[
+          {
+            width: ref.current.interpolate({
+              inputRange: [0, 1],
+              outputRange: ['0%', '100%'],
```
sternetj, line 46: You might double check this but I think I had this as `-0.1%` because the animation looked better when decreasing a value to 0. I think at 0% it just kind of disappears but the negative value has it shrink down correctly.

src/components/TrackTile/TrackerDetails/TrackerHistoryChart/Chart.tsx
```diff
@@ -160,40 +123,24 @@ export const Chart: FC<ChartProps> = (props) => {
   );
 };
 
-const ChartContent: FC<any> = ({ children, maxTick, hasXAxis }) => {
+const ChartContent: FC<any> = ({ children }) => {
```
sternetj, line 126: You could probably drop this component altogether

src/components/TrackTile/TrackerDetails/TrackerHistoryChart/Chart.tsx
```diff
@@ -46,62 +45,40 @@ export const Chart: FC<ChartProps> = (props) => {
...
-                  {tick < ticksMax && <View style={styles.tick} />}
-                </>
-              )}
-            </View>
-          ))}
-        </View>
-      )}
-
       {/* Foreground */}
       <ChartContent maxTick={ticksMax} hasXAxis={isDefault}>
```
sternetj, line 61: Props are no longer used.

src/components/TrackTile/TrackerDetails/TrackerHistoryChart/Bar.tsx
bruce-glazier: I was experiencing a problem with an empty graph rendering as full with the negative value only when the bar is horizontal.

sternetj: Gotcha 👍 yeah this always seemed like a hack anyways.

PR #229, opened 2023-05-26
[PE-124] Update TrackerDetails to accept reference date

evanheisler:
## Changes
* Adds optional prop to `TrackerDetails` and `AdvancedTrackerDetails` to receive a date
* If the `referenceDate` is provided, the component uses that date instead of the current date
* Prevents passing a `referenceDate` in the future
* Adds `knobs` to storybook

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/600650/091bf3f5-b3c3-484c-88f9-d4078da55628

### DIFF
--- a/src/components/TrackTile/TrackerDetails/TrackerHistoryChart/TrackerHistoryChart.tsx
+++ b/src/components/TrackTile/TrackerDetails/TrackerHistoryChart/TrackerHistoryChart.tsx
@@ -1,12 +1,12 @@
-import React, { FC, useState, useCallback, useEffect } from 'react';
+import React, { FC, useState, useCallback } from 'react';
 import { useTrackerValues } from '../../hooks/useTrackerValues';
 import { Chart, ChartProps } from './Chart';
 import {
-  endOfToday,
   startOfToday,
   addDays,
   startOfWeek,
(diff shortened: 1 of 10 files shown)


src/components/TrackTile/TrackerDetails/TrackerHistoryChart/TrackerHistoryChart.tsx
```diff
@@ -23,37 +23,42 @@ type TrackerHistoryChartProps = {
...
+  );
   const { loading, error, trackerValues } = useTrackerValues(
     valuesContext,
     dateRange,
   );
 
-  useEffect(() => {
-    setDateRange(getInitialRange(dateRangeType));
-  }, [dateRangeType]);
-
```
evanheisler, line 56: This is sort of unrelated but noticed the potential here for unnecessary re-renders. I wouldn't expect this to be a prop that would change (basically just toggles between a calendar week and last 7 days) but it would fall under [these conditions](https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes) where `useEffect` isn't the best choice.

sternetj (approved):

PR #231, opened 2023-05-31
feat: Update safe area view

namick:
## Changes

  - Turned on-simulator storybook UI off to prevent safe-area-view from automatically being added
  - Fixed all stories related to this change
  - Added a developerConfig prop to control `statusBarHeight` of the AppNavHeader
  - Moved example of additional tab navigator tab to the stories

### DIFF
--- a/example/storybook/index.tsx
+++ b/example/storybook/index.tsx
@@ -32,6 +32,7 @@ configure(() => {
 // To find allowed options for getStorybookUI
 const StorybookUIRoot = getStorybookUI({
   asyncStorage: null,
+  onDeviceUI: false,
 });
 
 // If you are using React Native vanilla and after installation you don't see your app name here, write it manually.
--- a/example/storybook/stories/ExampleApp/ExampleApp.stories.tsx
+++ b/example/storybook/stories/ExampleApp/ExampleApp.stories.tsx
@@ -11,6 +11,9 @@ import {
 import { withKnobs, color } from '@storybook/addon-knobs';
 import Color from 'color';
 import logo from './header-logo.png';
+import { t } from 'i18next';
+import { Menu } from '@lifeomic/chromicons-native';
+import { HelloWorldScreen } from '../../../src/screens/HelloWorldScreen';
 
 storiesOf('Example App', module)
   .addDecorator(withKnobs)
@@ -24,6 +27,34 @@ storiesOf('Example App', module)
           simpleTheme: {
             primaryColor,
           },
+        }}
+      >
+        <RootProviders authConfig={authConfig}>
+          <RootStack />
+        </RootProviders>
+      </DeveloperConfigProvider>
+    );
+  })
+  .add('With additional navigation tab', () => {
+    const rgbColorString = color('Primary Color', '#0477BF');
+    const primaryColor = Color(rgbColorString).hex();
+
+    return (
+      <DeveloperConfigProvider
+        developerConfig={{
+          simpleTheme: {
+            primaryColor,
+          },
+          additionalNavigationTabs: [
+            {
+              name: 'AdditionalTab',
+              component: HelloWorldScreen,
+              options: {
+                tabBarLabel: t('tabs-settings', 'Settings'),
+                tabBarIcon: Menu,
+              },
+            },
+          ],
         }}
       >
         <RootProviders authConfig={authConfig}>
@@ -86,6 +117,7 @@ storiesOf('Example App', module)
                 color: (theme) => theme.colors.onTertiary,
               },
             ],
+            statusBarHeight: 0,
           },
         }}
       >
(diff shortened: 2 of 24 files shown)


example/storybook/index.tsx
```diff
@@ -32,6 +32,7 @@ configure(() => {
 // To find allowed options for getStorybookUI
 const StorybookUIRoot = getStorybookUI({
   asyncStorage: null,
+  onDeviceUI: false,
 });
```
namick, line 36: This is the main change of this PR.  If you agree, I'd like to turn off the device the on-simulator storybook UI (and thus rely on `yarn web-controls` to navigate the stories.  This is because the storybook UI automatically adds margin at the top and bottom of the screen and makes debugging things like the header and tab navigation nearly impossible.

Here are screenshots of the before and after of this change alone:

| Before | After |
|-|-|
| Screenshot 2023-05-31 at 8 36 22 AM | Screenshot 2023-05-31 at 8 36 08 AM |


Obviously that isn't ideal. I designed the header and tab navigation not realizing that that auto margin was there.  

The rest of this PR fixes the stories by adding safe-area-view to individual components and fixes the header issue on the full Example app stories.

The tab navigation will be fixed in the next PR

example/storybook/stories/TrackTile/Indicator.stories.tsx
```diff
@@ -7,32 +7,35 @@ import Indicator, {
 import * as Icons from '@lifeomic/chromicons-native';
 import { kebabCase, invert } from 'lodash';
 import { View, Text, ScrollView } from 'react-native';
+import { CenterView } from '../../helpers/CenterView';
 
-storiesOf('Indicator', module).add('default', () => {
-  const names = Object.keys(Icons).map(kebabCase);
+storiesOf('Indicator', module)
+  .addDecorator((story) => <CenterView>{story()}</CenterView>)
```
namick, line 13: This is an example of a fixed story with the `CenterView` decorator.

| Before | After |
|-|-|
| Screenshot 2023-05-31 at 9 00 47 AM | Screenshot 2023-05-31 at 9 01 43 AM |

src/components/AppNavHeader.tsx
```diff
@@ -30,9 +30,10 @@ export function AppNavHeader({
     route,
     styles.titleText,
   );
+  const statusBarHeight = config.AppNavHeader?.statusBarHeight;
 
   return (
-    <Appbar.Header statusBarHeight={0} style={headerStyles}>
+    <Appbar.Header statusBarHeight={statusBarHeight} style={headerStyles}>
```
namick, line 36: The default header already includes safe-area-view.  That needs to be turned off dynamically when there is a logo header above it.

| Before | After |
|-|-|
| Screenshot 2023-05-31 at 9 55 12 AM | Screenshot 2023-05-31 at 9 55 38 AM |

example/storybook/stories/ExampleApp/ExampleApp.stories.tsx
```diff
@@ -32,6 +35,34 @@ storiesOf('Example App', module)
...
+            },
+          ],
+        }}
+      >
+        <RootProviders authConfig={authConfig}>
+          <RootStack />
+        </RootProviders>
+      </DeveloperConfigProvider>
+    );
+  })
```
namick, line 65: I also made a new story of the full Example app showing an additional navigation tab.  This is moved from the `AppDemo.tsx` way of running the app (without storybook) which doesn't have the ability to show different configurations.

^ @cmjarreau

sternetj (approved):

src/components/AppNavHeader.tsx
```diff
@@ -30,9 +30,10 @@ export function AppNavHeader({
     route,
     styles.titleText,
   );
+  const statusBarHeight = config.AppNavHeader?.statusBarHeight;
```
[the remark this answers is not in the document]
namick: Yes, it might.  Currently we are relying on [Paper's AppBar](https://github.com/callstack/react-native-paper/blob/main/src/components/Appbar/AppbarHeader.tsx) which uses safe-area-view-context to add margin there (or not).  Looking [at the code](https://github.com/th3rdwave/react-native-safe-area-context/blob/main/src/SafeAreaContext.tsx), I it using `Dimensions` from react-native, vs the library you linked which [uses presets for various ios hardware](https://www.npmjs.com/package/react-native-status-bar-height?activeTab=code) and I'm not sure which is better.

Specifically, with "statusBarHeight" above, we currently only need to set this to `0` to turn it off, or leave it `undefined` to have the default margin set by react-native-safe-area-context.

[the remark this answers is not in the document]
sternetj: Gotcha, lets stick with what you have then. I think the package I suggested is too old anyways and the safe area context should be all we need.

bruce-glazier (approved):

PR #232, opened 2023-05-31
PHC-4837 Redesign Tracker Details

bruce-glazier:
## Changes

Starting from top of screen/down
  - Uses theme.colors.elevation.level1 as background for consistency with the Home Screen
  - Re-styles top date navigation component
  - Adds optional image that renders if provided in tracker.image
  - Colors other buttons to theme.colors.background
  - Date stepper control now renders below the chart and the active chevron is the tracker color
  - Reimagined bar graph now renders horizontally instead of vertically

Supporting changes
- Created a new TriangleFilled icon and added to the IconProvider
- Added polyfills for Intl in order to leverage the compact notation for the value on the graph


lib/i18n/polyfills/index.js
sternetj, line 1: Any particular reason that this and other polyfill files are `js` and not `ts`?

lib/i18n/polyfills/numberFormat.js
```diff
@@ -0,0 +1,13 @@
+export const polyfill = () => {
```
sternetj, line 1: Just checking, this one doesn't need an `ifNeeded` correct?

src/components/BrandConfigProvider/icons/IconProvider.tsx
```diff
@@ -1,16 +1,20 @@
...
+import { TriangleFilled } from './TriangleFilled';
 
-export type Icons = typeof Chromicons & Record<string, React.ComponentType>;
+export type Icons = typeof Chromicons &
+  Record<string, React.ComponentType> & {
+    TriangleFilled: typeof TriangleFilled;
+  };
 
-const IconContext = createContext<Icons>(Chromicons);
+const IconContext = createContext<Icons>({ ...Chromicons, TriangleFilled });
```
sternetj, line 10: This seems totally fine for this PR but it would be nice if we could just add it to Chromicons in a future version. I see a lot of these one-off icons and it feels like we would get a lot more benefit making them available via Chromicons.

src/components/TrackTile/TrackerDetails/TrackerHistoryChart/Chart.tsx
```diff
@@ -134,93 +77,123 @@ export const Chart: FC<ChartProps> = (props) => {
...
+              <Bar
+                variant={variant}
+                style={!isToday(day) ? { barFlat: { opacity: 0.35 } } : {}}
+                color={darkenHexColor(
+                  color,
+                  isToday(day) || isDefault ? 0 : 45,
+                )}
+                testID={tID(`history-chart-value-bar-${index}`)}
+                percentComplete={
+                  values[index] / ticksMax >= 1 ? 1 : values[index] / ticksMax
```
sternetj, line 98: I think you might be able to drop all of the d3 scale and tick stuff since we always display the bars based on the target now and they never exceed that width. Although this might be handling how we handle a target of zero. Not really sure how that works.

sternetj (approved):
This looks great to me and if you want to address anything in a follow up that works for me. 

One more thought, should the down arrow for the target selection match the color of the tracker?
image

lib/i18n/polyfills/index.js
bruce-glazier: Ported it over as-is but I will evaluate this in my next PR.

lib/i18n/polyfills/numberFormat.js
bruce-glazier: Same comment as above, I ported this over directly and didn't check the contents with much scrutiny. I will look closer in another PR.

src/components/BrandConfigProvider/icons/IconProvider.tsx
bruce-glazier: Yeah, great point. I will follow up with a PR into Chromicons.

bruce-glazier:
Merging as-is for now just so there is less to review next time. I still have a few more items I am working on but its getting close.

PR #238, opened 2023-06-01
chore: Refactor push notification initialization logic to provider

cmjarreau:
## Changes

  - Rename NotificationScreen.stories.tsx to NotificationsScreens.stories.tsx (plural form) to be consistent with usage throughout the rest of the app
  - create PushNotificationProvider - which is intended to move the Push Notification initialization state logic up the component tree to live with the rest of the apps initialization logic (i couldn't quite meet this goal - details below)

## Screenshots
n/a - refactoring code


example/storybook/stories/Notifications/NotificationsScreen.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react-native';
+import { NotificationsScreen } from '.';
+import { PushNotificationsProvider } from 'src/hooks/usePushNotifications';
+
+storiesOf('NotificationsScreen', module).add('demo', () => (
+  <PushNotificationsProvider>
```
cmjarreau, line 7: This can be improved by moving the provider up the component tree to live with the rest in RootProviders. If we want to drop the PushNotificationsProvider into RootProviders, we are running into an issue where the context is undefined. 

The only fix that I could come up with to resolve the context issue was to move the initialization logic back into the component which defeated the purpose of the provider in the first place. So I'll need to lean into yall expertise about the best way to resolve this issue.

example/storybook/stories/Notifications/index.tsx
```diff
@@ -66,9 +56,13 @@ const styles = StyleSheet.create({
 });
 
 export const NotificationsScreen = () => {
-  const [events, setEvents] = useState<Event[]>([]);
-  const { httpClient } = useHttpClient();
-  const { account } = useActiveAccount();
+  const pushNotificationsContext = useContext(PushNotificationsContext);
+
+  if (!pushNotificationsContext) {
+    return <Text>Context is undefined</Text>;
```
cmjarreau, line 62: Couple of things here - probably dont want to explicitly check that the context is defined here (and even so probably want to do something else in the case where it is undefined)

Going to lean into yalls expertise for this piece as well

example/storybook/stories/Notifications/index.tsx
```diff
@@ -198,7 +146,7 @@ export const NotificationsScreen = () => {
         </TouchableOpacity>
       </View>
 
-      {events.map((event, idx) => (
+      {(events as unknown as Event[]).map((event: Event, idx: number) => (
```
cmjarreau, line 149: this is ugly and fallout from some typing a bit higher in the app tree. will detail it below in this PR

src/hooks/usePushNotifications.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const [events, setEvents] = useState<
+    Event | { type: string; notification: Notification }[]
+  >([]);
+  const { httpClient } = useHttpClient();
+  const { account } = useActiveAccount();
+
+  useEffect(() => {
+    // Handler called when a notification is pressed
+    onNotificationOpened((notification) => {
+      setEvents((events) => [
```
cmjarreau, line 41: typescript was complaining about typing here, so i tried strongly typing `events` and `setEvents` which caused a cascade of type errors. So to resolve had to do the ugly things called out elsewhere in this PR. I guess I couldn't quite figure out the right typing

example/storybook/stories/Notifications/index.tsx
```diff
@@ -198,7 +141,7 @@ export const NotificationsScreen = () => {
...
```
cmjarreau, line 144: this is ugly and fallout from some typing a bit higher in the app tree. will detail it below in this PR

cmjarreau:
sorry my comments got 'outdated' and moved in my attempt to resolve the build issues. they are all still valid

cmjarreau:
sorry my comments got 'outdated' and moved in my attempt to resolve the build issues. they are all still valid

example/storybook/stories/Notifications/NotificationsScreen.stories.tsx
evanheisler: I think this is fine, if we are aiming to have Notifications "opt-in" rather than enabled by default. If I were a consumer of this lib, I would look at this story and say, "okay, I need to wrap my component in a provider to get it to work".

The reason you are getting context being undefined, is because you set the default to `undefined` [here](https://github.com/lifeomic/react-native-sdk/pull/238/files#diff-21335a07ddbab7e65d193bfa347f491ddcab07229361b14275778f3e0d92d759R23). Also why you have the override [here](https://github.com/lifeomic/react-native-sdk/pull/238/files#diff-1f54bc7376235dcc4820ec4b651589e27ef330fe6f71a5dcf182a3682cdad53cR60).

The simple fix to this would be to initialize the context, something like
```typescript
export const PushNotificationsContext =
  createContext<PushNotificationsStateType>({
    events: [],
    setEvents: () => {},
    httpClient: defaultAxiosInstance,
    account: undefined,
  });
```

But now looking closer at this type, I don't think you should be including the `httpClient` or `account` in the context, since these are readily available as their own hooks. The context should just be about notifications, then the implementation can include the hook(s).

example/storybook/stories/Notifications/index.tsx
evanheisler: Your return type [here](https://github.com/lifeomic/react-native-sdk/pull/238/files#diff-21335a07ddbab7e65d193bfa347f491ddcab07229361b14275778f3e0d92d759R15) says that `events` can be either a single generic DOM `Event` or an array of objects. Why is `Event` an option? Is this defined by the notification library?

cmjarreau: yes i think it must be supplied by the notifications library. one of the type errors that im getting ultimately bubbles down to: `Type '(events: Event[]) => (Event | { type: string; notification: Notification; })[]' is not assignable to type '(prevState: Event[]) => Event[]'.`

cmjarreau: yes i think it must be supplied by the notifications library. one of the type errors that im getting ultimately bubbles down to: `Type '(events: Event[]) => (Event | { type: string; notification: Notification; })[]' is not assignable to type '(prevState: Event[]) => Event[]'.`

cmjarreau: oops i said wrong - the Event type is defined here: https://github.com/lifeomic/react-native-sdk/pull/238/files#diff-1f54bc7376235dcc4820ec4b651589e27ef330fe6f71a5dcf182a3682cdad53cR12

src/hooks/usePushNotifications.tsx
```diff
@@ -0,0 +1,108 @@
...
+    };
+
+    getInitial();
+  }, []);
+
+  // Set the notification channel for Android
+  useEffect(() => {
+    if (Platform.OS === 'android') {
+      Notifications.setNotificationChannel({
+        channelId: 'sdk-example-channel',
```
aymanenadi, line 86: Ideally, we should expose a way for the developer to set these parameters. For now, I'd suggest deleting all the references to `example`, and replacing it with something like `LifeOmic react native SDK`.

src/hooks/usePushNotifications.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+interface PushNotificationsStateType {
+  events: PushNotificationsEvent[];
+  setEvents: Dispatch<SetStateAction<PushNotificationsEvent[]>>;
+}
+
+export const PushNotificationsContext =
+  createContext<PushNotificationsStateType>({
+    events: [],
+    setEvents: () => [],
```
aymanenadi, line 31: Can you explain how you're planning on using `events` and `setEvents` once they're exposed?

cmjarreau: these are just hooks for the push notification event queue - so by having these from a provider a user can have a button to clear/reset their notification queue for example

aymanenadi: I'm confused by this provider's use case. I'd imagine we'd use it to register the device for PN, which is not the case currently. We can also use it to expose events such as notifications received or notifications opened.
However, I don't see how `setEvents` would be used. Can you elaborate on why would a developer need to be able to set the events list internally in the provider? For example, a developer can do something like:
`setEvents([id: '1', title: 'I never received this notification, but I'm setting it in the provider'])`.

namick (approved):

markdlv (approved):

PR #247, opened 2023-06-07
fix: Set pillarTrackers state during onChange

bruce-glazier:
## Changes

  - Fixes consistency issue for the pillar trackers where a changed value (like target) will not persist through screen navigation.
  - Fixes rounding issue when selecting a fractional value via the target selector

### DIFF
--- a/src/components/TrackTile/TrackerDetails/coerce-to-nonnegative-value.ts
+++ b/src/components/TrackTile/TrackerDetails/coerce-to-nonnegative-value.ts
@@ -8,5 +8,5 @@ export const coerceToNonnegativeValue = (num: string, defaultNum: number) => {
     return defaultNum;
   }
 
-  return Math.round(targetNum);
+  return targetNum;
 };
--- a/src/components/TrackTile/hooks/useTrackers.ts
+++ b/src/components/TrackTile/hooks/useTrackers.ts
@@ -42,35 +42,71 @@ export const useTrackers = (trackersToUse?: Trackers) => {
         return;
       }
 
+      const pillarTileTrackers = installedMetrics.filter(
+        (t) => t.system === TRACKER_PILLAR_CODE_SYSTEM,
+      );
+      const trackTileTrackers = installedMetrics.filter(
+        (t) => !pillarTileTrackers.includes(t),
+      );
+
       setTrackers((currentTrackers) =>
         sortTrackers(
           currentTrackers
             ?.filter(
               (t) =>
-                !installedMetrics.find(
+                !trackTileTrackers.find(
                   ({ id, metricId }) => t.id === id || t.id === metricId,
                 ),
             )
-            .concat(installedMetrics),
+            .concat(trackTileTrackers),
+        ),
+      );
+
+      setPillarTrackers((currentTrackers) =>
+        sortTrackers(
+          currentTrackers
+            ?.filter(
+              (t) =>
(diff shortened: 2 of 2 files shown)


markdlv (dismissed):

src/components/TrackTile/TrackerDetails/coerce-to-nonnegative-value.ts
```diff
@@ -8,5 +8,5 @@ export const coerceToNonnegativeValue = (num: string, defaultNum: number) => {
     return defaultNum;
   }
 
-  return Math.round(targetNum);
```
bruce-glazier, line 11: I don't see any reason why we would need to be rounding this. Perhaps we did not allow fractional targets in the past.

markdlv: @sternetj do you have any recollection?  As long as we have display logic to ensure we never show something like 1.3333333, I agree this should be fine.

markdlv (approved):

src/components/TrackTile/TrackerDetails/coerce-to-nonnegative-value.ts
bruce-glazier: I can't think of anywhere which we compute these values in a way that could result in that many digits.

sternetj: I bet this was to handle when we had a number input for selecting your target. A user could enter something like `13e-1` which is a valid JS number but not a whole number.

It seems like it is still used on the tracker value. Was this causing you problems @bruce-glazier

src/components/TrackTile/hooks/useTrackers.ts
```diff
@@ -42,35 +42,71 @@ export const useTrackers = (trackersToUse?: Trackers) => {
         return;
       }
 
+      const pillarTileTrackers = installedMetrics.filter(
+        (t) => t.system === TRACKER_PILLAR_CODE_SYSTEM,
+      );
```
sternetj, line 47: Should we pass the context to this hook so we can filter by the context system and not have to have the duplicate logic?

src/components/TrackTile/TrackerDetails/coerce-to-nonnegative-value.ts
bruce-glazier: We allow for selecting values such as 0.25, 0.5, 0.75 for units like hours and we use this function to convert from string to number for storing the value. It was rounding those to the nearest whole number.

src/components/TrackTile/hooks/useTrackers.ts
bruce-glazier: Sure, I'll take a look at that.

src/components/TrackTile/TrackerDetails/coerce-to-nonnegative-value.ts
sternetj: Gotcha, that makes sense. Sounds like its safe to remove

src/components/TrackTile/hooks/useTrackers.ts
bruce-glazier: I don't think the juice is worth the squeeze here as it results in a lot more lines of added code passing valuesContext around.

PR #249, opened 2023-06-12
fix: Set view height so it always uses 100%

bruce-glazier:
## Changes

  - Fixes bug where the view might not stretch to fit the entire screen depending on content/actions

### DIFF
--- a/src/components/Circles/PostsList.tsx
+++ b/src/components/Circles/PostsList.tsx
@@ -65,7 +65,7 @@ export const PostsList = ({ circleTile, onOpenPost }: PostsListProps) => {
   }
 
   return (
-    <View>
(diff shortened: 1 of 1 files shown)


src/components/Circles/PostsList.tsx
```diff
@@ -129,7 +129,8 @@ const defaultStyles = createStyles('PostsList', (theme) => {
   };
 
   return {
-    scrollView: {},
+    rootView: { minHeight: '100%', maxHeight: '100%' },
+    scrollView: { minHeight: '100%' },
```
markdlv, line 133: Is this `scrollView` used?

namick (approved):

PR #257, opened 2023-06-19
feat: Inject custom screens in the Home navigation stack

aymanenadi:
## Changes
- This PR exposes the Home navigation stack so custom screens can be added to it.

## Screenshots
- In this video, The FHIR Example screen is a custom screen that was added to the Home navigation stack.
- The `App Tile` screen is an existing screen, and demonstrates that we can from custom screens to "LifeOmic" screens.

https://github.com/lifeomic/react-native-sdk/assets/19395435/ccea8236-41fd-4175-9fa1-8a98459dacff

### DIFF
--- a/example/storybook/index.tsx
+++ b/example/storybook/index.tsx
@@ -32,7 +32,7 @@ configure(() => {
 // To find allowed options for getStorybookUI
 const StorybookUIRoot = getStorybookUI({
   asyncStorage: null,
-  onDeviceUI: false,
+  onDeviceUI: true,
 });
 
 // If you are using React Native vanilla and after installation you don't see your app name here, write it manually.
--- a/README.md
+++ b/README.md
@@ -85,6 +85,85 @@ export default function App() {
 }
 ```
 
+### Inject custom screens to the root stack
+
+1- Define your own custom screens.
+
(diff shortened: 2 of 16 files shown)


example/storybook/index.tsx
```diff
@@ -32,7 +32,7 @@ configure(() => {
 // To find allowed options for getStorybookUI
 const StorybookUIRoot = getStorybookUI({
   asyncStorage: null,
-  onDeviceUI: false,
+  onDeviceUI: true,
```
sternetj, line 35: Not sure if we want to commit this. I know @namick took it out because of style inconsistencies.

sternetj (approved):

example/storybook/index.tsx
aymanenadi: I was curious why this was deactivated. I couldn't use pick any of the storybook stories without this.

namick: You can use `yarn web-controls` to pick stories. without the in simulator UI.  

I took it out because it wraps the stories in another view that adds padding which doesn't work for debugging the header/footers.

That said, I'm fine with adding it back.  It is easy to disable when needed.

PR #258, opened 2023-06-20
fix: Fix Android app crash due to Firebase init

cmjarreau:
## Changes

  - BuildConfig variable injection allowing users to opt-in/opt-out of push notifications and firebase init
  - refactor build.gradle file to conditionally initialize firebase deps
  - add gradle property to gradle.properties that is used to conditionally initialize firebase deps
  -  update example app package.json
  - update README.md with directions for users to set up push notifications and firebase initialization for their apps

## Screenshots
n/a


README.md
```diff
@@ -113,6 +113,49 @@ both served from the `example` folder.
...
+
+##### 3. Add google-services.json
+
+Please bring in your own `google-services.json` file and add it to the
+appropriate location in your project. The `google-services.json` file is
+required for Firebase services to work correctly.
+
+Make sure to follow the Firebase documentation to obtain the
+`google-services.json` file specific to your Firebase project.
+
```
aymanenadi, line 158: Do we need to set `enableFirebase` and `IS_FIREBASE_ENABLED ` to true? Do we need 2 variables?

Do we need to update `pushNotificationsConfig` on the JS side?

cmjarreau: my thinking is that if we set `enableFirebase` and `IS_FIREBASE_ENABLED` to true, then the code paths that lead to firebase initialization are going to execute and cause the crash

the difference between these 2 variables is that one is used as a boolean in the gradle files and the other in the java files. So i think that we will still need them both unless there is room to remove conditionals all together like here possibly:
```
if (BuildConfig.IS_FIREBASE_ENABLED) {
  // If you opted-in for Push Notifications, Firebase is initialized.
  // FirebaseApp.initializeApp(this);
}
```
i could see an argument to just get rid of the conditional all together and leave initialization commented until a user is ready to use it

i dont think that we need to touch `pushNotificationsConfig` at this time but im open to ideas and suggestions. I might not be seeing something

namick: I think it is fine to remove the conditional like you suggested and leave it commented out as long as there are explicit instructions that lead from the README to this file, and then something like "Uncomment next line to enable firebase".

namick (approved):

package.json
```diff
@@ -141,6 +139,7 @@
     "events": "^3.3.0",
     "intl-pluralrules": "^1.3.1",
     "lodash": "^4.17.21",
+    "react-native-notifications": "^4.3.5",
```
sternetj, line 142: Do we want to make this a full dependency?

example/android/gradle.properties
```diff
@@ -42,3 +42,7 @@ newArchEnabled=false
 # Use this property to enable or disable the Hermes JS engine.
 # If set to false, you will be using JSC instead.
 hermesEnabled=true
+
+# Use this property to enable or disable push notifications and firebase init
+# If set to false, firebase will not be initialized
+enableFirebase=false
```
sternetj, line 48: Why not call this `enablePushNotifications` or something similar.

README.md
```diff
@@ -113,6 +113,49 @@ both served from the `example` folder.
...
 - type `i` or `a` to run on either platform
 
+#### Push Notifications for Android and Firebase services Integration Guide
+
+Push notifications require `react-native-notifications` and Firebase services.
+To integrate with your example app, follow the steps below:
+
+##### 1. Modify MainApplication.java
+
+Add the following imports to `MainApplication.java` file:
```
sternetj, line 123: Since the next step is uncomment, should we make this a comment instead of an add?

example/android/gradle.properties
cmjarreau: good call. i went back and forth with that trying to anticipate pulling in more firebase dependent features. updating to your recommendation!

example/android/app/src/main/java/com/example/MainApplication.java
```diff
@@ -55,7 +55,11 @@ public ReactNativeHost getReactNativeHost() {
   @Override
   public void onCreate() {
     super.onCreate();
-    FirebaseApp.initializeApp(this);
+
```
bruce-glazier, line 58: I assume based on your earlier comment you plan to uncomment this and use the enablePushNotification flag here to conditionally init. That sounds like a good plan to me.

package.json
```diff
@@ -119,7 +119,6 @@
     "react-native-gesture-handler": ">=2.9.0",
     "react-native-keychain": ">=8.1.1",
     "react-native-markdown-display": ">=6.1.6",
-    "react-native-notifications": ">=4.3.3",
```
bruce-glazier, line 122: Why should you remove this from peerDependencies? I'd imagine it still is a peer depend, no?

bruce-glazier:
Nit: While you are finalizing things tt would be nice to see the number of commits reduced and removal of the WIP commits.

package.json
cmjarreau: yep restored it back to peerDependency

example/android/app/src/main/java/com/example/MainApplication.java
bruce-glazier: Probably not worth the effort right now but there are ways to dynamically inject dependencies using Dagger. For now, uncommenting is easy enough but if we want to clean this up in the future we could consider an approach like that.

cmjarreau: sweet! yeah i was trying to dig in to a way to use dynamic imports in java but not much progress on that front. Dagger sounds cool!

cmjarreau:
closing this pull request and opening a new one with a better commit history. will re-add everyone

cmjarreau:
closing this pull request and opening a new one with a better commit history. will re-add everyone

PR #260, opened 2023-06-22
fix: PHC-4968 Add Chart Style Overrides

sternetj:
## Changes

  - Adds the ability to specify chart styles using style overrides
  - Makes the independent axis labels use the trace color
  - Adds the ability to set a color on the trace

## Screenshots

image

### DIFF
--- a/src/components/MyData/styles.ts
+++ b/src/components/MyData/styles.ts
@@ -1,0 +1,59 @@
+import { StyleSheet } from 'react-native';
+import { VictoryThemeDefinition } from 'victory-core';
+import { createStyles } from '../BrandConfigProvider';
+
+type StyleExtractor<T> = T extends { style: unknown } ? T['style'] : never;
+type VictoryStyleProp<K extends keyof VictoryThemeDefinition> = StyleExtractor<
+  Required<Required<VictoryThemeDefinition>[K]>
+>;
+
+export const defaultChartStyles = createStyles('ChartStyles', (theme) => ({
+  independentAxis: {
+    grid: {
+      strokeDasharray: '0 2 0',
+      stroke: theme.colors.outline,
+    },
+    axis: {
+      stroke: theme.colors.onBackground,
+    },
+    tickLabels: {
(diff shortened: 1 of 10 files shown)


src/components/MyData/styles.ts
```diff
@@ -0,0 +1,59 @@
...
+      stroke: theme.colors.outline,
+    },
+    axis: {
+      stroke: theme.colors.onBackground,
+    },
+    tickLabels: {
+      padding: 8,
+      fill: theme.colors.onBackground,
+    },
+  } as VictoryStyleProp<'axis'>,
```
bruce-glazier, line 23: I assume wrapping the T[P] with Partial was the trick to making this work correctly? Love it.

src/components/MyData/styles.ts
```diff
@@ -0,0 +1,59 @@
+import { StyleSheet } from 'react-native';
+import { VictoryThemeDefinition } from 'victory-core';
+import { createStyles } from '../BrandConfigProvider';
+
+type StyleExtractor<T> = T extends { style: unknown } ? T['style'] : never;
```
bruce-glazier, line 5: This is to resolve the correct nested style typings when passing in style overrides via a prop, is that right? Pretty cool!

bruce-glazier (approved):
Looks great!

src/components/MyData/styles.ts
sternetj: Yeah exactly. I didn't confirm but I would bet that not all of the keys of `VictoryThemeDefinition` have a `style` property (i.e. `VictoryThemeDefinition.[key].style?`) and so I needed to add the conditional typing so that typescript could strongly type the style prop when extracting it only if it exists.

src/components/MyData/styles.ts
sternetj: Yeah exactly, without it, a user would have to define all fields of `tickLabels` if they were defining a `tickLabel` style.

PR #269, opened 2023-06-28
feat: Add invite required screen

bruce-glazier:
## Changes

- Adds new screen to the RootStack for a user that does have both an account and project set
- Vigorous debugging of some react queries uncovered some inconsistencies around loading/enabled controls and this PR makes several updates to ensure correct loading behavior.

## Screenshots

Simulator Screen Shot - iPhone 14 Pro - 2023-06-28 at 13 39 58

### DIFF
--- a/src/hooks/useActiveProject.test.tsx
+++ b/src/hooks/useActiveProject.test.tsx
@@ -57,9 +57,13 @@ const renderHookInContext = async () => {
 beforeEach(() => {
   useSubjectProjectsMock.mockReturnValue({
     data: mockSubjectProjects,
+    isLoading: false,
+    isFetched: true,
   });
   useMeMock.mockReturnValue({
     data: mockMe,
+    isLoading: false,
+    isFetched: true,
   });
 
   useAsyncStorageSpy.mockReturnValue([
@@ -104,7 +108,7 @@ test('exposes some props from useSubjectProjects and useMe', async () => {
     isFetched: true,
   });
   useMeMock.mockReturnValue({
-    isLoading: true,
+    isLoading: false,
     isFetched: true,
     error,
   });
@@ -112,7 +116,7 @@ test('exposes some props from useSubjectProjects and useMe', async () => {
   await rerender({});
 
   expect(result.current).toMatchObject({
-    isLoading: true,
+    isLoading: false,
     isFetched: true,
     error,
   });
--- a/src/hooks/useActiveProject.tsx
+++ b/src/hooks/useActiveProject.tsx
@@ -63,10 +63,14 @@ export const ActiveProjectContextProvider = ({
   children?: React.ReactNode;
 }) => {
   const projectsResult = useSubjectProjects();
+  const projectLoading = projectsResult.isLoading || !projectsResult.isFetched;
+
   const useMeResult = useMe();
+  const useMeLoading = useMeResult.isLoading || !useMeResult.isFetched;
   const [hookReturnValue, setHookReturnValue] = useState<ActiveProjectProps>(
     {},
   );
+  const [isFetched, setIsFetched] = useState<boolean>(false);
   const [storedProjectIdResult, setStoredProjectId] =
     useAsyncStorage(selectedProjectIdKey);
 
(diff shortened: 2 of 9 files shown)


src/hooks/useSubjectProjects.tsx
```diff
@@ -27,10 +27,7 @@ export function useSubjectProjects() {
         )
         .then((res) => res.data.items),
     {
-      enabled:
-        !!accountHeaders &&
-        (subjects?.length ?? 0) > 0 &&
-        !!subjects?.every((s) => s.projectId),
+      enabled: !!accountHeaders,
```
bruce-glazier, line 30: This comment here explains the rationale behind this change:
https://github.com/lifeomic/react-native-sdk/pull/269/files#diff-ca654e5fdb78a75f406342fabca94da850d4e70528a8162257f297b2a9fd89e3R65-R71

src/hooks/useActiveProject.tsx
```diff
@@ -131,7 +139,11 @@ export const ActiveProjectContextProvider = ({
         ...hookReturnValue,
         setActiveProjectId,
         isLoading: !!(projectsResult.isLoading || useMeResult.isLoading),
-        isFetched: !!(projectsResult.isFetched && useMeResult.isFetched),
+        isFetched: !!(
+          projectsResult.isFetched &&
+          useMeResult.isFetched &&
+          isFetched
```
bruce-glazier, line 145: The hook return value will lag behind the projectsResult.isFetched and useMeResult.isFetched by at least one render cycle. Adding our own flag to indicate when the hooks logic has successfully computed the result helps consumers of this hook understand when it is actually ready.

src/hooks/useActiveProject.tsx
```diff
@@ -76,8 +80,8 @@ export const ActiveProjectContextProvider = ({
   useEffect(() => {
     if (
       hookReturnValue.activeProject?.id || // active project already set
-      !projectsResult.data?.length || // wait for projects endpoint to return data
-      !useMeResult.data?.length || // wait for subjects endpoint to return data
+      projectLoading || // wait for projects endpoint
```
bruce-glazier, line 83: The presence of data should not dictate when this hook begins computing an activeProject given it is valid to arrive here without subjects or projects associated to a user. For everything to flow correctly we should instead attempt to compute the hook result if the queries we depend on have settled one way or another.

src/hooks/useActiveProject.test.tsx
```diff
@@ -104,15 +108,15 @@ test('exposes some props from useSubjectProjects and useMe', async () => {
     isFetched: true,
   });
   useMeMock.mockReturnValue({
-    isLoading: true,
```
bruce-glazier, line 107: This was actually a pretty good indication that we had a problem here since isLoading and isFetched should be mutually exclusive .

namick (approved):

src/navigators/RootStack.tsx
```diff
@@ -25,22 +46,42 @@ export function RootStack() {
...
+      ? 'InviteRequired'
+      : shouldRenderConsentScreen
+      ? 'screens/ConsentScreen'
+      : 'app';
+
+    if (loadingAccountOrProject && initialRoute === 'InviteRequired') {
       return (
         <ActivityIndicatorView
-          message={t('root-stack-waiting-for-consent', 'Waiting for consent')}
+          message={t('root-stack-waiting-for-account-and-project', 'Loading')}
```
markdlv, line 61: Why take these arguably more helpful loading messages away?  With this `Loading` matching line 69, we won't be able to visually tell which the app is waiting for (outside of logs).

markdlv (approved):

src/navigators/RootStack.tsx
bruce-glazier: I see, I hadn't realized these messages actually appear in the UI after a timeout. Given that you are totally right, I'll update them.

markdlv: I'd like to get see this in a customer build asap, so merging and one of us can follow up on those loading messages later 👍

src/hooks/useSubjectProjects.tsx
markdlv: But the subjects' projectId prop goes into the URL itself.  Re-looking at this, setting a query up to purposefully fail doesn't feel like the right approach.  Along with the loading messages update to come - let's (you or I) make this more intelligent (i.e. to not fetch, but also not get stuck in any endless "loading" states).

src/hooks/useActiveProject.tsx
```diff
@@ -63,10 +63,14 @@ export const ActiveProjectContextProvider = ({
   children?: React.ReactNode;
 }) => {
   const projectsResult = useSubjectProjects();
+  const projectLoading = projectsResult.isLoading || !projectsResult.isFetched;
```
sternetj, line 66: I wonder if all we need is `!projectsResult.isFetched`. If the data is fetched but the hook still says it is loading does it matter? I would think we would only really need to monitor `isLoading` if we are doing a refetch or loading more results.

PR #272, opened 2023-07-05
fix: Improve handling when no subjects exist

bruce-glazier:
## Changes

Instead of performing a GET on an invalid URL return an empty array when subjects is not populated

### DIFF
--- a/src/hooks/useSubjectProjects.tsx
+++ b/src/hooks/useSubjectProjects.tsx
@@ -17,15 +17,22 @@ export function useSubjectProjects() {
   const { data: subjects } = useMe();
   const { httpClient } = useHttpClient();
 
-  return useQuery(
+  return useQuery<Project[]>(
     [`${account?.id}-projects`, subjects],
-    () =>
-      httpClient
-        .get<ProjectsResponse>(
+    async () => {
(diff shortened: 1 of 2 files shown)


src/hooks/useSubjectProjects.tsx
```diff
@@ -17,15 +17,22 @@ export function useSubjectProjects() {
...
           { headers: accountHeaders },
-        )
-        .then((res) => res.data.items),
+        );
+        return res.data.items;
+      } else {
+        // Having no subjects is a supported state.
+        // Instead of disabling the query to wait for subjects we are
+        // returning a mock response to keep downstream queries moving
+        return new Promise((resolve) => resolve([]));
```
markdlv, line 33: We're already in an async func (implicit promise) - why not just `return []`?

bruce-glazier: Opps, I allowed prettier to convert this into an async function. I'll fix it.

bruce-glazier: Done

markdlv (approved):

markdlv (approved):

PR #273, opened 2023-07-06
fix: fix android app crashes due to firebase init

cmjarreau:
## Changes
  - Updates to README detailing how users can opt-in to push notifications
  - mechanism to safely import react-native-notifications throughout SDK

## Screenshots
n/a

### DIFF
--- a/example/ios/example/AppDelegate.mm
+++ b/example/ios/example/AppDelegate.mm
@@ -3,7 +3,8 @@
 #import <React/RCTBundleURLProvider.h>
 #import <React/RCTLinkingManager.h>
 
-#import <RNNotifications.h>
+// Needed for push notifications
+// #import <RNNotifications.h>
 
 @implementation AppDelegate
 
@@ -15,7 +16,7 @@
   self.initialProps = @{};
 
   // Needed for push notifications
-  [RNNotifications startMonitorNotifications]; 
+  // [RNNotifications startMonitorNotifications]; 
 
   return [super application:application didFinishLaunchingWithOptions:launchOptions];
 }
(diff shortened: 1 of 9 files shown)


example/ios/example/AppDelegate.mm
```diff
@@ -3,7 +3,8 @@
 #import <React/RCTBundleURLProvider.h>
 #import <React/RCTLinkingManager.h>
 
-#import <RNNotifications.h>
```
cmjarreau, line 6: this approach and below is not ideal. i tried (only for a short time) to make a config file and opt-in that way but wasnt having much luck so decided to put the PR up for review and collect feedback from you guys on a potentially better way

src/common/Notifications.ts
```diff
@@ -1,13 +1,29 @@
...
+// react-native-notifications code to exist.
+export const safelyImportReactNativeNotifications = async () => {
+  const { pushNotificationsConfig } = useDeveloperConfig();
+  // we cant cast to the actual NotificationsRoot at this time due to dynamic imported dependency.
+  // this is true for types used below as well.
+  let rnnotifications: any;
+  if (pushNotificationsConfig?.enabled) {
+    const modulePath = 'react-native-notifications';
+    if (pushNotificationsConfig.applicationName !== 'example') {
+      const resolvePath = require.resolve(modulePath);
```
cmjarreau, line 18: this logic is to prevent app crashes in the r-n-starter app

src/common/Notifications.ts
```diff
@@ -34,53 +50,59 @@ export const registerDeviceToken = ({
...
+export const requestNotificationsPermissions = async (
   callback: ({
     deviceToken,
     denied,
     error,
   }: {
     deviceToken?: string;
     denied?: boolean;
-    error?: RegistrationError;
+    error?: any;
```
cmjarreau, line 71: i would have preferred not to use `any` here and below but there is no way to dynamically import types that i know of. Since we are relying on dynamic imports, my hand was forced. Maybe you guys know of a mechanism around this?

src/common/Notifications.ts
cmjarreau: there still exists an issue here - for the Starter App to not crash we need to use an import like:
```
const resolvePath = require.resolve(modulePath);
rnnotifications = await import(resolvePath);
```
for the SDK example app not to crash, we need to use a standard import 
```
rnnotifications = await import(modulePath);
```
the attempted setup is to dynamically import these in a way that is fit for whatever app is being used but the paths are being resolved at compile time (not runtime) so still throwing errors.

im not sure how we can satisfy both at once. ive tried everything i can think of

cmjarreau:
we are going to take this in a different direction - adding google-services.json file and hopefully simplifying the logic considerably

PR #274, opened 2023-07-07
fix: resolve consent screen not showing

markdlv:


example/package.json
```diff
@@ -39,6 +39,7 @@
     "react-native-gesture-handler": "^2.9.0",
     "react-native-keychain": "^8.1.1",
     "react-native-markdown-display": "^6.1.6",
+    "react-native-notifications": "^4.3.3",
```
markdlv, line 42: This is an unrelated change to the PR's focus, but it's required to get the example app back up & running.

src/navigators/RootStack.tsx
```diff
@@ -49,28 +49,31 @@ export function RootStack() {
...
+          message={t('root-stack-waiting-for-consents', 'Loading consents')}
         />
       );
     }
 
+    const initialRoute = !hasAccountAndProject
+      ? 'InviteRequired'
+      : shouldRenderConsentScreen
+      ? 'screens/ConsentScreen'
+      : 'app';
```
markdlv, line 75: An important trick with this `initialRouteName` approach is that you can't really change it once the `return` below has been rendered.  So after a recent refactor, we get `shouldRenderConsentScreen` initially coming through here as `false` and then later as `true`.  Unless we avoid rending the `Stack.Navigator` below on those fully loading (accounts & consents), the `initialRouteName` doesn't behave/react like we need it to.  The result in this case is I'm consistently seeing consents in a proposed state that aren't being shown to the user, without this PR.

src/navigators/RootStack.tsx
```diff
@@ -49,28 +49,31 @@ export function RootStack() {
...
-          message={t('root-stack-waiting-for-account-and-project', 'Loading')}
+          message={t(
+            'root-stack-waiting-for-account-and-project',
+            'Loading account',
+          )}
         />
       );
     }
 
-    if (loadingConsents && initialRoute === 'screens/ConsentScreen') {
```
bruce-glazier, line 66: I think the reason I did this is because the consents API call never becomes enabled if there is not a valid activeProject which leads to being stuck at this loading indicator indefinitely when we want to push the user to the Invite required screen. Obviously there is something else wrong here though if this resolves the problem you are facing.

cmjarreau (approved):

src/navigators/RootStack.tsx
markdlv: Yep- I hit that right after I put the PR up and was handling in a separate PR, but have now just committed the rest into this PR.

src/navigators/RootStack.tsx
```diff
@@ -28,14 +29,17 @@ export function RootStack() {
...
   const { useShouldRenderConsentScreen } = useConsent();
   const { shouldRenderConsentScreen, isLoading: loadingConsents } =
     useShouldRenderConsentScreen();
 
   const loadingProject = !isFetchedProject || isLoadingProject;
   const loadingAccount = !isFetchedAccount || isLoadingAccount;
-  const loadingAccountOrProject = loadingProject || loadingAccount;
+  const hasAccount = !loadingAccount && account?.id;
+  const loadingAccountOrProject =
+    loadingAccount || (hasAccount && loadingProject);
```
markdlv, line 42: `(hasAccount && loadingProject)` covers the case where the user is part of NO accounts.  In that case, we don't want to wait on `loadingProject` as it's indefinitely true due to accountHeaders never being set.

src/screens/InviteRequiredScreen.tsx
```diff
@@ -1,91 +1,31 @@
...
-  const {
-    activeProject,
-    isLoading: isLoadingProject,
-    isFetched: isFetchedProject,
-  } = useActiveProject();
-
-  const loadingProject = !isFetchedProject || isLoadingProject;
-  const loadingAccount = !isFetchedAccount || isLoadingAccount;
-  const loadingAccountOrProject = loadingProject || loadingAccount;
-  const hasAccountAndProject = !!(activeProject?.id && account?.id);
```
markdlv, line 36: While debugging, I found there to be too duplicated logic in here.  For example, this has the same issue for a user that has access to NO accounts (it will wait on `loadingProject` indefinitely).  Keeping it as simple as I've made it means that, once you reach this screen - unless RootStack reacts to something else - your _only_ option is to logout and start over.  This should be a rare edge case for a user to experience, so while we might want to improve that UX long-term, I like starting with simplicity for now.

bruce-glazier: I did it to be consistent with how we had the screen prior but I agree that it doesn't add much value if this can be handled via closing the app and relaunching and it is already an edge case to begin with.

src/navigators/RootStack.tsx
bruce-glazier: Oh I see, this is the escape hatch I was missing. Nice.

bruce-glazier (approved):

PR #278, opened 2023-07-12
fix: PHC-4968 Fix Data Selection on Device

sternetj:
## Changes

  - Reworked the data selection to use absolutely positioned TouchableWithoutFeedback for data selection. SVG onPress wasn't working on iOS on a real device
  - Switched the date range format because creating a new array was causing selection to be dropped on rerender

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/2295908/e5ed1cc9-9046-401e-8eae-cf792e07ec4d

### DIFF
--- a/src/components/MyData/LineChart/DataSelector.tsx
+++ b/src/components/MyData/LineChart/DataSelector.tsx
@@ -1,11 +1,11 @@
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
-import { StyleSheet, View } from 'react-native';
+import React, { useCallback, useMemo } from 'react';
+import { StyleSheet, View, TouchableWithoutFeedback } from 'react-native';
 import { VictoryBar } from 'victory-native';
 import { useCommonChartProps } from '../useCommonChartProps';
 import { eachDayOfInterval, format, isSameDay } from 'date-fns';
 import { sortBy } from 'lodash';
 import { PointData } from './useChartData';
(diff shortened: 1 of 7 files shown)


src/components/MyData/LineChart/DataSelector.tsx
```diff
@@ -168,7 +165,7 @@ const CustomLabel = ({ selection, x = 0, y = 0 }: CustomLabelProps) => {
   const { styles } = useStyles(defaultChartStyles);
 
   const decreaseBy = !selection.point1 || !selection.point2 ? 25 : 0;
-  const shiftX = Math.min(selection.chartWidth - (x + 115 - decreaseBy), 0);
+  const shiftX = Math.min(selection.chartWidth - (x + 15 - decreaseBy), 0);
```
markdlv, line 168: Nit-pick: What is this `15` and is there any way to reference it?  Just giving it a variable name, and opening up the possibility of future reuse, might be nice.

markdlv (dismissed):

src/components/MyData/LineChart/DataSelector.tsx
sternetj: Yeah great question, it is an amount to shift the tooltip by so that it doesn't get cut off. Let me see if I can make this a bit more obvious.

namick (approved):

PR #286, opened 2023-07-14
fix: Clear caches on logout

bruce-glazier:
## Changes

To support logout and user switching:
- Clear selected account when the user id changes
- Clear selected project when the user id changes
- Clear cached trackers when the user id changes
- Clear queryClient cache on logout

Also added a _DEV_ only field to the profile screen to show the current project name.

## Screenshots

N/A

### DIFF
--- a/src/components/TrackTile/hooks/useAxiosTrackTileService.ts
+++ b/src/components/TrackTile/hooks/useAxiosTrackTileService.ts
@@ -1,5 +1,5 @@
 import { AxiosRequestConfig } from 'axios';
-import { useRef } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import {
   InstalledMetric,
   TrackTileService,
@@ -28,6 +28,7 @@ import { pick, merge, pickBy, omit, fromPairs } from 'lodash';
 import { useHttpClient } from '../../../hooks/useHttpClient';
 import { useActiveAccount } from './../../../hooks/useActiveAccount';
 import { useActiveProject } from './../../../hooks/useActiveProject';
+import { useUser } from '../../../hooks';
 
 const axiosConfig = (obj: { account: string }): AxiosRequestConfig => ({
   headers: {
@@ -40,6 +41,8 @@ export const useAxiosTrackTileService = (): TrackTileService => {
   const { httpClient } = useHttpClient();
   const { activeSubjectId: patientId, activeProject } = useActiveProject();
   const { account } = useActiveAccount();
+  const { data } = useUser();
+  const [userId, setUserId] = useState(data?.id);
 
   const accountId = account?.id || '';
   const projectId = activeProject?.id || '';
@@ -49,6 +52,17 @@ export const useAxiosTrackTileService = (): TrackTileService => {
     trackerValues: ContextTrackerValues;
     ontologies: Record<string, CodedRelationship[]>;
   }>({ trackerValues: {}, ontologies: {} });
+
+  // Clear cached trackers when
+  // we've detected that the userId has changed
+  useEffect(() => {
+    if (userId !== data?.id) {
+      cache.trackers = undefined;
+      cache.trackerValues = {};
+      cache.ontologies = {};
+      setUserId(data?.id);
+    }
+  }, [userId, data?.id, cache]);
 
   const updateSettingsInCache = (settings: BulkInstalledMetricSettings) => {
     const tracker = cache.trackers?.find(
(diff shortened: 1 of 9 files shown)


src/components/TrackTile/hooks/useAxiosTrackTileService.ts
```diff
@@ -84,6 +98,7 @@ export const useAxiosTrackTileService = (): TrackTileService => {
 
     fetchTrackers: async () => {
       if (cache.trackers) {
+        console.log('Returning cached trackers!!!');
```
bruce-glazier, line 101: Remove....

src/components/TrackTile/hooks/useAxiosTrackTileService.ts
```diff
@@ -40,6 +41,8 @@ export const useAxiosTrackTileService = (): TrackTileService => {
   const { httpClient } = useHttpClient();
   const { activeSubjectId: patientId, activeProject } = useActiveProject();
   const { account } = useActiveAccount();
+  const { data } = useUser();
```
bruce-glazier, line 44: Rename...

src/hooks/useActiveProject.tsx
```diff
@@ -70,15 +71,19 @@ export const ActiveProjectContextProvider = ({
   const [hookReturnValue, setHookReturnValue] = useState<ActiveProjectProps>(
     {},
   );
+  const { data } = useUser();
+  const [userId, setUserId] = useState(data?.id);
```
sternetj, line 75: Might be a little easier to read if we map `data.id` to `userId` and then use a `usePrevious` on that value to see if it changes. That way `userId` will always be the current id.

sternetj (approved):

src/hooks/useActiveProject.tsx
bruce-glazier: Updated

src/components/TrackTile/hooks/useAxiosTrackTileService.ts
```diff
@@ -28,6 +28,7 @@ import { pick, merge, pickBy, omit, fromPairs } from 'lodash';
 import { useHttpClient } from '../../../hooks/useHttpClient';
 import { useActiveAccount } from './../../../hooks/useActiveAccount';
 import { useActiveProject } from './../../../hooks/useActiveProject';
+import { useUser } from '../../../hooks';
```
markdlv, line 31: Nit-pick: ideally we wouldn't import from index files like this.  It creates a scenario where, if someone wants to import this service but does NOT want to import another hook, an SDK consumer is out of luck.  If we only import non-index-files, it leaves the door open to avoid hooks/components/etc. an SDK consumer is having trouble with.

src/hooks/useActiveAccount.tsx
```diff
@@ -61,14 +62,20 @@ export const ActiveAccountContextProvider = ({
   const accountsResult = useAccounts();
   const accountsWithProduct = filterNonLRAccounts(accountsResult.data);
   const [activeAccount, setActiveAccount] = useState<ActiveAccountProps>({});
-  const [storedAccountIdResult, setStoredAccountId] =
-    useAsyncStorage(selectedAccountIdKey);
+  const { data } = useUser();
+  const { data: userData } = useUser();
```
markdlv, line 66: Can we remove line 65?

src/hooks/useActiveAccount.tsx
```diff
@@ -61,14 +62,20 @@ export const ActiveAccountContextProvider = ({
...
+  const [storedAccountIdResult, setStoredAccountId] = useAsyncStorage(
+    `${selectedAccountIdKey}:${userId}`,
+  );
 
   /**
    * Initial setting of activeAccount
    */
   useEffect(() => {
     if (
+      !data?.id || // require user id before reading/writing to storage
```
markdlv, line 78: Can we use `userId` instead?

src/screens/ProfileScreen.tsx
```diff
@@ -44,6 +46,9 @@ export const ProfileScreen = () => {
           value={userProfile.familyName}
         />
         <Field label={t('profile-email', 'Email')} value={userProfile.email} />
+        {__DEV__ && (
+          <Field label="Active Project" value={activeProject?.name} />
+        )}
```
markdlv, line 51: 🔥 nice.  I'd love to see us make project display (and even selection for rare multi-project subject cases) a first-class screen (just like accounts) someday.  We setup the hook `useActiveProject` just like `useActiveAccount` for this exact reason - we just haven't made the screen for it yet.  Not suggesting we do it in this PR, soon, or anything - just wanted to signal I'm all for it whenever it's convenient or needed.

src/hooks/useActiveAccount.tsx
bruce-glazier: Opps, yes!

sternetj (approved):

PR #289, opened 2023-07-19
fix: App Tiles can Register a Go Back Handler Now

sternetj:
## Changes

  - Adds logic to handle the back arrow or hardware back button

## Screenshots

<table>
<tr>
 <td><video src="https://github.com/lifeomic/react-native-sdk/assets/2295908/498fa7e1-4872-4528-b3c1-cfdac238104b" />
 <td><video src="https://github.com/lifeomic/react-native-sdk/assets/2295908/05a508a4-61c6-45ab-aea2-7023ad0f3c7b" />
</table>

### DIFF
--- a/src/components/AppNavHeader.tsx
+++ b/src/components/AppNavHeader.tsx
@@ -1,12 +1,19 @@
-import React from 'react';
+import React, { useCallback, useEffect } from 'react';
 
 import { Appbar, Text } from 'react-native-paper';
 import { NativeStackHeaderProps } from '@react-navigation/native-stack';
 import { createStyles, useIcons } from './BrandConfigProvider';
 import { useStyles } from '../hooks/useStyles';
-import { StyleProp, TextStyle, ViewStyle } from 'react-native';
+import {
+  BackHandler,
+  Platform,
+  StyleProp,
+  TextStyle,
+  ViewStyle,
+} from 'react-native';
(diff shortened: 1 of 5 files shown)


markdlv (dismissed):

aymanenadi (dismissed):

sternetj:
@aymanenadi I am actually seeing an issue with my proposed solution where on android, if I enter the app and leave and then come back, the app sends an event with canGoBack equal to true on the loading screen. This breaks the expected behavior. Does this sound like something you ran into before?

I am wondering if it is because the webview is reused or something like that.

aymanenadi:
@sternetj I don't remember TBH, it was a while since I implemented this feature in LX.

Have you considered using the [`beforeRemove`](https://reactnavigation.org/docs/preventing-going-back/) event? At the time, the `react-navigation` version used in LX was old enough that it did not have this event exposed. I wonder if this, combined with `hardwareBackPress`, would get you around this issue.

sternetj:
What ultimately fixed it was providing a `key` to the webView so that it didn't think the previous URI was navigable.

@aymanenadi Thanks for the suggestion. I did switch to using `beforeRemove` and I think the solution is much cleaner since we don't have to have the header know about the webview anymore; plus, React Navigation was complaining about unserializable data in the params with the old solution of passing around the function so this change also resolved that warning.

aymanenadi (approved):

PR #290, opened 2023-07-19
fix: Visual Bug Fixes

sternetj:
## Changes

  - Fixes the Keyboard Avoiding view to not cut off the `Post` button when the text is taller than a few lines
  - Uses `ScrollView` from react native (could cause the app to crash)
  - Requires an invite messaging was showing up briefly while loading the project. Fixes the logic to only show it once projects and accounts are loaded

## Screenshots

image
I think the missing X button is an issue with my set up and can be ignored. I am seeing it show up in other apps using the SDK

### DIFF
--- a/src/components/Circles/CreateEditPostModal.tsx
+++ b/src/components/Circles/CreateEditPostModal.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react';
-import { KeyboardAvoidingView, Platform, View, Modal } from 'react-native';
+import { KeyboardAvoidingView, View, Modal } from 'react-native';
 import {
   Appbar,
   Button,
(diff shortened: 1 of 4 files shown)


src/components/Circles/CreateEditPostModal.tsx
```diff
@@ -94,28 +94,28 @@ export const CreateEditPostModal = () => {
...
-                setPostText(text);
-                setCharacterCount(text.length);
-              }}
-            />
+
             <KeyboardAvoidingView
-              behavior={Platform.OS === 'ios' ? 'padding' : 'height'}
-              keyboardVerticalOffset={Platform.OS === 'ios' ? 20 : 40}
-              style={styles.toolbarContainer}
+              behavior="padding"
```
bruce-glazier, line 99: Did you try this on Android too? I recall there was some major hurdle with getting this to work correctly across both platforms but perhaps I missed something here.

src/navigators/RootStack.tsx
```diff
@@ -37,9 +37,7 @@ export function RootStack() {
 
   const loadingProject = !isFetchedProject || isLoadingProject;
   const loadingAccount = !isFetchedAccount || isLoadingAccount;
-  const hasAccount = !loadingAccount && account?.id;
-  const loadingAccountOrProject =
-    loadingAccount || (hasAccount && loadingProject);
+  const loadingAccountOrProject = loadingAccount || loadingProject;
```
bruce-glazier, line 40: I think this should be fine now that the useSubjectProjects query doesn't load indefinitely if there is no account.

src/components/Circles/CreateEditPostModal.tsx
sternetj: I did try this on android and it seemed like it was working to me. I think moving the view to wrap all the content with the flex grow on the text view made this behave better on both platforms

bruce-glazier: Interesting, sounds good! Glad it is working 👍

bruce-glazier (approved):

src/navigators/RootStack.tsx
markdlv: 🤔 are we sure?  Make sure to test the heck of a pure "user is not in ANY accounts' groups" scenario.  I'm looking at [this line](https://github.com/lifeomic/react-native-sdk/blob/master/src/hooks/useSubjectProjects.tsx#L37) and thinking `loadingProject` will be true until an account.  That's what the more nuanced line 42 was handling.

PR #294, opened 2023-07-21
fix: Account Loading Logic Waits for Selection

sternetj:
## Changes

  - Restores the Root Loading logic
  - Fixes an issue where `useActiveAccount` could be in a state where `isFetched: true` and `isLoading: false` but not actually have set the active account data if possible

### DIFF
--- a/src/hooks/useActiveAccount.tsx
+++ b/src/hooks/useActiveAccount.tsx
@@ -61,6 +61,7 @@ export const ActiveAccountContextProvider = ({
 }) => {
   const accountsResult = useAccounts();
   const accountsWithProduct = filterNonLRAccounts(accountsResult.data);
+  const [isSettled, setIsSettled] = useState(false);
   const [activeAccount, setActiveAccount] = useState<ActiveAccountProps>({});
   const { data: userData } = useUser();
   const userId = userData?.id;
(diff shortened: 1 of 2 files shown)


src/hooks/useActiveAccount.tsx
```diff
@@ -171,7 +183,7 @@ export const ActiveAccountContextProvider = ({
         accountsWithProduct,
         refetch,
         setActiveAccountId,
-        isLoading: accountsResult.isLoading,
+        isLoading: accountsResult.isLoading || !isSettled,
```
markdlv, line 186: Is `isSettled` needed?  `useActiveAccount` should initialize with `isLoading=true`, and RootStack is already paying attention to `isFetched`.

markdlv (approved):

src/hooks/useActiveAccount.tsx
sternetj: Yes it is needed. The purpose of `isSettled` is to sync up the selected account (if there is one) and `isLoading`. 

Before this change isFetched was being set to true and isLoading was set to false before the `useEffect` fired to determine which account to select. In that state, the RootStack logic briefly thought that the accounts had been loaded and that the user did not belong to any accounts and displayed the warning. 

After this change, the RootStack will never get to a state where it thinks the accounts have been loaded without the useActiveAccount logic running and selecting the active account if one exists.

PR #296, opened 2023-07-21
fix: Cache circles by circleId

bruce-glazier:
## Changes

Fix caching of circles when multiple circles are loaded by using the circleId in the `posts` queryKey.


bruce-glazier:
Thought more on this, I am just going to use a circleId context instead of passing around the id everywhere.

PR #299, opened 2023-07-24
fix: Use circleId context for circle caches

bruce-glazier:
## Changes

  - Your changes...


src/components/tiles/TilesList.tsx
```diff
@@ -31,9 +33,10 @@ export function TilesList({ navigation, styles: instanceStyles }: Props) {
 
   const onCircleTilePress = useCallback(
     (circleTile: CircleTile) => () => {
-      navigation.navigate('Home/Circle/Discussion', { circleTile });
+      setCircleTile?.(circleTile);
+      navigation.navigate('Home/Circle/Discussion');
```
sternetj, line 37: I feel like we should move this inside the circle screen. I think it would be better to set the active circle based on the route params. That way other sources, such as notification, can just pass the circleId when navigating to the circles view.

src/hooks/Circles/useActiveCircleTile.tsx
```diff
@@ -0,0 +1,30 @@
...
+  );
+};
+
+export const useActiveCircleTile = () => {
+  const circleTileContext = useContext(CircleTileContext);
+  return {
+    circleTile: circleTileContext?.circleTile,
+    setCircleTile: circleTileContext?.setCircleTile,
+  };
+};
```
sternetj, line 30:
```suggestion
export const useActiveCircleTile = () => useContext(CircleTileContext);
```

src/hooks/Circles/useInfinitePosts.tsx
```diff
@@ -57,7 +59,7 @@ export function useInfinitePosts({ circleId }: useInfinitePostsProps) {
     error: postsError,
     hasNextPage,
     fetchNextPage,
-  } = useInfiniteQuery('posts', queryPosts, {
+  } = useInfiniteQuery(`posts-${circleTile?.circleId}`, queryPosts, {
```
sternetj, line 62: I don't think it is a big deal but should we use `['posts', circleTile?.circleId]` instead? Not quite sure what the advantages could be but I think is is how RQ expects you to use dependencies.

sternetj (dismissed):
This seems like it will work well so I am approving, but, as I mentioned, I wonder if it would be better to move the context setting logic into the Circle/Post Screens instead that way directly routing to it will "just work" and won't require the addition context update step.

With this approach are we still able to rout directly to a Post (for notifications)

src/components/tiles/TilesList.tsx
bruce-glazier: Updated

src/hooks/Circles/useActiveCircleTile.tsx
bruce-glazier: Updated

PR #300, opened 2023-07-24
Use Memo for Account/Project Selection

sternetj:
## Changes

  - Reverts the `isSettled` changes that did not work
  - Switches from state/effect logic to `useMemo` for account and project selection. The old way allowed for some extra renders where `isLoading: false` and `isFetched: true` but the correct account/project had not actually been selected which caused the `need an invitation` warning to be displayed

### DIFF
--- a/src/hooks/useActiveAccount.tsx
+++ b/src/hooks/useActiveAccount.tsx
@@ -4,6 +4,7 @@ import React, {
   useState,
   useContext,
   useCallback,
+  useMemo,
 } from 'react';
 import { Account, useAccounts } from './useAccounts';
 import { QueryObserverResult } from 'react-query';
@@ -61,10 +62,9 @@ export const ActiveAccountContextProvider = ({
 }) => {
   const accountsResult = useAccounts();
   const accountsWithProduct = filterNonLRAccounts(accountsResult.data);
-  const [isSettled, setIsSettled] = useState(false);
-  const [activeAccount, setActiveAccount] = useState<ActiveAccountProps>({});
   const { data: userData } = useUser();
   const userId = userData?.id;
+  const [selectedId, setSelectedId] = useState(accountIdToSelect);
   const [previousUserId, setPreviousUserId] = useState(userId);
   const [storedAccountIdResult, setStoredAccountId] = useAsyncStorage(
     `${selectedAccountIdKey}:${userId}`,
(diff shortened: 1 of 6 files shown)


bruce-glazier (dismissed):
This looks much better!

src/hooks/useActiveAccount.tsx
```diff
@@ -73,60 +73,51 @@ export const ActiveAccountContextProvider = ({
...
     storedAccountIdResult.data,
     storedAccountIdResult.isLoading,
-    accountsResult.isFetched,
-    accountsResult.isLoading,
     userId,
   ]);
 
+  useEffect(() => {
+    if (activeAccount?.account?.id) {
+      setStoredAccountId(activeAccount.account.id);
```
markdlv, line 109: It seems like this could write to async storage more often than we mean to.  You might want to make the dependency on line 111 `activeAccount?.account?.id` instead.

src/hooks/useActiveProject.tsx
```diff
@@ -68,97 +69,98 @@ export const ActiveProjectContextProvider = ({
...
     useMeLoading,
     userId,
+    selectedId,
   ]);
 
+  useEffect(() => {
+    if (hookReturnValue?.activeProject?.id) {
+      setStoredProjectId(hookReturnValue?.activeProject?.id);
+    }
+  }, [hookReturnValue, setStoredProjectId]);
```
markdlv, line 124: Same here as comment in useActiveAccount - might make this first dep `hookReturnValue?.activeProject?.id` instead.

src/hooks/useSubjectProjects.test.tsx
```diff
@@ -57,7 +57,7 @@ test('fetches and returns projects related to useMe', async () => {
   axiosMock.onGet().reply(200, projectsResponse);
   const { result } = await renderHookInContext();
   await waitFor(() => result.current.isSuccess);
-  expect(axiosMock.history.get[0].url).toBe('/v1/projects?id=proj1,proj2');
+  expect(axiosMock.history.get[0].url).toBe('/v1/projects?id=proj1&id=proj2');
```
markdlv, line 60: This doesn't look right to me.  It seems like the previous way was correct.  I'm going to approve but please triple-check this works as expected.

markdlv (dismissed):

src/hooks/useSubjectProjects.test.tsx
sternetj: It was working as expected for me after my change. Looking at the code I think it is expecting it to be in this format.

PR #301, opened 2023-07-24
feat: Youtube video player

aymanenadi:
https://github.com/lifeomic/react-native-sdk/assets/19395435/2af02cfa-c233-4392-83d6-df15048162cf

### DIFF
--- a/src/screens/YoutubePlayerScreen.tsx
+++ b/src/screens/YoutubePlayerScreen.tsx
@@ -1,0 +1,75 @@
+import React from 'react';
+import { NativeStackScreenProps } from '@react-navigation/native-stack';
+import { View, SafeAreaView, StyleSheet } from 'react-native';
+import { WebView, WebViewNavigation } from 'react-native-webview';
+import { HomeStackParamList } from '../navigators';
+
+export const videoUrl = (youtubeVideoId: string) => {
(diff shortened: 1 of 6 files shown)


src/screens/YoutubePlayerScreen.tsx
```diff
@@ -0,0 +1,75 @@
...
+}
+
+export const onShouldStartLoadWithRequest = (
+  videoPlayer: React.RefObject<WebView<{}>>,
+  event: WebViewNavigation,
+) => {
+  if (event.url.indexOf('embed') !== -1 || event.url === 'about:blank') {
+    return true;
+  } else {
+    videoPlayer.current && videoPlayer.current.stopLoading();
```
sternetj, line 23: What does this do for us?

sternetj (approved):

src/screens/YoutubePlayerScreen.tsx
aymanenadi: We used this logic in `life-extend`. I think it's because of this:

> If you want to avoid the the players from being redirected to different URLs from the iframe you can used two methods — onShouldStartLoadWithRequest for iOS and onNavigationStateChange for Android.

Source: https://surajmdurgad.medium.com/embedding-youtube-videos-in-a-react-native-app-8c556a18fd76

PR #303, opened 2023-07-24
Bug-fixes around AdvancedTrackerEditor nav save button

markdlv:


src/components/HeaderSaveButton.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { HeaderButtons, Item } from 'react-navigation-header-buttons';
+import { HeaderButtonComponent } from './HeaderButtonComponent';
+
+interface Props {
+  onPress: () => void;
+}
+
+export const HeaderSaveButton = ({ onPress }: Props) => {
+  return (
+    <HeaderButtons HeaderButtonComponent={HeaderButtonComponent}>
```
sternetj, line 11: Should this use [Appbar.Action](https://callstack.github.io/react-native-paper/docs/components/Appbar/AppbarAction) instead?

sternetj (approved):

src/components/HeaderSaveButton.tsx
markdlv: I thought the same thing, but it seems like `Appbar.Action` only allows icons (and not text like "Save").  I don't love the idea of using our chromicons save icon, or really any icon - I personally prefer the "Save" text here and anywhere else we need it.

sternetj: Ahh nice that makes sense.

sternetj (approved):

PR #304, opened 2023-07-24
chore: use defaults when without profile info

bruce-glazier:
## Changes

- Adds placeholders for when an author's profile information is not available on the post data
- Uses theme title styles for a post author's name

### DIFF
--- a/src/components/Circles/ProfileImage.tsx
+++ b/src/components/Circles/ProfileImage.tsx
@@ -4,6 +4,8 @@ import { Avatar } from 'react-native-paper';
 import { initials } from './initials';
 import type { Post } from '../../hooks/Circles/types';
 import { IconSource } from 'react-native-paper/lib/typescript/src/components/Icon';
+import { createStyles } from '../BrandConfigProvider';
+import { useStyles } from '../BrandConfigProvider/styles/StylesProvider';
 
 type Props = {
   post: Post;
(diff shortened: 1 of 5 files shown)


src/components/Circles/ProfileImage.tsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+export const ProfileImage = ({ post, size, fallbackIcon, style }: Props) => {
+  const { styles } = useStyles(defaultStyles);
+  if (post.author?.profile.picture) {
+    return (
+      <Avatar.Image
+        size={size}
+        style={style}
+        source={{ uri: post.author?.profile.picture }}
```
sternetj, line 24: Should this have a fallback or does that not work if the picture is invalid?

aymanenadi (approved):

sternetj (approved):

src/components/Circles/ProfileImage.tsx
```diff
@@ -0,0 +1,70 @@
...
+        testID={tID('profile-image')}
+        size={size}
+        style={style}
+        label={initials(post?.author?.profile?.displayName)}
+      />
+    ),
+    [post?.author?.profile?.displayName, size, style],
+  );
+
+  if (post.author?.profile.picture) {
```
sternetj, line 33: This could just be `if (post.author?.profile.picture && !imageError) {` which would remove the ternary logic and not require the initial Avatar extraction.

sternetj (approved):

src/components/Circles/ProfileImage.tsx
bruce-glazier: It's so obvious now. I'll update it.

bruce-glazier: Done

sternetj (approved):

PR #308, opened 2023-07-28
chore: Use navigation options to modify header

bruce-glazier:
## Changes

Respect some header navigation options in our custom navigation header:
- headerTitle: Override title with either a title string or components
- headerLeft: Inject buttons/etc into the header in the left position
- headerRight: Inject buttons/etc into the header in the right position
- headerShown: Hide the header as needed

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/76954025/cbfe1c0b-1fbc-47b0-85f9-34983c6476a0

### DIFF
--- a/src/components/AppNavHeader.tsx
+++ b/src/components/AppNavHeader.tsx
@@ -9,6 +9,7 @@ import {
   Platform,
   StyleProp,
   TextStyle,
+  View,
   ViewStyle,
 } from 'react-native';
 import { useDeveloperConfig, useTheme } from '../hooks';
@@ -22,7 +23,15 @@ export function AppNavHeader({
 }: NativeStackHeaderProps) {
   const { styles } = useStyles(defaultStyles);
   const { ChevronLeft } = useIcons();
-  const title = options.title || route.name;
+  const { headerTitle } = options;
+  const { headerString, headerFunction } =
+    typeof headerTitle === 'string'
+      ? { headerString: headerTitle, headerFunction: undefined }
+      : typeof headerTitle === 'function'
+      ? { headerFunction: headerTitle, headerString: undefined }
+      : { headerFunction: undefined, headerString: undefined };
+
+  const title = options.title || headerString || route.name;
   const config = useDeveloperConfig();
   const headerStyles = useColorMapping(
     'backgroundColor',
@@ -55,6 +64,10 @@ export function AppNavHeader({
     return () => handler.remove();
   }, [backNavigationHandler]);
 
+  if (options.headerShown === false) {
+    return null;
+  }
+
   return (
     <Appbar.Header statusBarHeight={statusBarHeight} style={headerStyles}>
       {back ? (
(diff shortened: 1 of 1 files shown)


src/components/AppNavHeader.tsx
```diff
@@ -22,7 +23,15 @@ export function AppNavHeader({
 }: NativeStackHeaderProps) {
   const { styles } = useStyles(defaultStyles);
   const { ChevronLeft } = useIcons();
-  const title = options.title || route.name;
+  const { headerTitle } = options;
+  const { headerString, headerFunction } =
+    typeof headerTitle === 'string'
```
sternetj, line 28: I think the type of `headerTitle` is `string | function` so you could probably reverse this and be able to remove the second ternary.

sternetj (approved):

src/components/AppNavHeader.tsx
bruce-glazier: 👍 Good point

src/components/AppNavHeader.tsx
```diff
@@ -65,13 +78,35 @@ export function AppNavHeader({
           style={styles.backAction}
         />
       ) : null}
+      {options.headerLeft && (
```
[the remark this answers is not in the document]
bruce-glazier: I wasn't sure if there was ever a scenario where someone could want both but I suppose the could always pass a back button into the header left function.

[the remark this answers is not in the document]
sternetj: Yeah that's kind of what I was thinking. If they want to override the back button then they need to provide their own if they want one.

src/components/AppNavHeader.tsx
```diff
@@ -65,13 +78,35 @@ export function AppNavHeader({
...
+            <Title text={title} style={titleStyles} />
+          )
+        }
         style={[styles.contentView, back && styles.contentViewWithBackButton]}
       />
-      {options.headerRight?.({
-        canGoBack: !!back,
-      })}
+      {options.headerRight && (
+        <View>
```
markdlv, line 103: Just curious, why was this extra View needed?  Another approach could be, if/when it is needed, it's provided inside the `headerRight` func provided.  But I'm guessing we're betting this View will be needed 99% of the time?

src/components/AppNavHeader.tsx
```diff
@@ -65,13 +78,35 @@ export function AppNavHeader({
           style={styles.backAction}
         />
       ) : null}
+      {options.headerLeft && (
+        <View>
+          {options?.headerLeft?.({
+            tintColor: options.headerTintColor,
+            canGoBack: !!back,
+          })}
+        </View>
```
markdlv, line 87: Same thing goes here as below - why wrap `headerLeft` in this `View`?

src/components/AppNavHeader.tsx
bruce-glazier: I had intended to do something else with this view that I never did implement, I can probably drop it.

sternetj (approved):

PR #309, opened 2023-07-28
fix: Prevent Writing Same Value to Async Storage

sternetj:


src/hooks/useAsyncStorage.ts
```diff
@@ -6,10 +7,16 @@ export function useAsyncStorage(key: string) {
...
-    );
+  const setItem = useCallback(
+    (value: string) => {
+      if (value !== itemResult.data) {
+        AsyncStorage.setItem(key, value).catch((error) =>
+          console.error(`[LifeOmicSDK]:${error}`),
+        );
+      }
+    },
+    [key, itemResult.data],
```
markdlv, line 18: Does `itemResult.data` somehow update after one calls `setItem('value')`?  I wouldn't think so, so couldn't something like this happen?

- `itemResult.data` is `1`
- `setItem('2')` -> AsyncStorage is used
- `setItem('2')` -> AsyncStorage is used again
- `setItem('2')` -> AsyncStorage is used again

sternetj: Ahh yeah I'll test this but I bet you are correct.

src/hooks/useAsyncStorage.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+const renderHookInContext = async (key: string) => {
+  return renderHook(() => useAsyncStorage(key), {
+    wrapper: ({ children }) => (
+      <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+    ),
+  });
+};
+
+describe('useAsyncStorage', () => {
+  test('configured appTiles are returned', async () => {
```
markdlv, line 23: Copy pasta title?

markdlv (dismissed):

src/hooks/useAsyncStorage.test.tsx
sternetj: 🤦‍♂️ yes

markdlv (approved):

PR #312, opened 2023-08-01
feat: Onboarding course

aymanenadi:
This PR adds an onboarding course screen. This screen is rendered either after accepting the consent if it exists, or after accepting the invitation. When a course is launched, a key is saved in `AsyncStorage` so we don't launch it a second time.

https://github.com/lifeomic/react-native-sdk/assets/19395435/1dca127b-5b41-485b-8077-41456dbb8bfa

### DIFF
--- a/src/hooks/useOnboardingCourse.tsx
+++ b/src/hooks/useOnboardingCourse.tsx
@@ -1,0 +1,73 @@
+import React, { createContext, useCallback, useEffect, useState } from 'react';
+import { useAppConfig } from './useAppConfig';
+import { useActiveProject } from './useActiveProject';
+import { useAsyncStorage } from './useAsyncStorage';
+
+export type OnboardingCourseContextProps = {
+  shouldLaunchOnboardingCourse: boolean;
+  onboardingCourseUrl?: string;
+  onboardingCourseTitle?: string;
+  onOnboardingCourseOpen: () => void;
+  isLoading: boolean;
+  isFetched: boolean;
+  error?: any;
+};
+
+export const OnboardingCourseContext = createContext({
+  shouldLaunchOnboardingCourse: false,
+  onboardingCourseUrl: undefined,
+  onOnboardingCourseOpen: () => {},
+} as OnboardingCourseContextProps);
+
+export const OnboardingCourseContextProvider = ({
+  children,
+}: {
+  children: React.ReactNode;
+}) => {
+  const { data, isLoading, isFetched, error } = useAppConfig();
+  const onboardingCourseUrl = data?.onboardingCourse?.url;
+  const onboardingCourseTitle = data?.onboardingCourse?.title;
+  const { activeProject } = useActiveProject();
+
+  const [storedDidLaunchResult, storeDidLaunch] = useAsyncStorage(
+    `${activeProject?.id}-didLaunchOnboardingCourse`,
+  );
+  const [didLaunchCourse, setDidLaunchCourse] = useState(false);
+
+  useEffect(() => {
+    if (storedDidLaunchResult.isFetched && activeProject?.id) {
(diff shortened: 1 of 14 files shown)


src/screens/OnboardingCourseScreen.tsx
```diff
@@ -0,0 +1,70 @@
...
+    useNavigation<
+      StackNavigationProp<
+        LoggedInRootParamList,
+        'screens/OnboardingCourseScreen'
+      >
+    >();
+
+  return (
+    <HeaderButton
+      title="Done"
```
sternetj, line 26: Needs i18n

src/hooks/useOnboardingCourse.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const [storedDidLaunchResult, storeDidLaunch] = useAsyncStorage(
+    `${activeProject?.id}-didLaunchOnboardingCourse`,
+  );
+  const [didLaunchCourse, setDidLaunchCourse] = useState(false);
+
+  useEffect(() => {
+    if (storedDidLaunchResult.isFetched && activeProject?.id) {
+      setDidLaunchCourse(storedDidLaunchResult.data === 'true');
+    }
+  }, [storedDidLaunchResult, activeProject?.id]);
```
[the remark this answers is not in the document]
aymanenadi: I think I understand what you mean. The 2 scenarios you outlined worked as expected for me during testing. I believe the reason is fetching from the async storage resolves before getting the app config from the server.

I believe what I should do is set `loading = isAppConfigLoading || storedDidLaunchResult.isLoading` to wait for both the async storage and the API queries to resolve. I can set `didLaunchCourse`'s initial value to `undefined` too.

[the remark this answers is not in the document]
sternetj: If it works in testing then I'm good with whatever you want to do.

sternetj (dismissed):

src/hooks/useOnboardingCourse.tsx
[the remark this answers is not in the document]
aymanenadi: I made a new commit to have the hook reflect waiting for both the AppConfig and the Async Storage queries.
I also attached a video showcasing the onboarding screen being launched once.

https://github.com/lifeomic/react-native-sdk/assets/19395435/10b4099a-23ae-4c89-9aad-351b05bf1ea0

sternetj (approved):

src/navigators/RootStack.tsx
```diff
@@ -84,6 +104,8 @@ export function RootStack() {
       ? 'InviteRequired'
       : shouldRenderConsentScreen
       ? 'screens/ConsentScreen'
+      : shouldLaunchOnboardingCourse
+      ? 'screens/OnboardingCourseScreen'
       : 'app';
```
markdlv, line 109: Lot of permutations in here.  New/existing user, has/has not given consent, has/has not seen onboarding course.  @aymanenadi did you manually verify each one works as expected?

aymanenadi: Yes, I think I covered all these cases during my tests. I believe there is room for improvement here.

sternetj: I was talking to Swain about this a little bit and I think it would be awesome if we could make this configurable by a consumer of the sdk. That way they could use developer config to inject their own initial screens if they want and when they are done they can navigate to the app (possibly via a callback we hand them). Using this system we could probably clean up this logic a bit, make it more deterministic, make it easier to test, and allow for customization.

sternetj: https://lifeomic.atlassian.net/browse/PHC-5160

PR #313, opened 2023-08-01
fix: Fix My Data Query for Large Date Ranges

sternetj:
## Changes

  - Default query uses 10 as the page size. This changes the query to query for more data based on days in the interval and the number of codes.

### DIFF
--- a/src/components/MyData/LineChart/useChartData.ts
+++ b/src/components/MyData/LineChart/useChartData.ts
@@ -1,7 +1,7 @@
 import { sortBy } from 'lodash';
 import { useFhirClient } from '../../../hooks';
 import { Trace } from './TraceLine';
-import { startOfDay } from 'date-fns';
+import { startOfDay, differenceInDays } from 'date-fns';
 
 type Props = {
   trace1: Trace;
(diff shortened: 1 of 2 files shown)


src/components/MyData/LineChart/useChartData.ts
```diff
@@ -13,17 +13,21 @@ export const useChartData = (props: Props) => {
   const { trace1, trace2, dateRange } = props;
   const { useSearchResourcesQuery } = useFhirClient();
 
+  const days = Math.abs(differenceInDays(...dateRange));
+
   const trace1Res = useSearchResourcesQuery({
     resourceType: trace1.type,
     coding: trace1.coding,
     dateRange,
+    pageSize: Math.max(days * trace1.coding.length, 10),
```
markdlv, line 22: `trace1.coding` represents an array of coding codes to query, but the results of that query are merged cloud-side.  e.g. `trace1.coding = [ sleep-InBed, sleep-Asleep ]` the results will be all be in the response items array (or however it's structured).  If that assumption is correct, why incorporate codings into pageSize?  Do we try to render each coding separately currently?

markdlv: I feel like I'm still just not understanding.  The bottom line in my mind is we don't know how many results there are until we query.  It could be the 4 records spanning 2 days, or it could be 4000.  By applying a pageSize (explicitly or taking the default), we're (potentially) taking a subset of the results.  Unless we use fhir-search, we won't be sampling evenly across each day either.  Does that make sense?  For a setup with `1` coding and `2` days range - so pageSize=2, and let's say there are `5` records unexpectedly stored each day - we'd be just as likely to get 2 records for a single day as we would 1 for each day.

sternetj: Yeah I agree with this thinking. I'm just not sure how to handle it exactly. Maybe we should exhaustively fetch the data instead? I think in other charting scenarios we have done, we have added the ability to configure a page size based on the user's expected volume of data (message me offline if this is too vague).

markdlv (approved):

PR #315, opened 2023-08-03
PHC-5145 - Allow configuration of image with radial progress

namick:
## Changes

  - Update `TrackAmountControl` and `RadialProgress` to take a `styles` prop from `TrackerDetails`
  - Update `TrackerDetails` with simple target message
  - Update `TrackerDetails` with ability to have product image with radial progress
  - Update styling ability of `TrackerDetails`


src/components/TrackTile/TrackerDetails/TrackerDetails.tsx
```diff
@@ -188,6 +196,11 @@ export const TrackerDetails: FC<TrackerDetailsProps> = (props) => {
     [saveNewValue],
   );
 
+  const trackerImage =
+    typeof tracker.image === 'string' ? { uri: tracker.image } : tracker.image;
+
+  console.log({ selectedUnit, currentValue, currentTarget });
```
bruce-glazier, line 202: Left over log statement

src/components/TrackTile/TrackerDetails/TrackerDetails.tsx
```diff
@@ -50,6 +51,13 @@ export const TrackerDetails: FC<TrackerDetailsProps> = (props) => {
     onError,
     canEditUnit,
   } = props;
+  const { componentProps } = useDeveloperConfig();
+  const {
+    showSimpleTargetMessage,
+    radialProgressStrokeWidth,
+    radialProgressRadius,
+    radialProgressStrokeLinecap,
```
sternetj, line 59: These seem like basic style properties. Should this be part of `useStyles` instead?

sternetj: With our current set up it might make sense to just move these to the styles altogether

sternetj (approved):

src/components/TrackTile/TrackerDetails/TrackerDetails.tsx
namick: Yes, that does make sense and I can update them in a future PR.  The reason I chose to include them as props is because `radius` and `strokeWidth` are already props and I are used to calculate other values in that component and then passed into basic SVG components as props themselves.  This seems seems like the correct decision in that it may be slightly less fragile than fishing them out of styles and performing calculations.  

The argument for including as styles might be that it might be slightly easier to configure without having to pull out developer config `componentProps`.  However, we are already doing that for the partner requesting these changes, I'm not sure if this applies.

sternetj (approved):

PR #317, opened 2023-08-04
Support multiple consent acceptance items

aymanenadi:
I tried to add checkboxes, but I kept getting errors about missing fonts. I can look at that separately.

**Before:**
Screenshot 2023-08-04 at 8 13 02 AM

**After:**
Screenshot 2023-08-04 at 8 56 03 AM

### DIFF
--- a/src/screens/ConsentScreen.tsx
+++ b/src/screens/ConsentScreen.tsx
@@ -79,9 +79,17 @@ export const ConsentScreen = ({
   const consentText = consentToPresent?.form?.item?.find(
     (f) => f.linkId === 'terms',
   )?.text;
-  const acceptText = consentToPresent?.form?.item?.find(
-    (f) => f.linkId === 'acceptance',
-  )?.text;
+  const acceptanceItems = consentToPresent?.form?.item?.filter((f) => {
+    if (f.code) {
+      return f.code.find(
+        (c) =>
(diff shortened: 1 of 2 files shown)


src/screens/ConsentScreen.tsx
```diff
@@ -100,7 +105,14 @@ export const ConsentScreen = ({
         </View>
       </ScrollView>
       <View style={styles.buttonsContainer}>
-        {acceptText && <Text style={styles.acceptText}>{acceptText}</Text>}
+        {acceptanceItems?.map(
```
sternetj, line 108: Should we do a slice here to just grab the first 2?

sternetj (approved):

src/screens/ConsentScreen.tsx
aymanenadi: My original thought is since we limit the consent builder to have a max of 2 acceptance items, we should be safe on the mobile side of things. IF we ever decide to increase that limit, we'll update the consent builder, and we shouldn't have to modify the SDK, and bump the SDK version in the client's mobile app.

aymanenadi: No, but most of users (if not all) use the consent builder.

aymanenadi: I made a new commit to limit to 2 acceptance items max.

PR #321, opened 2023-08-04
chore: Provide a more helpful error message to devs hitting an oidc provider not handing out refresh_token

markdlv:


jkdowdle (approved):

swain (approved):

aecorredor (approved):

bruce-glazier (approved):

swain:
@markdlv non-blocking: I'd think we'd want a `fix:` commit here, so that this change gets released in a new package version

sternetj (approved):

markdlv:
> @markdlv non-blocking: I'd think we'd want a `fix:` commit here, so that this change gets released in a new package version

For me at least, a DX improvement like this better fits a "ride the next release" type of change.  The fact that the SDK is released almost every weekday influences that opinion.

PR #322, opened 2023-08-04
Fix: Make AdvancedTrackerDetails history chart default to same style as TrackerDetails

markdlv:
Can anyone think of a reason not to default these to the same styling?  I think they just accidentally evolved/diverged over time.

## Screenshots
Before:
Simulator Screen Shot - iPhone 14 Pro - 2023-08-04 at 14 16 22

Simulator Screen Shot - iPhone 14 Pro - 2023-08-04 at 14 16 33

After:
Simulator Screen Shot - iPhone 14 Pro - 2023-08-04 at 14 41 58

### DIFF
--- a/src/components/TrackTile/hooks/test/useTrackerValues.test.ts
+++ b/src/components/TrackTile/hooks/test/useTrackerValues.test.ts
@@ -354,17 +354,17 @@ describe('useTrackerValues', () => {
       fetchOntology,
     } as any);
 
-    const valuesContext: TrackerValuesContext = {
+    const customValuesContext: TrackerValuesContext = {
       system: TRACKER_CODE_SYSTEM,
       codeBelow: TRACKER_CODE,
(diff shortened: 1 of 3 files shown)


src/components/TrackTile/hooks/test/useTrackerValues.test.ts
```diff
@@ -354,17 +354,17 @@ describe('useTrackerValues', () => {
       fetchOntology,
     } as any);
 
-    const valuesContext: TrackerValuesContext = {
+    const customValuesContext: TrackerValuesContext = {
```
markdlv, line 357: Unrelated lint warning fix

sternetj (approved):
This makes sense to me. I don't think this has been used in an app yet, so making them the same seems logical.

namick (approved):

PR #324, opened 2023-08-07
feat: Configure LogoHeader per-screen at DeveloperConfig

bruce-glazier:
## Changes

  - Adds stack navigation listeners to sync LogoHeader with options specified in DeveloperConfig

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/76954025/f006f11d-d934-4ac1-9319-bb9b0d7046e1

https://github.com/lifeomic/react-native-sdk/assets/76954025/364cad91-6697-4819-a57c-6a5c6f1fca28

### DIFF
--- a/src/hooks/useLogoHeaderOptions.tsx
+++ b/src/hooks/useLogoHeaderOptions.tsx
@@ -59,7 +59,6 @@ export const navigationScreenListeners = (
   focus: (e: any) => {
     // Coerce unique route event target to route name
     const routeName = e.target?.split('-')?.[0] as Route | undefined;
-    console.log('Found routeName', routeName);
     if (typeof routeName === 'undefined') {
       return;
     }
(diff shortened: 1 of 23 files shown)


namick (approved):

src/hooks/useLogoHeaderOptions.tsx
sternetj, line 1: There is a mix of referencing the Logo header and not referencing it in the names and events in this module. I think with a little rework we could make this all generic so that any screen/component could react to the current route and update the options. That being said, I don't know how likely that scenario is, so it makes sense to just leave it as-is and make this generic the next time we run into this scenario.

sternetj (approved):

src/hooks/useLogoHeaderOptions.tsx
bruce-glazier: I did consider that we could decouple this from the logoHeader should we ever want a simple way to react to screen events but decided we could just cross that bridge when/if we need it. Good feedback, thanks.

PR #327, opened 2023-08-08
feat: Update Tracker Settings Editor

sternetj:
## Changes

  - Updates the tracker settings screen to allow reordering and toggling of all items
  - reworks the installed logic to be based off of a new field added to the tracker install settings so that a user can remove a installed flag if desired

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/2295908/3bab3bfe-b5ab-4a85-9b8c-57ba51db8816

### DIFF
--- a/example/android/app/src/main/java/com/example/RNTLaunchScreen.java
+++ b/example/android/app/src/main/java/com/example/RNTLaunchScreen.java
@@ -23,7 +23,7 @@ public class RNTLaunchScreen extends SimpleViewManager<View> implements ReactPac
         return View.inflate(context, R.layout.splash_screen, null);
     }
 
-    @Overrides
+    @Override
     public List<ViewManager> createViewManagers(ReactApplicationContext reactContext) {
         return Arrays.<ViewManager>asList(new RNTLaunchScreen());
     }
--- a/src/components/TrackTile/ManageTrackers/ManageTrackerRow.tsx
+++ b/src/components/TrackTile/ManageTrackers/ManageTrackerRow.tsx
@@ -15,6 +15,7 @@ import { tID } from '../common/testID';
 import { useIcons, createStyles } from '../../BrandConfigProvider';
 import { useStyles, useTheme } from '../../../hooks';
 import { IconButton } from 'react-native-paper';
+import { Easing } from 'react-native-reanimated';
 
 type Props = Omit<TouchableHighlightProps, 'onPress' | 'onPressIn'> & {
   tracker: Tracker;
@@ -39,10 +40,11 @@ export const ManageTrackerRow: FC<Props> = (props) => {
   useEffect(() => {
     Animated.timing(animation, {
       toValue: !isEditable ? 0 : 1,
-      duration: isEditable ? 400 : 500,
+      duration: 400,
       useNativeDriver: false,
+      easing: isEditable ? Easing.linear : Easing.in(Easing.linear),
     }).start();
-  }, [isEditable]);
+  }, [isEditable, animation]);
 
   return (
     <TouchableHighlight
--- a/src/components/TrackTile/ManageTrackers/ManageTrackerRow.tsx
+++ b/src/components/TrackTile/ManageTrackers/ManageTrackerRow.tsx
@@ -1,9 +1,11 @@
-import React, { FC } from 'react';
+import React, { FC, useEffect, useRef } from 'react';
 import {
   View,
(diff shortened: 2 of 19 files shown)


src/screens/TrackTileSettingsScreen.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { useTrackers } from '../components/TrackTile/hooks/useTrackers';
+import { TrackTileServiceProvider } from '../components/TrackTile/main';
+
+jest.mock('../components/HeaderButton', () => ({
+  HeaderButton: jest.requireActual('react-native').Button,
+}));
+jest.mock('../components/TrackTile/hooks/useTrackers', () => ({
+  useTrackers: jest.fn(() => ({})),
+}));
+jest.mock('i18next', () => ({
```
bruce-glazier, line 14: I'm assuming you mocked this because you are relying on some interpolation that is not available because the set defaultValues should cover most cases. In the few cases it is not covered you could manually add those entries here: https://github.com/lifeomic/react-native-sdk/blob/f1c6bbf1b98132af90732ea650bf106002f8e158/jest.setup.ts#L105 

This, of course, does not really add to the overall value of the test so definitely consider this a nit.

bruce-glazier (approved):

src/screens/TrackTileSettingsScreen.test.tsx
sternetj: I'm actually not sure why I mocked that. Removed!

example/android/app/src/main/java/com/example/RNTLaunchScreen.java
```diff
@@ -23,7 +23,7 @@ public View createViewInstance(ThemedReactContext context) {
         return View.inflate(context, R.layout.splash_screen, null);
     }
 
-    @Overrides
+    @Override
```
markdlv, line 26: Unrelated bug-fix?

src/components/TrackTile/ManageTrackers/ManageTrackerRow.tsx
```diff
@@ -1,41 +1,95 @@
...
+                  inputRange: [0, 1],
+                  outputRange: [0, 24],
+                }),
+              },
+            ]}
+            testID={tID(`toggle-button-${id}`)}
+          >
+            <IconButton
+              onPress={onToggleInstall}
+              icon={() =>
```
markdlv, line 81: I've been seeing this warning (`Do not define components during render.`) a lot more lately.  It kind of forces you to create a component, with a parameter that `tracker.installed` can feed.  I've also been in a situation where I couldn't find a way _out_ of this warning no matter what I did (home screen refresh button).  I think we should definitely let it help nudge us towards preferring components/memos over inline func components like this, though.

src/components/TrackTile/hooks/useTrackers.ts
```diff
@@ -77,36 +73,34 @@ export const useTrackers = (trackersToUse?: Trackers) => {
...
-                    ...omit(t, ['target', 'unit', 'metricId']),
-                    id: tracker.metricId,
-                  };
-                }
-                return t;
-              }),
-            ),
-          );
-        }
+      if (tracker.system === TRACKER_PILLAR_CODE_SYSTEM) {
```
markdlv, line 76: It'll definitely be good to heavily test both trackers and pillars after these changes.

markdlv (dismissed):

src/components/TrackTile/ManageTrackers/ManageTrackerRow.tsx
bruce-glazier: To address I've been wrapping the components with useCallback which suppresses the warning.

example/android/app/src/main/java/com/example/RNTLaunchScreen.java
sternetj: Yeah it is unrelated 😬

src/components/TrackTile/hooks/useTrackers.ts
sternetj: Yeah great point. I think we are fairly insulated here since we don't have the ability to remove a pillar at the moment. This change makes it so we run the logic every time and set `installed` to `false`.

I will definitely play around with pillars before merging just to make sure everything is okay.

markdlv (approved):

PR #328, opened 2023-08-09
Consolidate tab developerConfig settings

bruce-glazier:
## Changes

Consolidates all tab bar related config into a single spot and allows for a more dynamic way to compose tabs.

### DIFF
--- a/src/common/DeveloperConfig.ts
+++ b/src/common/DeveloperConfig.ts
@@ -66,7 +66,6 @@ export type DeveloperConfig = {
   appTileScreens?: AppTileScreens;
   simpleTheme?: SimpleTheme;
   apiBaseURL?: string;
-  additionalNavigationTabs?: NavigationTab[];
   AppNavHeader?: {
     headerColors?: RouteColor[];
     onHeaderColors?: RouteColor[];
@@ -78,11 +77,7 @@ export type DeveloperConfig = {
     };
     TabBar?: {
       showLabels?: boolean;
-      tabs?: [
-        Partial<NavigationTab>,
-        Partial<NavigationTab>,
-        Partial<NavigationTab>,
-      ];
+      tabs?: NavigationTab[];
     };
     TrackerDetails?: {
       showSimpleTargetMessage: boolean;
--- a/src/common/DeveloperConfig.ts
+++ b/src/common/DeveloperConfig.ts
@@ -66,7 +66,6 @@ export type DeveloperConfig = {
   appTileScreens?: AppTileScreens;
   simpleTheme?: SimpleTheme;
   apiBaseURL?: string;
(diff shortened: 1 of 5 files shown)


src/navigators/TabNavigator.tsx
```diff
@@ -78,30 +39,51 @@ export function TabNavigator() {
...
+      />
+      <Tab.Screen
+        name={tabs?.[2].name ?? 'SettingsTab'}
+        component={tabs?.[2].component ?? SettingsStack}
+        key={tabs?.[2].name ?? 'SettingsTab'}
+        options={{
+          tabBarLabel: tabs?.[2].label ?? t('tabs-settings', 'Settings'),
+          tabBarIcon: tabs?.[2].icon ?? Settings,
+          tabBarColor: tabs?.[2].color,
+          headerShown: tabs?.[2].headerShown ?? false,
```
markdlv, line 73: This totally works, but it also feels somewhat fragile around how many items are in `tabs`, right?  Seems ok for a first pass, but we want to make this more dynamic/responsive to the `tabs` count passed in over time, right?

markdlv (dismissed):

src/navigators/TabNavigator.tsx
bruce-glazier: Yeah, that's a great point, and I think I see a clean way to do it and also consolidate the "AdditionalTabs" into it.

src/navigators/TabNavigator.tsx
```diff
@@ -78,37 +40,51 @@ export function TabNavigator() {
...
-              key={tab.name}
-              options={{
-                tabBarLabel: tab.options.tabBarLabel,
-                tabBarIcon: tab.options.tabBarIcon,
-                headerShown: false,
-              }}
-            />
-          ))
-        : defaultTabs()}
-      {!useTabBar &&
```
bruce-glazier, line 95: By the looks of it, the `useTabBar` conditional was to prevent an out-of-bounds error for additional tabs but I've updated the TabBar to force additional tabs to use the same styles from the last tab.

src/common/DeveloperConfig.ts
```diff
@@ -66,8 +66,6 @@ export type DeveloperConfig = {
   appTileScreens?: AppTileScreens;
   simpleTheme?: SimpleTheme;
   apiBaseURL?: string;
-  navigationTabs?: NavigationTab[];
-  additionalNavigationTabs?: NavigationTab[];
```
markdlv, line 70: This is considered a breaking change, removing props.  Definitely not ideal, but I think you & team might be the only one using this atm.  So just expect this to break the build during upgrade.  Down the road, we really need to take a "only add, never remove" approach to these kinds of things 👍

markdlv (approved):

PR #331, opened 2023-08-15
experiment: adding use oidc provider hook based on expo-auth-session

jkdowdle:
## Changes

I've been worried about the idea of adding anything expo specific here but if we did I thought it might make sense to do it under an `/expo` import path or something. Here is an idea of some code that would connect to our `oidc-provider` based on something could be worked into the `AuthConfig` that is already used in the sdk. After our discussion, earlier I doubt we would take this approach but thought it would be worth putting it up to look and see how we might make these interfaces less implementation specific


jest.config.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
     '<rootDir>[/\\\\](|coverage|node_modules)[/\\\\]',
     'src/components/TrackTile/TrackerDetails/AdvancedTrackerEditor/AdvancedTrackerEditor.test.tsx',
   ],
-  preset: 'react-native',
+  preset: 'jest-expo',
```
jkdowdle, line 17: changing this config mostly for experimentation but ideally it would be seperate and only "expo" things would need to use a different config

src/expo/hooks/useOpenIDProvider.ts
```diff
@@ -0,0 +1,76 @@
...
+      codeVerifierRef.current = auth.codeVerifier;
+    }
+  });
+
+  const prompt = async (): Promise<AuthSessionResult> => {
+    const session = await _prompt();
+
+    if (session.type === 'success' && typeof session.params.code === 'string') {
+      const params = session.params as { code: string };
+      const exchange = await exchangeCodeAsync(
```
aecorredor, line 48: nit: maybe worth calling this `authentication` directly.

aecorredor (commented):
Something like this seems right. Have we been able to manually test it (maybe with a simple copy/paste into our second app?)

Also, I think there's nothing much that's LifeOmic specific, so I guess that's good, but this is probably a gateway to adding a full flow? i.e. invites, etc.

jkdowdle:
> Something like this seems right. Have we been able to manually test it (maybe with a simple copy/paste into our fountain app?)

Yeah! let me post a few videos. tested with just copy/paste. I think metro recently started supporting symlinks? so I might be able to locally link the package. I could try that too

> Also, I think there's nothing much that's LifeOmic specific, so I guess that's good, but this is probably a gateway to adding a full flow? i.e. invites, etc.

I am sure this will change after our discussion, but I thought maybe we could have single pieces that deal with that common flow and then compose them 


Starting with an invite that has evc auth + project 

1. sign-in with evc (this pr)
1. store tokens 
1. accept invite

jkdowdle:
@aecorredor demo videos. For android I had to do a release build to get linking back into the app to work as expected

<details>
  <summary>sudo jsx sample code

</summary>
  
```jsx

import { useOpenIDProvider } from '@lifeomic/react-native-sdk/expo'

...

const App = () => {
  const [user, setUser] = useState<{} | null>(null)
  const prompt = useOpenIDProvider(oauthConfig)

  const promptUserWithAuth = async () => {
    const result = await prompt()

    if (result.type === 'success') {
      console.log(result.authentication?.accessToken)
      setUser({})
    } else {
      setUser(null)
    }
  }

  return (
    <>
      <Text>State: {user ? 'Signed In' : 'Signed Out'}</Text>
      <Button
        onPress={() => {
          void promptUserWithAuth()
        }}
      >
        Sign In
      </Button>
    </>
  )
}
  ```
</details>



https://github.com/lifeomic/react-native-sdk/assets/19804196/cf2864a7-6c3d-4d5f-8957-9c90da542e7d

https://github.com/lifeomic/react-native-sdk/assets/19804196/d7da8c34-ba4a-47c7-a5ff-049ce3758918

aecorredor (commented):
Thanks for putting up this example + the screenshots up. I think with the steps we'll be taking, we'll probably integrate this in a reusable way, so it seems like we might not merge this PR directly.

markdlv:
> After our discussion, earlier I doubt we would take this approach but thought it would be worth putting it up to look and see how we might make these interfaces less implementation specific

I'd strongly agree.  This PR is a great experiment and it demonstrates one path that involve (needed?) expo-related packages and a new auth hook.  I'd be pretty bummed about not being able to reuse the existing `useOAuthFlow`, and would be concerned this is a harbinger for needing to re-create most hooks/components/etc.  As we all know, what inevitably happens is a bug-fix goes into (one `useOAuthFlow` or `useOpenIDProvider`) and not the other.  Hopefully we can discover reuse capabilities without the need to re-write - perhaps via more injection interfaces like you guys have mentioned.

PR #332, opened 2023-08-16
fix: Convert to preferred unit in AdvancedTrackerDetails

markdlv:


src/components/TrackTile/TrackerDetails/AdvancedTrackerDetails/AdvancedTrackerDetails.tsx
```diff
@@ -387,7 +387,7 @@ export const AdvancedTrackerDetails = (props: AdvancedTrackerDetailsProps) => {
           hasBorderTop={index === 0}
           color={color}
           title={value.code.coding[0]?.display ?? tracker.name}
-          subTitle={unitDisplay(value.value)}
+          subTitle={unitDisplay(convertToPreferredUnit(value.value, tracker))}
```
markdlv, line 390: In place of screenshot: imagine seeing "30 min" instead of the bug: "1800 min".

swain (approved):

PR #336, opened 2023-08-17
PHC-5213 Add remaining messaging screens and features

bruce-glazier:
## Changes

  - Adds MessageTile and Messages screen to access DMs
  - Adds new queries and UI needed to support unread messages badges
  - Adds theme hooks for styling the messages
  - Queries user data to get profile info to display on messages screen
  
## Possible future PRs
- Unread badge clears on a timer (10 seconds) even if you visit the DM and leave but would be cleaner if it cleared instantly
- Consider implementing Subscriptions API to reduce overhead caused by polling

## Updated screenshots not included in the video
Screenshot 2023-08-18 at 11 59 48 AM
<br>
Screenshot 2023-08-18 at 11 59 57 AM

Screenshot 2023-08-18 at 12 00 11 PM

https://github.com/lifeomic/react-native-sdk/assets/76954025/be89a2b7-3bfe-4593-a195-398105953467


src/components/tiles/TilesList.tsx
```diff
@@ -29,7 +30,7 @@ export function TilesList({ navigation, styles: instanceStyles }: Props) {
   const todayTileEnabled = data?.homeTab?.tiles?.includes?.('todayTile');
   const myDataTileEnabled = data?.homeTab?.tiles?.includes?.('myDataTile');
   const todayTile = data?.homeTab?.todayTile;
-
+  const messagesTileEnabled = data?.homeTab?.tiles?.includes?.('messagesTile');
```
bruce-glazier, line 33: This is the equivalent of feature flagging this off, merging this will not impact anyone.

src/components/MessagesTile/index.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { Tile } from '../tiles/Tile';
+import { HomeStackScreenProps } from '../../navigators';
+import { useUser, useHasNewMessagesFromUsers } from '../../hooks';
+
+interface Props extends Pick<HomeStackScreenProps<'Home'>, 'navigation'> {}
+
+export function MessagesTile({ navigation }: Props) {
+  const { data } = useUser();
+  const mockUserIds = ['brucetest0005', '6fab714']; // TODO: Fetch from tile config
```
bruce-glazier, line 10: Hard-coded for further development until the backend work is finished

src/hooks/Circles/usePrivatePosts.ts
```diff
@@ -81,6 +82,55 @@ export function useInfinitePrivatePosts(userId: string) {
...
+    );
+  };
+
+  return useQueries(
+    (userIds ?? []).map((userId) => {
+      return {
+        queryKey: ['privatePosts', userId],
+        queryFn: () => queryPosts(userId),
+        enabled: !!accountHeaders?.['LifeOmic-Account'],
+        refetchInterval: 10000,
```
bruce-glazier, line 111: Currently the only way badge notifications get displayed and cleared...we can do something about the cleared but the displayed requires polling.

src/hooks/useAsyncStorage.ts
```diff
@@ -12,12 +12,14 @@ export function useAsyncStorage(key: string) {
 
   const setItem = useCallback(
     (value: string) => {
-      if (value !== itemResult.data) {
-        AsyncStorage.setItem(key, value)
-          .catch((error) => console.error(`[LifeOmicSDK]:${error}`))
-          .finally(() => {
```
bruce-glazier, line 18: This was causing an undefined function error that is resolved by using a try/catch here. I checked all the values at the time the error occurs, like AsyncStorage.SetItem but it all appears to be populated with functions as expected.

bruce-glazier: I will also add that I've run into this error unrelated to this PR.

src/components/tiles/TilesList.tsx
namick: This will need to be:
```
const messagesTileEnabled = data?.homeTab?.messageTiles?.length
```

See:
https://github.com/lifeomic/life-research-service/pull/116

src/components/tiles/TilesList.tsx
```diff
@@ -99,6 +100,7 @@ export function TilesList({ navigation, styles: instanceStyles }: Props) {
         />
       )}
       <View style={styles.tilesView}>
+        {messagesTileEnabled && <MessagesTile navigation={navigation} />}
```
namick, line 103: Will need to support multiple message tiles from the `appConfig.homeTab.messageTiles` array.

Also maybe consider renaming `MessagesTile` to `MessageTile` to align with multiple `messageTiles` in the appConfig or the other way around, should we rename it in appConfig?  Or maybe not worry about it at all...  😁

namick (commented):
Looking awesome!  🔥

src/components/tiles/TilesList.tsx
bruce-glazier: 👍 I wasn't worrying about the specifics in this PR but wanted a placeholder here for when we worked those all out. I will certainly update here shortly.

src/components/tiles/Tile.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { TouchableOpacity, View, ViewStyle } from 'react-native';
-import { Text, shadow } from 'react-native-paper';
+import { Badge, Text, shadow } from 'react-native-paper';
```
markdlv, line 3: 🔥

src/hooks/useDirectMessageEndCursor.tsx
```diff
@@ -0,0 +1,81 @@
...
+    fetchEndCursors();
+  });
+
+  return result;
+};
+
+const useGetRemoteCursorsForUsers = ({ userIds }: Props) => {
+  const endCursorMap = new Map<string, string>();
+  const userQueries = usePrivatePostsForUsers(userIds);
+  userQueries.forEach(async (userQuery, index) => {
```
markdlv, line 50: This `forEach(async` looks like it might not actually await them each blocks.  I'd recommend switching to `for .. of` here.

src/screens/DirectMessagesScreen.tsx
```diff
@@ -10,26 +10,36 @@ import { createStyles } from '../components';
...
-  const { styles } = useStyles(defaultStyles);
 
   const { data: userData } = useUser();
   const { data, isLoading, hasNextPage, fetchNextPage, isFetchingNextPage } =
     useInfinitePrivatePosts(recipientUserId);
-  const incomingMessages = data?.pages?.flatMap((page) =>
-    page.privatePosts.edges.flatMap((edge) => postToMessage(edge.node)),
+  const [_, setEndCursor] = useAsyncStorage(
+    `${userData?.id}-${recipientUserId}-endCursor`,
   );
```
markdlv, line 33: Instead of using asyncStorage directly here, can we encapsulate how cursors are stored (and this formatted key) into a helper method, perhaps exposed from `useDirectMessageEndCursor` hook/file?

src/components/MessagesTile/index.tsx
markdlv: 👍 feels like should be able to resolve this TODO before master merge, the way the backend is shaping up.

src/components/tiles/TilesList.tsx
markdlv: +1 on prepping for their possibly being multiple Message(s)Tiles.

Also: I think (as a default) this should go down by the `circleTiles` instead of up at the top.

src/components/MessagesTile/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const recipients = mockUserIds.filter((v) => v !== data?.id);
+  const unreadIds = useHasNewMessagesFromUsers({
```
markdlv, line 12: I think we're good here, but just to say it out loud: it'd be really bad if we started message polling when a MessagesTile isn't needed.  I've traced the polling back to here, which shouldn't be executed unless we render MessagesTile, which I think you have sufficient conditional logic on in TilesList.tsx 👍

src/components/tiles/TilesList.tsx
bruce-glazier: Updated but still hanging on until I know what the final API shakes out to be.

src/components/tiles/TilesList.tsx
bruce-glazier: Moved down to circles tile and added support for multiple tiles but might need to make some edits depending on my assumptions of the API.

src/screens/DirectMessagesScreen.tsx
bruce-glazier: Done, I already a helper for this but failed to utilize it here.

src/screens/MessageScreen.tsx
```diff
@@ -0,0 +1,156 @@
...
+    userLoading
+  ) {
+    return (
+      <ActivityIndicatorView
+        message={t('loading-messages-screen', 'Loading messages')}
+      />
+    );
+  }
+
+  const userList = compact(
```
sternetj, line 84: Might be good to memoize this since useHasNewMessagesFromUsers polls and would cause this to run everytime.

sternetj (approved):

src/screens/MessageScreen.tsx
```diff
@@ -0,0 +1,156 @@
...
+        scrollEventThrottle={400}
+        contentContainerStyle={styles.scrollView}
+      >
+        {prioritizedUserList?.map((user) => (
+          <TouchableOpacity
+            key={`message-${user.userId}`}
+            onPress={handlePostTapped(user.userId, user.displayName)}
+            activeOpacity={0.6}
+          >
+            <List.Item
```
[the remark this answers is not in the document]
bruce-glazier: That would be nice! I **think** that might be achievable without too much of a lift.

src/screens/MessageScreen.tsx
bruce-glazier: Good call, I will grab in a follow up since this has already grown so big.

PR #337, opened 2023-08-17
Support substituting an authorize implementation in

jkdowdle:
## Changes

Getting a rough poc working before PTO for allowing a consumer to replace the `authorize` `react-native-app-auth` implementation. We plan to continue to support that implementation as the default 

This gets us up to the point where we attempt to securely store the tokens, where we will need another place to tie in and allow for changing that implementation

Things to do:

- We can definitely re-consider names of types/variables 
- Define a more robust `ProvidedAuthorizeResult`, maybe something like https://lifeomic.slack.com/archives/C05KMKQNTAR/p1692289334198989?thread_ts=1692288197.696699&cid=C05KMKQNTAR
- We likely also need to sub in implementations for `revoke` and `refresh` and create the interfaces that a consumer can conform to

Tested with https://github.com/lifeomic/fountain-life-member-app/pull/36 after running

```
yarn copyToProject fountain-life-member-app    
```

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/19804196/fa30f10a-e286-4239-a6c3-ef433955882a

### DIFF
--- a/src/hooks/useOAuthFlow.tsx
+++ b/src/hooks/useOAuthFlow.tsx
@@ -7,13 +7,26 @@ import React, {
 import {
   authorize,
   AuthConfiguration,
-  AuthorizeResult,
+  // AuthorizeResult,
   refresh,
   revoke,
 } from 'react-native-app-auth';
 import { AuthResult, useAuth } from './useAuth';
 import { usePendingInvite } from './usePendingInvite';
 import { useQueryClient } from 'react-query';
+
+// Should we re-use the `AuthorizeResult` type exported from `'react-native-app-auth'` as a dev/type dep
+// import type { AuthorizeResult } from 'react-native-app-auth';
+// type ProvidedAuthorizeResult = AuthorizeResult;
+export type ProvidedAuthorizeResult = {
+  accessToken: string;
+  accessTokenExpirationDate: string;
+  idToken: string;
+  refreshToken: string;
+  tokenType: string;
+};
+
+type ProvidedAuthorize = () => Promise<ProvidedAuthorizeResult>;
 
 export interface OAuthConfig {
   login: (params: LoginParams) => Promise<void>;
@@ -22,7 +35,7 @@ export interface OAuthConfig {
 }
 
 export interface LoginParams {
-  onSuccess?: (result: AuthorizeResult) => void;
+  onSuccess?: (result: ProvidedAuthorizeResult) => void;
   onFail?: (error?: any) => void;
 }
 
@@ -39,9 +52,11 @@ const OAuthContext = createContext<OAuthConfig>({
 export const OAuthContextProvider = ({
   authConfig,
   children,
+  providedAuthorize = () => authorize(authConfig),
 }: {
   authConfig: AuthConfiguration;
   children?: React.ReactNode;
+  providedAuthorize?: ProvidedAuthorize;
 }) => {
   const {
     isLoggedIn,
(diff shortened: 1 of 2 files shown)


src/hooks/useOAuthFlow.tsx
```diff
@@ -39,9 +52,11 @@ const OAuthContext = createContext<OAuthConfig>({
 export const OAuthContextProvider = ({
   authConfig,
   children,
+  providedAuthorize = () => authorize(authConfig),
```
bruce-glazier, line 55: You could consider hoisting this injected authorizer into the DeveloperConfig for a simple override without having to rebuild the default providers.

jkdowdle: looking briefly, if I am understanding, that looks great! that way it would be outside of the react context? yeah, we should definitely prefer doing that when possible cc @aecorredor

jkdowdle: And I think we might want something more like `methods: { authorize, refresh, revoke }`

jkdowdle: or maybe you mean just adding to `src/common/RootProviders.tsx`? I agree, we should do that so that users can continue using that root provider and not re-create their own

src/hooks/useOAuthFlow.tsx
```diff
@@ -7,22 +7,35 @@ import React, {
...
-  AuthorizeResult,
+  // AuthorizeResult,
   refresh,
   revoke,
 } from 'react-native-app-auth';
 import { AuthResult, useAuth } from './useAuth';
 import { usePendingInvite } from './usePendingInvite';
 import { useQueryClient } from 'react-query';
 
+// Should we re-use the `AuthorizeResult` type exported from `'react-native-app-auth'` as a dev/type dep
```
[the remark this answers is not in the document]
aecorredor: If we re-use and exposed that type in the implementation that the end consumer uses, we'd be exposing the `react-native-app-auth` details. Maybe we just need to find a common interface and expose that instead?

aecorredor (commented):
@swain lmk once you've taken a look at this. I'm tackling the storage abstraction now, so just want to know if you'd like to keep going with this PR or if I should grab it once I'm free.

aecorredor:
We're also missing drawing out how `refresh` and `revoke` would be overridden.

aecorredor:
After meeting up to talk about where this idea was going, we didn't love the end state of it: an SDK where consumers were forced to do one of two things:
1. Implement their own SecureStore + OAuth flow
2. Import our default implementation from another place in the sdk, something like `@lifeomic/react-native-sdk/secure-store` or `@lifeomic/react-native-sdk/oauth`. The reason for that is that in the current way Josh had it set up, we're passing a defulat implementation through a default function parameter, which still requires us to import `react-native-app-auth`, and that in itself would cause consumers to also have to install that.

Today we're pushing through getting the SDK to just work out of the box, like we were doing yesterday mid-day. Conceptually, we could do that. We're just missing a few config pieces.

aecorredor:
After meeting up to talk about where this idea was going, we didn't love the end state of it: an SDK where consumers were forced to do one of two things:
1. Implement their own SecureStore + OAuth flow
2. Import our default implementation from another place in the sdk, something like `@lifeomic/react-native-sdk/secure-store` or `@lifeomic/react-native-sdk/oauth`. The reason for that is that in the current way Josh had it set up, we're passing a defulat implementation through a default function parameter, which still requires us to import `react-native-app-auth`, and that in itself would cause consumers to also have to install that.

Today we're pushing through getting the SDK to just work out of the box, like we were doing yesterday mid-day. Conceptually, we could do that. We're just missing a few config pieces.

PR #338, opened 2023-08-18
feat!: upgrade to react-query v4

swain:
**Note**: I _strongly_ recommend reviewing _by commit_. I split the commits into 3:
- making the package.json-level changes for the upgrading
- replacing the package name at all the import sites
- making the functional changes required to keep the same behavior while upgrading to v4.

## Changes
- Upgrading the devDependency + peerDependency for react-query to v4.

**Why**
I'm interested in testing out an approach for using `one-query` to reduce a lot of the boilerplate in interacting with APIs . `one-query` requires react-query v4.

**How**
I followed the upgrade guide here: https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4

**Misc**
- This is a breaking change. I have a follow-up PR which adds guidance for consumers that need to make the upgrade: #339


aecorredor (approved):
LGTM

markdlv (approved):
_Beautifully_ separated by commit- thanks for doing that.

bruce-glazier (approved):

PR #341, opened 2023-08-21
chore: add core one-query integration

swain:
## Changes
This PR proposes the foundational files for using `one-query` in this project. This PR is intimately tied to its followup, which actually _uses_ one-query for the first time: #342.

Overall, I think that using one-query will _really_ help to:
- _Reduce boilerplate_ required for integrating with new API endpoints (you _only_ need to update the master type definition to integrate with a new endpoint)

- _Encourage consistency_ in API interaction. All API calls go through the same flow, and are defined + implemented the same way.

Both of these wins contribute to delivering the big one: _help us move faster_ 😄

## Screenshots
See #342 for a demo video showcasing the developer experience of using the new utilities.

### DIFF
--- a/src/test-utils/rest-api-mocking.ts
+++ b/src/test-utils/rest-api-mocking.ts
@@ -1,0 +1,6 @@
+import { createAPIMockingUtility } from '@lifeomic/one-query/test-utils';
+import { RestAPIEndpoints } from '../types/rest-types';
+
+export const useRESTAPIMocking = createAPIMockingUtility<RestAPIEndpoints>({
+  baseUrl: 'https://api.us.lifeomic.com',
+});
--- a/jest.config.js
+++ b/jest.config.js
@@ -40,6 +40,6 @@ module.exports = {
     },
   },
   transformIgnorePatterns: [
-    'node_modules/(?!(jest-)?@?react-native|query-string|decode-uri-component|split-on-first|filter-obj|victory)',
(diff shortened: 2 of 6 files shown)


bruce-glazier (approved):

aecorredor (approved):

src/test-utils/rest-api-mocking.ts
```diff
@@ -0,0 +1,6 @@
+import { createAPIMockingUtility } from '@lifeomic/one-query/test-utils';
+import { RestAPIEndpoints } from '../types/rest-types';
+
+export const useRESTAPIMocking = createAPIMockingUtility<RestAPIEndpoints>({
+  baseUrl: 'https://api.us.lifeomic.com',
```
markdlv, line 5: Whether in this PR or a future one, let's make this a default, and allow for `apiBaseUrl` from `DeveloperConfig` override it.

```
const { apiBaseURL } = useDeveloperConfig();
...
baseUrl: apiBaseURL ?? 'https://api.us.lifeomic.com',
```

markdlv (approved):

src/test-utils/rest-api-mocking.ts
swain: Sure, I'm open to trying to inherit it here somehow. I'll keep an eye on for a future PR.

PR #342, opened 2023-08-21
feat!: convert two hooks to one-query

swain:
## Changes
This PR builds off of #341 by adopting one-query for two existing endpoints:
- the `GET /v1/accounts` endpoint
- the `PATCH /v1/invitations/:inviteId` endpoint

The brief demo video below showcases the nice DX behavior of the package.

Overall, I think that using one-query will _really_ help to:
- _Reduce boilerplate_ required for integrating with new API endpoints (you _only_ need to update the master type definition to integrate with a new endpoint)

- _Encourage consistency_ in API interaction. All API calls go through the same flow, and are defined + implemented the same way.

Both of these wins contribute to delivering the big one: _help us move faster_ 😄

## Screenshots
This demo video showcases the developer experience in-code when using the endpoints added in this PR.

**Turn the audio on** 😄

https://github.com/lifeomic/react-native-sdk/assets/14932834/9ee7c482-0e41-4910-b337-5fd65f7f6157


bruce-glazier (dismissed):
I honestly love the looks of this. A solid improvement to the developer experience.

src/components/Invitations/InviteProvider.tsx
```diff
@@ -38,7 +39,10 @@ export const InviteProvider = ({ children }: ProviderProps) => {
   const acceptInvite = useCallback(
     async (inviteId: string) => {
       try {
-        const acceptedInvite = await mutateAsync(inviteId);
```
aecorredor, line 41: How was this one previously setting the `ACCEPTED` status?

src/hooks/useActiveAccount.test.tsx
```diff
@@ -83,52 +85,59 @@ test('without provider, methods fail', async () => {
...
+
 test('exposes some props from useAccounts', async () => {
-  const error = new Error('uh oh');
-  useAccountsMock.mockReturnValue({
-    isLoading: true,
-    isFetched: true,
-    error,
+  api.mock('GET /v1/accounts', async () => {
+    // wait to simulate a loading state
+    await wait(200);
```
aecorredor, line 102: Do we definitely need this explicit wait?

aecorredor (dismissed):

jkdowdle (approved):

src/types/rest-types.ts
```diff
@@ -1,8 +1,32 @@
...
+
 /**
  * This type definition describes known endpoints in our REST API.
  *
  * It is structured in the format specified by [one-query](https://github.com/lifeomic/one-query).
  */
 export type RestAPIEndpoints = {
-  // TODO: see follow-up PR.
+  'GET /v1/accounts': {
+    Request: {};
```
sternetj, line 22: So if we wanted to require the account header, could we add it here as a required part of the request?

swain: @sternetj Great question. short answer: no -- `one-query` intentionally doesn't really deal in headers. So, this PR changes nothing about how we'd specify the LifeOmic-Account header for requests -- it can still be provided via `one-query` using the `axios` option, like so:

```typescript
useRestQuery(
  'GET ...',
   { ... },
   {
     axios: { // <-- a special one-query option available adjacent to the standard react-query options
       headers: { ... } // <-- provide header here.
     }
   }
)
```

Longer answer: the "idea" behind one-query is mainly about encouraging a consistent pattern of API interaction. It does this by forcing us to provide a single Axios client that can make all the HTTP requests. In use cases up until now, one-query has only been used in scenarios where we should always be providing the header (rather than only in certain endpoints), so we've always built it in the client directly.

Looking at the SDK code + the various API calls it makes, it seems like it is _much_ more common to _use_ the LifeOmic-Account header than to _not_ use it -- there are only a couple of endpoints that don't need it. It seems this has resulted in code like this being duplicated many times:

```typescript
export function useMyQuery() {
  const { accountHeaders } = useActiveAccount(); // <-- get the account header

  return useQuery(
    ['...'],
    () =>
      httpClient
        .get(`/v1/my-url`, { headers: accountHeaders }) // <-- specify it here
        .then((res) => res.data),
    {
      enabled: !!accountHeaders, // <-- don't forget to disable the query in case there is no header
    },
  );
}
```

To me, it seems like it'd be worth optimizing for the common case (using the header) by:
- "building in" this account header logic to the core HTTP client, so that we don't have to so regularly duplicate this code
AND
- providing an escape hatch to allow _not_ specifying the header in scenarios where it shouldn't be provided

I've spent a bit of time trying to make this^ work, and have struggled just a bit. I'll do some more thinking on it. Regardless, I'd like to treat it as "outside the scope of this PR", if that's alright with you.

sternetj: Absolutely! Definitely don't want to hold up the PR for this.
Yeah that pattern is annoying but understandable and as you said it would be nice to switch to an escape hatch model.

sternetj (dismissed):

swain:
@sternetj @aecorredor @jkdowdle @bruce-glazier Just had to re-request reviews because the targeted PR merged. No new changes since your approvals.

bruce-glazier (approved):

swain:
Merging, since several others already approved this changeset.

PR #346, opened 2023-08-24
feat: allow for optional HomeScreen refresh button

markdlv:
https://github.com/lifeomic/react-native-sdk/assets/1445395/d0e562f7-496e-4d7a-aeb3-eb6d21ed2e37

### DIFF
--- a/src/common/RefreshNotifier.tsx
+++ b/src/common/RefreshNotifier.tsx
@@ -1,0 +1,25 @@
+import { EventEmitter } from 'events';
+
+export type RefreshParams = {
+  context?: 'HomeScreen';
+};
+
+export const refreshEventType = 'refreshRequested';
+export type RefreshHandler = (refreshParams: RefreshParams) => void;
+export class RefreshNotifier {
+  private emitter = new EventEmitter();
+
+  public addListener(listener: RefreshHandler) {
+    return this.emitter.addListener(refreshEventType, listener);
+  }
+
+  public removeListener(listener: RefreshHandler) {
+    return this.emitter.removeListener(refreshEventType, listener);
+  }
+
+  public emit(params: RefreshParams) {
+    return this.emitter.emit(refreshEventType, params);
+  }
+}
+
+export const refreshNotifier = new RefreshNotifier();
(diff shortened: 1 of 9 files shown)


src/screens/HomeScreen.tsx
```diff
@@ -7,14 +7,27 @@ import { TilesList } from '../components/tiles/TilesList';
...
 
 export const HomeScreen = (navProps: HomeStackScreenProps<'Home'>) => {
   const { isLoading: loadingAccount } = useActiveAccount();
   const { isInitialLoading: loadingAppConfig, data: appConfig } =
     useAppConfig();
   const { isInitialLoading: loadingJoinCircles } = useJoinCircles();
+  const [refreshing, setRefreshing] = useState(false);
   const { navigation } = navProps;
   const customHeaderTitle = appConfig?.homeTab?.screenHeader?.title;
+  const headerRefreshEnabled = appConfig?.homeTab?.screenHeader?.enableRefresh;
```
markdlv, line 21: It's important to me that for apps that _don't_ want/need this refresh, there's hardly any cost.  The current cost is the additional useEffect & useCallback quick no-op invocations, which I'm thinking is a solid first pass to see how/where this is utilized.

src/components/TrackTile/hooks/useAxiosTrackTileService.ts
```diff
@@ -67,6 +71,19 @@ export const useAxiosTrackTileService = (): TrackTileService => {
     }
   }, [userId, cache, previousUserId]);
 
+  // Clear cached tracker values during HomeScreen refresh
+  useEffect(() => {
+    const refreshHandler = (refreshParams: RefreshParams) => {
+      if (refreshParams.context === 'HomeScreen') {
+        cache.trackerValues = {};
```
sternetj, line 78: Is the thinking that trackers and ontologies won't change enough to warrant clearing the cache?

src/common/RefreshNotifier.tsx
```diff
@@ -0,0 +1,25 @@
...
+  public addListener(listener: RefreshHandler) {
+    return this.emitter.addListener(refreshEventType, listener);
+  }
+
+  public removeListener(listener: RefreshHandler) {
+    return this.emitter.removeListener(refreshEventType, listener);
+  }
+
+  public emit(params: RefreshParams) {
+    return this.emitter.emit(refreshEventType, params);
```
sternetj, line 21: I wonder if there is a way we could make this an async function so that a call could determine when it was all settled. Maybe each listener would return a promise or something. 

Probably don't need it for now but it bugs me that we just have a timeout for the refresh control loading.

sternetj (approved):
Looks great!

namick (approved):

src/components/TrackTile/hooks/useAxiosTrackTileService.ts
markdlv: Yep, that's exactly what I landed on.  I could see this becoming config-based if we need a deeper refresh in some cases.

PR #347, opened 2023-08-24
fix: Only render some items on message screen

bruce-glazier:
## Changes

  - Adds chunking to the message screen so all users in the cohort aren't rendered at once
  - Fixes app performance issues where 100+ users are available to message

### DIFF
--- a/src/hooks/Circles/usePrivatePosts.ts
+++ b/src/hooks/Circles/usePrivatePosts.ts
@@ -119,7 +119,10 @@ export function useInfinitePrivatePosts(userId: string) {
   This hook will query continuously so use it sparingly and only
   when absolutely necessary.
 */
-export function usePollPageInfoForUsers(userIds?: string[]) {
+export function usePollPageInfoForUsers(
+  userIds?: string[],
+  refetchInterval: number = 10000,
+) {
   const { graphQLClient } = useGraphQLClient();
   const { accountHeaders } = useActiveAccount();
   const { data } = useUser();
@@ -145,7 +148,7 @@ export function usePollPageInfoForUsers(userIds?: string[]) {
         queryKey: ['privatePosts', userId],
         queryFn: () => queryPosts(userId),
         enabled: !!accountHeaders?.['LifeOmic-Account'],
-        refetchInterval: 10000,
+        refetchInterval,
         // Continuously polls while query component is focused
         // so new unread messages can be detected
       };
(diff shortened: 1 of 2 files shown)


src/hooks/Circles/usePrivatePosts.ts
```diff
@@ -170,11 +175,11 @@ export const useLookupUsers = (userIds: string[]) => {
...
 
   return useQueries({
-    queries: (userIds ?? []).map((userId) => {
+    queries: (userList ?? []).map((user) => {
       return {
-        queryKey: ['user', userId],
-        queryFn: () => queryUser(userId),
-        enabled: !!accountHeaders?.['LifeOmic-Account'],
+        queryKey: ['user', user.userId],
+        queryFn: () => queryUser(user.userId),
```
sternetj, line 181: Does this combine multiple requests or do we do N network requests for N users?

src/screens/MessageScreen.tsx
```diff
@@ -87,31 +133,37 @@ export function MessageScreen({
...
+  const handleScroll = useCallback(
+    (event: NativeSyntheticEvent<NativeScrollEvent>) => {
+      if (
+        event.nativeEvent.contentOffset.y +
+          event.nativeEvent.layoutMeasurement.height >=
+        event.nativeEvent.contentSize.height
+      ) {
+        if (recipientsListChunked.length > scrollIndex) {
+          // Add new users to view on scroll
+          setScrollIndex(scrollIndex + 1);
```
sternetj, line 145: This could prevent an extra render.
```suggestion
          setScrollIndex(currentIndex => currentIndex + 1);
```

src/screens/MessageScreen.tsx
```diff
@@ -18,21 +22,63 @@ export function MessageScreen({
...
+
+  // Break-down list of users into smaller chunks
+  const recipientsListChunked = useMemo(
+    () => chunk(sortedList, 10),
+    [sortedList],
+  );
+
+  // Use current scroll-index to expand the list of users
+  // in view up to the maximum
+  const usersInView = useMemo(
```
sternetj, line 50: Does this mean that we won't fetch the unread status until the user scrolls so that a recipient is in view?

sternetj (approved):

src/hooks/Circles/usePrivatePosts.ts
bruce-glazier: We do N network requests for N users based on which users have the enabled flag set to true.

src/screens/MessageScreen.tsx
bruce-glazier: I think not based on this logic:
- The recipientListChunked is created using the sortedList 
- The sortedList is created by sorting the recipientsUserIds based on the unreadIds
- The unreadIds checks for unread messages from every user

namick (approved):

PR #348, opened 2023-08-25
fix: Set User's Real Name on Launch if Missing

sternetj:
## Changes

  - Adds a call for updating a user's profile info
  - Adds an effect hook to set the user's last/first name based on the patient if it has a name already


src/hooks/useMe.ts
```diff
@@ -40,9 +32,10 @@ export function useMe() {
             (entry) =>
               ({
                 subjectId: entry.resource.id,
-                projectId: entry.resource.meta.tag.find(
+                projectId: entry.resource.meta?.tag?.find(
```
markdlv, line 35: 😬 that's a good fix.

namick (approved):

PR #349, opened 2023-08-25
chore: use push notifications to set unread badges

bruce-glazier:
## Changes

Uses onNotificationReceived to and the notification payload to track when users have sent new messages

**Features**
- Stores userIds in state and exposes a hook via contextProvider so they can be read from any screen
- Syncs unreads with async-storage so they persist between app launches
- Clears unread notifications from state/storage when a message is read
 
**Still to determine**
- Handling background notifications: This does not handle push notifications received when the app is backgrounded/closed. Background notification handlers for  _should_ be working so that needs further investigation.
- Handling closed notifications: This will require some other solution to handle, needs further evaluation.


namick (dismissed):

src/hooks/useUnreadMessages.tsx
```diff
@@ -0,0 +1,70 @@
...
+          ? uniq([...unreadIds, incomingUserId])
+          : [incomingUserId];
+      };
+      setUnreadMessages((currentUnreads) => {
+        const newUserIds = computeIds(currentUnreads);
+        setStoredUnreadMessages(newUserIds.join('#'));
+        return newUserIds;
+      });
+    }
+  });
```
sternetj, line 48: Should this be in a `useEffect` so we only register the handler once?

src/hooks/useUnreadMessages.tsx
```diff
@@ -0,0 +1,70 @@
...
+    'UnreadMessageUserIds',
+  );
+
+  const [unreadMessagesUserIds, setUnreadMessages] = useState<string[]>([]);
+
+  useEffect(() => {
+    if (initialUnreadMessages.isFetchedAfterMount) {
+      setUnreadMessages(
+        initialUnreadMessages.data
+          ? initialUnreadMessages.data?.split('#')
```
sternetj, line 28: Why split by `#`? Why not just stringify the JSON?

src/hooks/useUnreadMessages.tsx
```diff
@@ -0,0 +1,70 @@
...
+  children?: React.ReactNode;
+}) => {
+  const [initialUnreadMessages, setStoredUnreadMessages] = useAsyncStorage(
+    'UnreadMessageUserIds',
+  );
+
+  const [unreadMessagesUserIds, setUnreadMessages] = useState<string[]>([]);
+
+  useEffect(() => {
+    if (initialUnreadMessages.isFetchedAfterMount) {
```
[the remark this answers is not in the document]
bruce-glazier: My understanding and testing of `isFetchedAfterMount` is it the boolean will only be true when the data was first fetched as soon as the component mounted. At initially created an infinite render loop but was able to solve it with that.

src/hooks/useUnreadMessages.tsx
bruce-glazier: That was a pretty obvious miss on my part, I will update.

src/hooks/useUnreadMessages.tsx
bruce-glazier: Yes, probably, but this particular code will no longer exists in the latest commit.

src/common/LoggedInProviders.tsx
```diff
@@ -1,13 +1,19 @@
...
+import { UnreadMessagesContextProvider } from '../hooks';
 
 export const LoggedInProviders = ({
   children,
 }: {
   children?: React.ReactNode;
 }) => {
   return (
-    <ActiveAccountContextProvider>{children}</ActiveAccountContextProvider>
+    <NotificationsManagerProvider>
```
bruce-glazier, line 13: I struggled to get the useNotification query to actually wait until the app was authenticated on initial launch. The `enabled` flag did not appear to be respected (which probably needs a closer look). Moving these providers here works well and is really where they below.

src/hooks/useUnreadMessages.tsx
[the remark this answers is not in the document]
sternetj: It wasn't super clear to me which is why I asked. Honestly, I still can't make sense of their [docs](https://tanstack.com/query/v4/docs/react/reference/useQuery) 😂 
> `isFetchedAfterMount: boolean`
> 	- Will be true if the query has been fetched after the component mounted.
> 	- This property can be used to not show any previously cached data.

src/common/LoggedInProviders.tsx
sternetj: I feel like this is an awesome pattern and we should adopt this for more cases.

src/hooks/useNotificationManager.tsx
```diff
@@ -0,0 +1,147 @@
...
+  lastReadAt: Date | null | undefined;
+  privatePostsUserIds: string[] | undefined;
+}
+
+export const NotificationsManagerContext =
+  createContext<NotificationsManagerContextType>({
+    unreadCount: 0,
+    privatePostsUserIds: [],
+    setNotificationsRead: () => {},
+    lastReadAt: null,
```
sternetj, line 36: Should this be `undefined` per your comment above?

src/hooks/useNotificationManager.tsx
```diff
@@ -0,0 +1,147 @@
...
+  useEffect(() => {
+    if (lastNotificationReadTime.isFetchedAfterMount) {
+      updateLastReadAt(
+        lastNotificationReadTime.data
+          ? new Date(lastNotificationReadTime.data)
+          : null,
+      );
+    }
+  }, [
+    lastNotificationReadTime.data,
```
sternetj, line 73: Should `updateLastReadAt` be included here?

src/hooks/useNotificationManager.tsx
```diff
@@ -0,0 +1,147 @@
...
+   * closure created grabbing the last read time from async storage
+   */
+  const lastReadAtRef = useRef<NotificationsManagerContextType['lastReadAt']>();
+  const [lastReadAt, setLastReadAt] =
+    useState<NotificationsManagerContextType['lastReadAt']>();
+
+  const updateLastReadAt = (readAt: Date | null) => {
+    setLastReadAt(readAt);
+    lastReadAtRef.current = readAt;
+  };
```
sternetj, line 59: useCallback?

src/hooks/useNotificationManager.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+export const NotificationsManagerContext =
+  createContext<NotificationsManagerContextType>({
+    unreadCount: 0,
+    privatePostsUserIds: [],
+    setNotificationsRead: () => {},
+    lastReadAt: null,
+  });
+
+export const NotificationsManagerProvider = ({
```
sternetj, line 39: This name seems a little vague since this provider deals with private messages directly. Is the thinking that this would be expanded to other types so that it is responsible for managing all types of notifications and tracking the last read information?

src/hooks/useNotifications.tsx
```diff
@@ -82,24 +97,41 @@ const notificationQuery = gql`
...
   'PostReplyNotification',
   'CircleAdminPostNotification',
   'SurveyAssignedNotification',
-];
+  'PrivatePostNotification',
+] as const;
+
+export type SupportedNotifications = (typeof supportedNotifications)[number];
+
+export const supportedNotifications = supportedNotificationsConst.map(
```
sternetj, line 124: Why did we need to do this? Do our typings not include a `__typename` for `PrivatePostNotification`?

src/hooks/useUnreadMessages.tsx
sternetj: I wonder if it would make sense to move the serialization into `useAsyncStorage`.

sternetj (dismissed):

src/hooks/useNotificationManager.tsx
bruce-glazier: This was the original name given from where I copied this code over from. The idea is that this will eventually be more for the notifications tab and tracking new notifications from the feed rather than used for private messaging but I thought it was OK to do it this way for a first pass.

src/hooks/useNotificationManager.tsx
bruce-glazier: Yes, indeed

src/hooks/useNotificationManager.tsx
bruce-glazier: The linter was even warning about it. Fixed.

src/hooks/useUnreadMessages.tsx
bruce-glazier: I attempted to move this into useAsyncStorage for this PR but hit a bit of a wall on deserialization and type casting, decided to revisit this later.

src/hooks/useUnreadMessages.tsx
```diff
@@ -0,0 +1,70 @@
...
+    }
+  });
+
+  const markMessageRead = (userId: string) => {
+    if (unreadMessagesUserIds.includes(userId)) {
+      setUnreadMessages((currentIds) =>
+        currentIds?.filter((id) => id !== userId),
+      );
+    }
+  };
```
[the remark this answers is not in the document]
bruce-glazier: Done

src/hooks/useNotifications.tsx
bruce-glazier: I updated this with a solution but wondering if I might be missing something more concise/obvious? Let me know what you think.

src/hooks/useUnreadMessages.tsx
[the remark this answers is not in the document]
bruce-glazier: FWIW, the docs are not super clear to me either but I assumed the difference between "isFetched" and "isFetchedAfterMount" here and it seems to work how I was expecting it to.

src/hooks/usePushNotifications.tsx
```diff
@@ -1,9 +1,9 @@
 import React, { createContext, useContext, useEffect } from 'react';
 import {
-  PushNotificationsConfig,
   registerDeviceToken,
   requestNotificationsPermissions,
-} from '../../src/common';
+} from '../../src/common/Notifications';
```
bruce-glazier, line 5: This reduces the imported dependencies when importing this hook and makes unit testing more concise.

src/hooks/useNotificationManager.tsx
```diff
@@ -0,0 +1,165 @@
...
+    updateLastReadAt(new Date());
+  }, [updateLastReadAt]);
+
+  let unreadCount: number | undefined;
+  let privatePostsUserIds: string[] | undefined;
+
+  const notificationsToAuthorId = (_notifications: FeedNotification[]) =>
+    compact(
+      _notifications.map((n) => {
+        if (n.__typename === 'PrivatePostNotification') {
```
sternetj, line 133: If we make `__typename` a property of each notification type, then we won't need to do casting if we first check the type name.

src/hooks/useNotifications.tsx
```diff
@@ -82,24 +97,41 @@ const notificationQuery = gql`
...
+            }
+          }
         }
       }
     }
   }
 `;
 
-const supportedNotificationTypes = [
+export type SupportedNotifications = [
```
sternetj, line 115: If we move the type name to the notification types. This could become
`export type SupportedNotifications = FeedNotification['__typename'];`

src/hooks/useNotifications.tsx
```diff
@@ -82,24 +97,41 @@ const notificationQuery = gql`
...
   'CircleAdminPostNotification',
   'SurveyAssignedNotification',
+  'PrivatePostNotification',
 ];
+
+function isSupportedNotification(
+  __typename: any,
+): __typename is SupportedNotifications {
+  return true;
+}
```
sternetj, line 126: Is this right? We want it to always be true?

src/screens/MessageScreen.tsx
```diff
@@ -1,37 +1,112 @@
...
 
+  const { styles } = useStyles(defaultStyles);
   const { User } = useIcons();
+  const [scrollIndex, setScrollIndex] = useState<number>(0);
+
+  // Get userIds that have sent new messages
+  const { unreadMessagesUserIds: unreadUserIds } = useUnreadMessages();
+
+  // Sort recipientsList by unread messages
+  const sortedList = useMemo(
```
namick, line 40: In the future it might be nice to move some of this work into hooks so that this component can be more presentational and we don't need `useMemo` so much.  This component will likely get much more complicated as we start adding view and style options.

src/screens/MessageScreen.tsx
```diff
@@ -1,37 +1,112 @@
...
 import { tID } from '../common';
+import { useUnreadMessages } from '../hooks/useUnreadMessages';
+import { useNotificationManager } from '../hooks/useNotificationManager';
+import { useFocusEffect } from '@react-navigation/native';
 
 export function MessageScreen({
   navigation,
   route,
 }: HomeStackScreenProps<'Home/Messages'>) {
   const { recipientsUserIds } = route.params;
```
namick, line 25: We can move this into a hook just get the `messageTile`'s `id` from params.  (In a future PR)

namick (dismissed):
LGTM!

src/hooks/useNotifications.tsx
```diff
@@ -82,24 +97,41 @@ const notificationQuery = gql`
...
+
 const selectNotifications = (
   data: NotificationQueryResponse,
 ): NotificationQueryResponse => {
   return {
     notificationsForUser: {
       edges: data.notificationsForUser.edges.filter((edge) =>
```
sternetj, line 133: Are notifications sorted by most recent? If not that would be nice so the most recent message would show up at the top of the list

src/screens/MessageScreen.tsx
bruce-glazier: Indeed, great suggestion. I'll hopefully be doing some refactor/cleanup work in the near future.

src/screens/MessageScreen.tsx
bruce-glazier: 👍

src/screens/MessageScreen.tsx
```diff
@@ -1,37 +1,112 @@
...
+        recipientsUserIds,
+        (userId) => unreadUserIds?.includes(userId),
+        'desc',
+      ),
+    [unreadUserIds, recipientsUserIds],
+  );
+
+  // Break-down list of users into smaller chunks
+  const recipientsListChunked = useMemo(
+    () => chunk(sortedList, 10),
```
sternetj, line 52: Does 10 recipients constitute an entire screen? What happens if the device is large, will all the users still eventually get loaded?

src/screens/MessageScreen.tsx
```diff
@@ -1,37 +1,112 @@
...
       title: t('messages-title', 'My Messages'),
     });
   }, [navigation]);
 
+  const { styles } = useStyles(defaultStyles);
   const { User } = useIcons();
+  const [scrollIndex, setScrollIndex] = useState<number>(0);
+
+  // Get userIds that have sent new messages
+  const { unreadMessagesUserIds: unreadUserIds } = useUnreadMessages();
```
sternetj, line 37: It might be nice to split out some of this logic into it's own hook to separate it from the component/layout logic.

sternetj (dismissed):

src/screens/MessageScreen.tsx
bruce-glazier: 👍 Yeah, totally. I'll follow up with a refactor PR in the near future.

src/screens/MessageScreen.tsx
bruce-glazier: Scrolling down will load the next chunk of users but it is possible that on a very tall device that 10 items may not fill the entire height of the screen BUT unread messages will always be added to the top regardless. It could make for a strange UX is it does not fill the entire screen because they will have to scroll down to load the next set of 10 users. I will consider how to make this more dynamic this in a follow up PR.

src/hooks/useNotifications.tsx
bruce-glazier: Good question, these might be loaded earliest to latest which would result in the newest unread message at the bottom of the list. I will follow up on this in another PR.

src/hooks/useNotifications.tsx
bruce-glazier: This is a mistake, I will update it.

src/common/LoggedInProviders.tsx
```diff
@@ -1,13 +1,19 @@
 import React from 'react';
 
 import { ActiveAccountContextProvider } from '../hooks/useActiveAccount';
+import { NotificationsManagerProvider } from '../hooks/useNotificationManager';
+import { UnreadMessagesContextProvider } from '../hooks';
```
markdlv, line 5: Let's change this to import from the most-specific file possible, to avoid a bunch of potentially-undesired imports for SDK consumers.

src/common/LoggedInProviders.tsx
markdlv: Yep, agreed

src/components/MessagesTile/MessagesTile.tsx
```diff
@@ -1,10 +1,13 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 import { Tile } from '../tiles/Tile';
 import { HomeStackScreenProps } from '../../navigators';
-import { useUser, useHasNewMessagesFromUsers } from '../../hooks';
+import { useUser } from '../../hooks';
```
markdlv, line 4: Same here with the import- we want want to be careful not to import from files like `/hooks` because that includes a bunch of files that aren't relevant to this component.

src/hooks/usePushNotifications.tsx
markdlv: ☝️ we really need _always_ do this.  If you see other imports from an index file like `/src/common`, it needs to be updated semi-soon so we're not over-importing what's needed for each file/module.

src/common/LoggedInProviders.tsx
bruce-glazier: Updated - and grabbed a bunch more in the other files I touched too.

src/hooks/useNotifications.tsx
bruce-glazier: Updated - fairly certain what I have now is correct - let me know if you disagree

src/hooks/useNotifications.tsx
bruce-glazier: Done

src/hooks/useNotificationManager.tsx
bruce-glazier: Done

src/hooks/usePushNotifications.tsx
bruce-glazier: I will follow up with another PR to fix these across the entire project.

src/screens/MessageScreen.tsx
sternetj: Sounds good. Specifically for the tall device scenario I would be worried that there might not be enough pages to fully load all the users.

src/hooks/useNotifications.tsx
```diff
@@ -82,24 +98,34 @@ const notificationQuery = gql`
...
 
-const supportedNotificationTypes = [
-  'PostReplyNotification',
-  'CircleAdminPostNotification',
-  'SurveyAssignedNotification',
-];
+function isSupportedNotification(
+  __typename: string,
+): __typename is SupportedNotificationTypes {
+  return (__typename as SupportedNotificationTypes) !== undefined;
```
sternetj, line 119: I don't think this is doing what you want it to do. I think you will need either an enum, an array of strings, or an exhaustive switch to check if it is an expected type. It's probably okay for the moment though since only supported types are being queried.

sternetj (dismissed):

src/hooks/useUnreadMessages.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { useNotificationManager } from './useNotificationManager';
+
+type UnreadMessageContextProps = {
+  unreadMessagesUserIds?: string[];
+  markMessageRead?: (userId: string) => void;
+};
+
+const UnreadMessagesContext = createContext<UnreadMessageContextProps>({});
+
+export const UnreadMessagesContextProvider = ({
```
namick, line 19: This hook is crashing for me.  I'm looking into it...

Screenshot 2023-08-30 at 2 21 01 PM

namick (changes requested):

src/hooks/useUnreadMessages.tsx
bruce-glazier: Interesting, I didn't hit this. I wonder if something went wrong with the rebase.

src/hooks/useNotifications.tsx
bruce-glazier: I'll write a unit test for this and update accordingly. I believed this to be correct based on [this documentation](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) but perhaps I'm misunderstanding how this works.

bruce-glazier: You are correct, I misunderstood what that could do. Reverted just to use an array of supported type name strings.

src/hooks/useUnreadMessages.tsx
namick: That is likely.  I put up [this PR](https://github.com/lifeomic/react-native-sdk/pull/350) that fixes it

src/screens/MessageScreen.tsx
sternetj: I am thinking of the scenario where they have a tall device and only 11 people to message. All 11 people are visible when the app loads. There is no way to scroll to the next page so number 11 never gets loaded.
I wonder if React Native has an interaction observer similar to the web.

bruce-glazier: 💡 Okay, I follow you now.

src/hooks/useUnreadMessages.tsx
bruce-glazier: Merged! Thank you.

namick (approved):
🎉

PR #357, opened 2023-09-07
Improve loading consistency

bruce-glazier:
## About

The example app has been exhibiting some inefficient behavior which is addressed by this PR:
- Race condition in async-storage reads; with no built-in wait mechanism for useAsyncStorage reads would sometimes occur with the key containing `undefined`, .e.g, a key of `${user?.id}-myData` would result in `undefined-myData` and thus fail to fetch the stored data at times.
- API calls occurring before prerequisite auth/account headers ready. Sometimes this is the result of calling refetch() which bypasses the enabled conditions, other times I'm speculating that the async-storage bugs mentioned above can lead to this.
- Duplicate loading indicators on the home page could lead to excessive loading indicators; these required conditions were already being handled by the RootStack Provider.
- Multiple calls to app-config

Recommend reviewing commit-by-commit.

## Changes

  - useAsyncStorage no longer wraps async-storage with react-query and instead uses async-storage directly which improves overall performance when reading from storage. 
  - Providers that require authentication have been removed from RootProviders and placed into LoggedInProviders. This prevents any Providers that require authentication from firing too early.
 - Extracted project/account/consent loading logic and screens from RootStack to a new stack `LoggedInStack` which ensures that the required Providers for the project/account/consent hooks have already initialized. 
 - Converted useAppConfig to a Provider which only fetches the app-config once depending on if the appConfig state value has been set or not.


src/screens/HomeScreen.tsx
```diff
@@ -50,33 +43,6 @@ export const HomeScreen = (navProps: HomeStackScreenProps<'Home'>) => {
     }
   }, [navigation, refreshing, refresh, headerRefreshEnabled]);
 
-  if (loadingAccount) {
```
bruce-glazier, line 53: Redundant - if we are at this screen then we've already satisfied these checks.

src/hooks/useHttpClient.tsx
```diff
@@ -12,12 +14,18 @@ export const defaultAxiosInstance = axios.create({
   headers: defaultHeaders,
 });
 
+export const defaultAPIClientInstance = new APIClient<RestAPIEndpoints>(
+  defaultAxiosInstance,
+);
+
 export interface HttpClient {
   httpClient: AxiosInstance;
+  apiClient: APIClient<RestAPIEndpoints>;
```
bruce-glazier, line 23: I think consolidating these two clients should happen but I figured I better wait to do that in a different PR.

src/hooks/useAsyncStorage.tsx
```diff
@@ -0,0 +1,27 @@
...
+import AsyncStorage from '@react-native-async-storage/async-storage';
+
+export function useAsyncStorage(key: string, enabled: boolean = true) {
+  const [value, setValue] = useState<string | null>(null);
+
+  useEffect(() => {
+    (async () => {
+      if (enabled) {
+        const storedValue = await AsyncStorage.getItem(key);
+        setValue(storedValue);
```
bruce-glazier, line 11: We used to have quite a bit of flow control in the app that relied on react-query's isLoading/isFetched conditions here. Instead we can just rely on the state update to trigger a re-render.

PR #358, opened 2023-09-08
fix: Separate LoggedIn Providers from Root

bruce-glazier:
## About

The example app has been exhibiting some inefficient behavior which is addressed by this PR:
- Race condition in async-storage reads; with no built-in wait mechanism for useAsyncStorage reads would sometimes occur with the key containing `undefined`, .e.g, a key of `${user?.id}-myData` would result in `undefined-myData` and thus fail to fetch the stored data at times.
- API calls occurring before prerequisite auth/account headers ready. Sometimes this is the result of calling refetch() which bypasses the enabled conditions, other times I'm speculating that the async-storage bugs mentioned above can lead to this.
- Duplicate loading indicators on the home page could lead to excessive loading indicators; these required conditions were already being handled by the RootStack Provider.
- Multiple calls to app-config

## Changes

  - useAsyncStorage no longer wraps async-storage with react-query and instead uses async-storage directly which improves overall performance when reading from storage. 
  - Providers that require authentication have been removed from RootProviders and placed into LoggedInProviders. This prevents any Providers that require authentication from firing too early.
 - Extracted project/account/consent loading logic and screens from RootStack to a new stack `LoggedInStack` which ensures that the required Providers for the project/account/consent hooks have already initialized. 
 - Converted useAppConfig to a Provider which only fetches the app-config once depending on if the appConfig state value has been set or not.

### DIFF
--- a/src/navigators/LoggedInStack.tsx
+++ b/src/navigators/LoggedInStack.tsx
@@ -1,0 +1,129 @@
+import React from 'react';
+import { TabNavigator } from './TabNavigator';
+import { ConsentScreen } from '../screens/ConsentScreen';
+import { CircleThreadScreen } from '../screens/CircleThreadScreen';
+import { InviteRequiredScreen } from '../screens/InviteRequiredScreen';
+import { OnboardingCourseScreen } from '../screens/OnboardingCourseScreen';
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
+import { LoggedInRootParamList } from './types';
+import { useConsent } from '../hooks/useConsent';
+import { useActiveAccount } from '../hooks/useActiveAccount';
+import { useActiveProject } from '../hooks/useActiveProject';
+import { useOnboardingCourse } from '../hooks/useOnboardingCourse';
+import { ActivityIndicatorView } from '../components/ActivityIndicatorView';
+import { t } from 'i18next';
+import { usePendingInvite } from '../hooks/usePendingInvite';
+import { useSetUserProfileEffect } from '../hooks/useSetUserProfileEffect';
+
+export function LoggedInStack() {
+  const Stack = createNativeStackNavigator<LoggedInRootParamList>();
+  const { inviteParams } = usePendingInvite();
+  useSetUserProfileEffect();
+  const {
+    account,
+    isLoading: isLoadingAccount,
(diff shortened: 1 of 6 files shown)


src/navigators/LoggedInStack.tsx
```diff
@@ -0,0 +1,129 @@
...
+    );
+  }
+
+  if (inviteParams?.inviteId) {
+    return (
+      <ActivityIndicatorView
+        message={t('root-stack-accepting-invitation', 'Accepting invitation')}
+      />
+    );
+  }
```
sternetj, line 94: Maybe not in this PR but I think it would be good to just have one `ActivityIndicatorView` and change the message depending on the state. By returning different components we will see the indicator restart instead of just spinning smoothly.

src/screens/HomeScreen.tsx
```diff
@@ -1,24 +1,17 @@
...
 import { HomeStackScreenProps } from '../navigators/types';
-import { useJoinCircles } from '../hooks';
 import { refreshNotifier } from '../common/RefreshNotifier';
 import { HeaderLeftRefreshButton } from '../components/HeaderLeftRefreshButton';
 
 export const HomeScreen = (navProps: HomeStackScreenProps<'Home'>) => {
-  const { isLoading: loadingAccount } = useActiveAccount();
-  const { isInitialLoading: loadingAppConfig, data: appConfig } =
-    useAppConfig();
-  const { isInitialLoading: loadingJoinCircles } = useJoinCircles();
```
sternetj, line 17: This appears to be gone now. Need to make sure we add this back so users autojoin circles

sternetj (approved):

src/screens/HomeScreen.tsx
bruce-glazier: Nice catch!!

bruce-glazier: I will add this back on the 3rd PR so I don't have to rebase each one. I plan on merging these one after another anyway.

src/navigators/LoggedInStack.tsx
bruce-glazier: That's an excellent suggestion and would clean up the code too. I'll grab that in a future PR.

namick (approved):

PR #360, opened 2023-09-08
Improve Loading #3: Drop react-query from async-storage

bruce-glazier:
## Changes

  - Stops using react-query for async-storage to make usage simpler and address some caching/loading issues
  
**Note:** For a better diff of useAsyncStorage review the second commit

For a full description and reasoning behind this: https://github.com/lifeomic/react-native-sdk/pull/358


example/storybook/helpers/oauthConfig.ts
```diff
@@ -14,6 +14,7 @@ export const authConfig: AuthConfiguration = {
   additionalParameters: {
     prompt: 'consent',
   },
+  iosPrefersEphemeralSession: true,
```
bruce-glazier, line 17: Unrelated change to suppress warning

src/hooks/useOnboardingCourse.tsx
```diff
@@ -36,17 +36,18 @@ export const OnboardingCourseContextProvider = ({
...
+    !!activeProject?.id,
   );
   const [didLaunchCourse, setDidLaunchCourse] = useState<boolean | undefined>(
     undefined,
   );
 
-  const isLoading = isAppConfigLoading || storedDidLaunchResult.isLoading;
-  const isFetched = isAppConfigFetched && storedDidLaunchResult.isFetched;
+  const isLoading = isAppConfigLoading;
+  const isFetched = isAppConfigFetched;
```
sternetj, line 46: Should we check to make sure `didLaunchCourse` should be populated here? In theory it seems like the async storage being async could make this hook return `isLoading: false, isFetched: true` before the value is actually loaded.

sternetj (approved):

src/hooks/useOnboardingCourse.tsx
bruce-glazier: Coming up in the next PR.

sternetj (approved):

PR #362, opened 2023-09-11
Improve invite accept handling

bruce-glazier:
## Changes

  - The default "Login" button now shows the text "Accept Invite" when the app is opened with an invite
  - When an invite is accepted and the user is already signed in and waiting at the Invite Required screen they are now correctly navigated to the next screen (Consent or Home).

### DIFF
--- a/src/hooks/useMe.ts
+++ b/src/hooks/useMe.ts
@@ -1,7 +1,9 @@
+import { useEffect } from 'react';
 import { useQuery } from '@tanstack/react-query';
 import { useActiveAccount } from './useActiveAccount';
 import { useHttpClient } from './useHttpClient';
 import { Patient } from 'fhir/r3';
+import { inviteNotifier } from '../components/Invitations/InviteNotifier';
 
 export interface Subject {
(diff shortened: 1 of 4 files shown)


sternetj (approved):

src/hooks/useMe.ts
```diff
@@ -43,4 +45,16 @@ export function useMe() {
       enabled: !!accountHeaders,
     },
   );
+
+  useEffect(() => {
+    const listener = async () => {
+      await useMeQuery.refetch();
```
sternetj, line 51: We might want to refetch the app config if we aren't already doing it. If the user is different but the project is the same, they might be in different cohorts and need to get the correct app config version

bruce-glazier: Good point, I will look into that.

bruce-glazier: With my recent refactor all components part of the LoggedInStack are unmounted on logout. This means app-config will be refetched every time there is a fresh login.

PR #363, opened 2023-09-13
Refresh session info with inviteId

bruce-glazier:
## Changes

  - Fixes invite flow when the user does not yet have an account
  - Add script to watch for changes in SDK and build/copy to target project for easier development

### DIFF
--- a/bin/scripts/syncWithProject.js
+++ b/bin/scripts/syncWithProject.js
@@ -1,0 +1,25 @@
+const { TscWatchClient } = require('tsc-watch/client');
+const { execSync } = require('child_process');
+const project = process.argv[2];
+
+const watch = new TscWatchClient();
+
+if (typeof project !== 'string') {
+  throw new Error('Expect string for argument project');
+}
+
+const syncChanges = () => {
+  console.log('Copying changes to', project);
+  execSync('yarn copyAdditionalFiles && yarn buildDeclaredStyles', {
+    stdio: 'inherit',
+  });
+  execSync(
+    `cp -r ./dist/ ../${project}/node_modules/@lifeomic/react-native-sdk`,
+    {
+      stdio: 'inherit',
+    },
+  );
+  console.log('Copy complete');
+};
+watch.start('--project', 'tsconfig.build.json', '--outDir', 'dist');
+watch.on('success', syncChanges);
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "build": "rm -rf dist && tsc -p tsconfig.build.json --outDir dist && yarn copyAdditionalFiles && yarn buildDeclaredStyles",
     "precopyToProject": "yarn build",
     "copyToProject": "sh -c 'cp -r ./dist/ ../${0}/node_modules/@lifeomic/react-native-sdk'",
+    "keepProjectSynced": "node ./bin/scripts/syncWithProject.js",
     "copyToStarter": "yarn copyToProject react-native-starter",
     "copyAdditionalFiles": "./bin/scripts/copyAdditionalFiles",
     "buildDeclaredStyles": "node ./bin/scripts/buildDeclaredStyles.js"
@@ -64,6 +65,7 @@
     "metro-react-native-babel-preset": "^0.74.1",
     "msw": "^1.2.3",
     "nock": "^13.3.0",
+    "node-watch": "^0.7.4",
     "prettier": "^2.8.3",
     "query-string": "^8.1.0",
     "react": "^18.2.0",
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -64,6 +65,7 @@
     "metro-react-native-babel-preset": "^0.74.1",
     "msw": "^1.2.3",
     "nock": "^13.3.0",
+    "node-watch": "^0.7.4",
```
bruce-glazier, line 68: Todo: Remove

bruce-glazier: Done

bin/scripts/syncWithProject.js
```diff
@@ -0,0 +1,25 @@
...
+  });
+  execSync(
+    `cp -r ./dist/ ../${project}/node_modules/@lifeomic/react-native-sdk`,
+    {
+      stdio: 'inherit',
+    },
+  );
+  console.log('Copy complete');
+};
+watch.start('--project', 'tsconfig.build.json', '--outDir', 'dist');
```
sternetj, line 24: Nice so does this build it and then sync it into another projects node_modules or do I need to run the build separately?

src/hooks/useMe.ts
```diff
@@ -21,11 +20,12 @@ interface MeResponse {
 }
 
 export function useMe() {
+  const { authResult } = useAuth();
   const { accountHeaders } = useActiveAccount();
   const { httpClient } = useHttpClient();
 
   const useMeQuery = useQuery(
-    ['fhir/dstu3/$me'],
+    ['fhir/dstu3/$me', authResult?.accessToken],
```
sternetj, line 28: This doesn't seem quite right. This would fire even when the access token is refreshed.

sternetj: Maybe we should consider adding a sessionId to `useAuth` to handle this for us. It could just be a unique string (timestamp?) that changes whenever a user logs in for the first time after being logged out.

bin/scripts/syncWithProject.js
```diff
@@ -0,0 +1,25 @@
...
+  throw new Error('Expect string for argument project');
+}
+
+const syncChanges = () => {
+  console.log('Copying changes to', project);
+  execSync('yarn copyAdditionalFiles && yarn buildDeclaredStyles', {
+    stdio: 'inherit',
+  });
+  execSync(
+    `cp -r ./dist/ ../${project}/node_modules/@lifeomic/react-native-sdk`,
```
markdlv, line 17: Would be sweet to reuse `copyToProject` here if easy enough

bruce-glazier: `copyToProject` is setup to clear the `dist` folder and run the build command which is handled separately by `tsc-watch`.

bin/scripts/syncWithProject.js
bruce-glazier: This handles building and copying so no other commands are needed.

src/hooks/useActiveAccount.tsx
```diff
@@ -66,6 +66,12 @@ export const ActiveAccountContextProvider = ({
     {},
     { select: (data) => data.accounts },
   );
+  console.log(
+    'accountResults',
+    accountsResult.isInitialLoading,
+    accountsResult.isLoading,
+    accountsResult.data,
+  );
```
sternetj, line 74: Remove

src/hooks/useMe.ts
bruce-glazier: Updated to bust cache when the inviteId changes which should work nicely.

sternetj (approved):

src/hooks/useAuth.tsx
```diff
@@ -157,10 +158,19 @@ export const AuthContextProvider = ({
...
-  const refreshForAuthFailure = useCallback(async () => {
-    // TODO: Add throttling, etc.
-    return refreshIfNeeded(true);
-  }, [refreshIfNeeded]);
+  const refreshForAuthFailure = useCallback(
+    async (error: AxiosError) => {
+      const refreshTokenEndpointBlacklist = ['device-endpoints'];
+      if (
+        refreshTokenEndpointBlacklist.every(
+          (endpoint) => !error?.config?.url?.includes(endpoint),
```
bruce-glazier, line 166: Device endpoints will fail if we attempt to re-register the same device with different attributes and cause an unnecessary token refresh.

sternetj: What are the different attributes? I would think we would want to try to reregister for a different account so we can get notifications for all user accounts.

sternetj (approved):

src/hooks/useAuth.tsx
bruce-glazier: Responded offline, all set

PR #366, opened 2023-09-15
onAppSessionStart hook

aymanenadi:
This PR adds an `onAppSessionStart` hook that is triggered when a user signs in to the app. It can be used for analytics, check subscription status and navigate to a different page if a subscription is inactive, etc.

https://github.com/lifeomic/react-native-sdk/assets/19395435/3ab737f7-1a14-4f2d-9b45-a0942f48f112

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,39 @@ export default function App() {
 }
 ```
 
+### onAppSessionStart
+
+`onAppSessionStart` is a function that is invoked when a new logged-in app
+session start (user sign-in/sign-up, app refresh, etc.). It can be used for
+analytics purposes, checking subscription status, etc.
+
+````typescript
+```typescript
+import React, { FC } from 'react';
+import { authConfig } from './authConfig';
+import { RootProviders, RootStack } from '@lifeomic/react-native-sdk';
+
+export default function App() {
+  const onAppSessionStart = async (resumeAppSession: () => void) => {
+   // track app session start metric
+   await trackAppSessionStart();
+   resumeAppSession();
+  };
+
+  return (
+    <DeveloperConfigProvider
+      developerConfig={{
+        onAppSessionStart,
+      }}
+    >
+      <RootProviders authConfig={authConfig}>
+        <RootStack />
+      </RootProviders>
+    </DeveloperConfigProvider>
+  );
+}
+````
+
 ### Customizing the Default Login Screen with a Native View
 
 The default login screen adds a long button to a native view with the name
@@ -497,4 +530,3 @@ better solution.
 
 For specific guidance on upgrading between versions of this SDK, see
 [the upgrade guide](./UPGRADING.md).
-
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ export default function App() {
 ### onAppSessionStart
 
 `onAppSessionStart` is a function that is invoked when a new logged-in app
-session start (user sign-in/sign-up, app refresh, etc.). It can be used for
+session starts (user sign-in/sign-up, app refresh, etc.). It can be used for
 analytics purposes, checking subscription status, etc.
 
 ````typescript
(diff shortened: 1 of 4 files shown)


src/navigators/LoggedInStack.tsx
```diff
@@ -71,6 +80,11 @@ export function LoggedInStack() {
       return t('root-stack-accepting-invitation', 'Accepting invitation');
     } else if (loadingJoinCircles) {
       return t('root-stack-waiting-for-circle-join', 'Joining Circles');
+    } else if (shouldWaitForOnUserSignIn) {
+      return t(
+        'root-stack-waiting-for-on-user-signin',
+        'Processing user sign-in',
```
aymanenadi, line 86: For v2, I was thinking we can expose a variable via `developerConfig` to set this loading message. We can also update the `onUserSignIn` hook to dynamically set the loading message.

src/navigators/LoggedInStack.tsx
```diff
@@ -79,9 +93,41 @@ export function LoggedInStack() {
...
+        shouldWaitForOnUserSignIn &&
+        hasAccountAndProject &&
+        hasUserAndPatient
+      ) {
+        await onUserSignIn?.({
+          userId: userData.id,
+          account: account.id,
+          email: userData.profile.email,
+          patientId: activeSubjectId,
+          project: activeProject.id,
```
markdlv, line 113: I dig this as an initial, hopefully final, list of data to pass into this func.  Just to say it out loud, the implementor of `onUserSignIn` can also use hooks like `useUser` - so I think it'll be wise of us to not add more data (esp. involving network requests) to these props.  With this in mind, we could actually remove the `userData` additions in this file/PR if we wanted to be as lean as possible.

markdlv (dismissed):
Made some nit-picky comments offline in slack, but big picture I think this looks great and will get us what we need.  You might consider adding an example usage of this hook in this repo's example app, in this PR or the next, to make absolute sure we're happy with the interface/DX of implementing this func.  Obviously want to avoid churn / breaking changes if possible.

README.md
```diff
@@ -219,6 +219,39 @@ export default function App() {
 }
 ```
 
+### onAppSessionStart
+
+`onAppSessionStart` is a function that is invoked when a new logged-in app
+session start (user sign-in/sign-up, app refresh, etc.). It can be used for
```
markdlv, line 225: Nit- `start[s]`

src/common/DeveloperConfig.ts
```diff
@@ -105,6 +107,7 @@ export type DeveloperConfig = {
   ontology?: {
     educationContentOverrides?: Record<string, EducationContent>;
   };
+  onAppSessionStart?: (resumeAppSession: () => void) => Promise<void>;
```
markdlv, line 110: Nit- it'd be great to make the first arg to this method an object/type that we can expand over time.  I anticipate sibling props like `setLoadingIndicatorTitle` etc.

src/navigators/LoggedInStack.tsx
```diff
@@ -27,6 +28,7 @@ export function LoggedInStack() {
   } = useActiveAccount();
   const {
     activeProject,
+    activeSubjectId,
```
markdlv, line 31: Last one- I don't think we're using this in a meaningful way and this can be removed, right?

src/navigators/LoggedInStack.tsx
```diff
@@ -71,6 +81,11 @@ export function LoggedInStack() {
       return t('root-stack-accepting-invitation', 'Accepting invitation');
     } else if (loadingJoinCircles) {
       return t('root-stack-waiting-for-circle-join', 'Joining Circles');
+    } else if (shouldWaitForOnAppStart) {
+      return t(
+        'root-stack-waiting-for-on-app-session-start',
+        'Processing app session start',
+      );
```
markdlv, line 88: Not for this PR, and possibly not ever, but something to think about: what should SDK consumers expect if their func errors out and never calls `resumeAppSession`?  More than the decision, I think something that's key is that we document this so devs can write defensive code accordingly.  My gut is to be somewhat like a Promise ctor, and document the danger of an `onAppSessionStart` that never calls `resumeAppSession` -> it will _never_ progress.

markdlv (approved):

src/navigators/LoggedInStack.tsx
aymanenadi: Making calling `resumeAppSession` required to resume the app progression is a drawback of the callback approach. I'll update the doc in a follow-up PR.

PR #374, opened 2023-09-21
Add session cache

bruce-glazier:
## Changes

```
 * Uses stored session context to populate infrequently changing values
 * such as user, account, subjects, projects, and appConfig to improve app loading performance.
 * A new session is queried for when:
 * - No existing session is found in storage
 * - After a session is greater than 7 days old
 * - An new project invite is accepted
 * - On logout
 * - The Clear Session button on the settings screen is pressed (__DEV__ only)
 */
```

**Note**: Review commit by commit; most of the file changes are import/reference updates

### TODO:
- [ ] Manual testing (invite acceptance/invite flow, appConfig notifier, switch users, switch accounts)
- [ ] Add unit tests
- [ ] Clear session on logout


src/hooks/useAsyncStorage.tsx
```diff
@@ -23,6 +23,29 @@ export function useAsyncStorage(key: string, enabled: boolean = true) {
     [key, value, enabled],
   );
 
+  const mergeItem = useCallback(
```
bruce-glazier, line 26: Added this but it's not actually used anywhere in this PR but I know of at least a couple cases that could benefit from this down the line. Needs to be tested.

src/hooks/useActiveProject.tsx
```diff
@@ -6,164 +6,87 @@ import React, {
...
 
   useEffect(() => {
-    if (hookReturnValue?.activeProject?.id) {
-      setStoredProjectId(hookReturnValue?.activeProject?.id);
-    }
-  }, [hookReturnValue?.activeProject?.id, setStoredProjectId]);
-
-  // Clear selected project when
-  // we've detected that the userId has changed
-  useEffect(() => {
```
sternetj, line 129: I assume this is handled elsewhere now.

src/hooks/useAsyncStorage.tsx
```diff
@@ -23,6 +23,29 @@ export function useAsyncStorage(key: string, enabled: boolean = true) {
...
+  );
+
+  const clearItem = useCallback(() => {
+    AsyncStorage.removeItem(key);
+    setValue(null);
+  }, [key]);
+
   const isLoaded = typeof value !== 'undefined'; // Expect string or null when loaded
-  return [value, setItem, isLoaded] as [typeof value, typeof setItem, boolean];
+  return [value, setItem, isLoaded, mergeItem, clearItem] as [
```
sternetj, line 44: It would probably be good to convert this to an object in another PR

src/hooks/useActiveProject.tsx
```diff
@@ -6,164 +6,87 @@ import React, {
...
+    if (!isLoaded || !isStorageLoaded) {
+      return undefined;
     }
 
-    return {};
+    const projectToSelect = selectedId ?? storedProjectIdResult;
+    const subjectAndProject = projectToSelect
+      ? configuration?.subjects.find((s) => s.projectId === projectToSelect)
+      : configuration?.subjects[0];
+    return subjectAndProject ? { activeSubject: subjectAndProject } : undefined;
```
sternetj, line 56: I feel like it might be better to just return `subjectAndProject`. The API looks a little funky right now `useActiveProject` returning a single value called `activeSubject`

src/hooks/index.ts
sternetj, line 1: These changes make me think this should be a breaking change

src/hooks/useOnboardingCourse.tsx
```diff
@@ -24,32 +21,29 @@ export const OnboardingCourseContextProvider = ({
...
     );
+
   const [didLaunchCourse, setDidLaunchCourse] = useState<boolean | undefined>(
     undefined,
   );
 
-  const isLoading = isAppConfigLoading || !isStorageLoaded;
-  const isFetched = isAppConfigFetched && isStorageLoaded;
+  const isOnboardingLoading = isLoading || !isStorageLoaded;
+  const isFetched = isStorageLoaded;
```
sternetj, line 40: Should this be
```suggestion
  const isFetched = !!data && isStorageLoaded;
```

src/hooks/useSession.tsx
```diff
@@ -0,0 +1,211 @@
...
+      email: Session.placeholder,
+    },
+  },
+  configurations: [],
+  accounts: [],
+  sessionTime: new Date().toISOString(),
+};
+
+export const SessionContext = createContext({
+  userConfiguration: configurationPlaceholder,
```
sternetj, line 51: Is the thought that we might add more of these types of objects to the context over time?

src/hooks/useSession.tsx
```diff
@@ -0,0 +1,211 @@
...
+          const projectResponse = await apiClient.request(
+            'GET /v1/projects',
+            {
+              id: s.projectId,
+            },
+            { headers: accountHeader },
+          );
+          const project = projectResponse.data.items[0];
+          const appConfig = await apiClient.request(
+            'GET /v1/life-research/projects/:projectId/app-config',
```
sternetj, line 168: We could fetch these at the same time by assuming the `s.projectId` is valid.

src/hooks/useSession.tsx
```diff
@@ -0,0 +1,211 @@
...
+/**
+ * Uses stored session context to populate infrequently changing values
+ * such as user, account, subjects, projects, and appConfig to improve app loading performance.
+ * A new session is queried for when:
+ * - No existing session is found in storage
+ * - After a session is greater than 7 days old
+ * - An new project invite is accepted
+ * - On logout
+ * - The Clear Session button on the settings screen is pressed (__DEV__ only)
+ */
```
sternetj, line 66: Very cool, something we might want to add in the future would be to fetch the app config for the active project in the background and then add a toast that says `there is an app update, tap to update now` or something like that so they get a fast app load but we can avoid them falling too far behind if their cohorts, tiles, or anything has changed that would cause them to get a different app config.

src/hooks/useAppConfig.tsx
```diff
@@ -1,146 +0,0 @@
...
-        {
-          projectId: activeProject?.id!,
-        },
-        { headers: accountHeaders },
-      )
-      .then((res) => {
-        isLoading.current = false;
-        error.current = undefined;
-        setAppConfig(res.data);
-        appConfigNotifier.emit(res.data);
```
sternetj, line 114: I think this is now missing

src/hooks/useActiveProject.tsx
bruce-glazier: Since this hook no longer has anything to do with fetching subject projects, it no longer needs to keep track of the userId and force a refresh. Instead, when a session ends, either by a user logout or failure to fresh, the cached session will be discarded. This is not implemented in the code yet yet but it is on my TODO list.

src/hooks/useAsyncStorage.tsx
bruce-glazier: I had the same thought, didn't want to bloat this PR but I will do that in a follow up.

src/hooks/useOnboardingCourse.tsx
bruce-glazier: My actual intention was to drop the isFetched flag and just use the isLoading flag instead. I will review and update accordingly.

src/hooks/useSession.tsx
bruce-glazier: That's an excellent idea!

src/hooks/useSession.tsx
bruce-glazier: Great point!

src/hooks/useSession.tsx
bruce-glazier: I've at least envisioned using the result of appConfig to determine if trackers should be fetched and stored in session for the sake of reducing/removing the loading time for the track tile on the Home Screen.

src/hooks/useActiveAccount.tsx
```diff
@@ -61,123 +51,64 @@ export const ActiveAccountContextProvider = ({
...
-  const userId = userData?.id;
+  const { isLoaded, userConfiguration, clearSession } = useSession();
+  const { user, accounts } = userConfiguration;
   const [selectedId, setSelectedId] = useState(accountIdToSelect);
-  const [previousUserId, setPreviousUserId] = useState(userId);
   const [storedAccountIdResult, setStoredAccountId, isStorageLoaded] =
     useAsyncStorage(
-      `${selectedAccountIdKey}:${userId}`,
-      !!selectedAccountIdKey && !!userId,
+      `${selectedAccountIdKey}:${user}`,
```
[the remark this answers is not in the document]
bruce-glazier: Opps, missed that, yes.

bruce-glazier:
This is almost finished but the plan is to back burner this for a little bit and revisit soon. Closing for now.

What is left to complete this:
1. Rebase with latest and address and conflicts
2. All existing unit tests are fixed but new tests are needed for useSession itself
3. More manual testing around the invite flow, etc.
4. Split into more precise commits and PRs.

PR #378, opened 2023-09-25
feat: Add Setting Screen Menu Item Injection

sternetj:
## Changes

  - Adds the ability to modify default menu items and inject new ones via the developer config

## Screenshots

image
image

### DIFF
--- a/src/common/DeveloperConfig.ts
+++ b/src/common/DeveloperConfig.ts
@@ -47,6 +47,8 @@ import { EducationContent } from '../components/TrackTile/services/TrackTileServ
  * must be the `coding.code` value, e.g. '41950-7' for steps.
  *
  * @param onAppSessionStart Allows for providing a callback to be called when a logged in session starts.  This is useful for analytics tracking, subscription status checks, etc.
+ *
+ * @param modifySettingScreenMenuItems Allows modifying and injecting menu items into the settings screen
  */
 
 export interface RouteColor {
@@ -112,6 +114,9 @@ export type DeveloperConfig = {
     educationContentOverrides?: Record<string, EducationContent>;
   };
   onAppSessionStart?: (params: OnAppSessionStartParams) => Promise<void>;
+  modifySettingScreenMenuItems?: (
+    items: { id: string; title: string; action: () => void }[],
+  ) => { id: string; title: string; action: () => void }[];
 };
 
 export type LogoHeaderConfig = { [key in Route]?: LogoHeaderOptions };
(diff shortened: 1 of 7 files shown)


src/common/DeveloperConfig.ts
```diff
@@ -112,6 +114,9 @@ export type DeveloperConfig = {
     educationContentOverrides?: Record<string, EducationContent>;
   };
   onAppSessionStart?: (params: OnAppSessionStartParams) => Promise<void>;
+  modifySettingScreenMenuItems?: (
+    items: { id: string; title: string; action: () => void }[],
+  ) => { id: string; title: string; action: () => void }[];
```
markdlv, line 119: So you get the default SDK menu items passed to you, then return the ones you want shown.  Beautiful.

src/hooks/useHttpClient.tsx
```diff
@@ -48,7 +48,10 @@ export const HttpClientContextProvider = ({
   const { authResult, refreshForAuthFailure } = useAuth();
 
   const axiosInstance = injectedAxiosInstance || defaultAxiosInstance;
-  const apiClient = defaultAPIClientInstance;
+  const apiClient = useMemo(
+    () => new APIClient<RestAPIEndpoints>(axiosInstance),
+    [axiosInstance],
+  );
```
markdlv, line 54: Why is this needed?

sternetj: I wanted to provide an app config and it uses the new one query rest client that didn't work with the DataProviderDecorator prior to this change. I could probably back this out and just use the AppConfigContext.Provider instead.

markdlv (dismissed):
See my note about useHttpClient file changes- seems unrelated

src/hooks/useHttpClient.tsx
markdlv: My gut reaction would be yes, if this is only beneficial for storyboard purposes, it'd be sweet to refactor to where this change isn't needed.  At the same time, I'm not overly concerned about the change, so if it's a can of worms nbd.

sternetj: Not a big deal. I switched it to just providing my own config

src/common/DeveloperConfig.ts
sternetj: Yeah exactly. It also has the added benefit of being able to change the default order in addition to injecting/removing default values.

markdlv (approved):

PR #381, opened 2023-09-26
fix: handle existing appTileSettings without an app tile title override

aymanenadi:


src/screens/AuthedAppTileScreen.tsx
```diff
@@ -20,7 +20,7 @@ export const AuthedAppTileScreen = ({
   const webViewRef = useRef<WebView>(null);
   const { data: appConfig } = useAppConfig();
   const titleOverride =
-    appConfig?.homeTab?.appTileSettings?.appTiles[appTile.id].title;
+    appConfig?.homeTab?.appTileSettings?.appTiles[appTile.id]?.title;
```
mdlavin, line 23: I'm surprised the TypeScript compiler didn't catch this for you. Can the typing be improved?

markdlv (approved):

PR #382, opened 2023-09-26
Fix typing on app tile settings

mdlavin:
This fixes https://github.com/lifeomic/react-native-sdk/pull/381#discussion_r1337183582 so that the compiler will catch the bug even without a testcase


aymanenadi (approved):
Thank you @mdlavin !

namick (approved):
Nice, thanks!

PR #384, opened 2023-09-26
feat: Set a custom home screen

aymanenadi:
This PR adds support for setting a custom home screen.

Screenshot 2023-09-26 at 12 27 16 PM

### DIFF
--- a/src/navigators/HomeStack.tsx
+++ b/src/navigators/HomeStack.tsx
@@ -22,14 +22,15 @@ import { MessageScreen } from '../screens/MessageScreen';
 const Stack = createNativeStackNavigator<HomeStackParamList>();
 
 export function HomeStack() {
-  const { getAdditionalHomeScreens, logoHeaderConfig } = useDeveloperConfig();
+  const { getAdditionalHomeScreens, logoHeaderConfig, CustomHomeScreen } =
+    useDeveloperConfig();
 
   return (
     <Stack.Navigator
       screenOptions={{ header: AppNavHeader }}
       screenListeners={navigationScreenListeners(logoHeaderConfig)}
     >
-      <Stack.Screen name="Home" component={HomeScreen} />
+      <Stack.Screen name="Home" component={CustomHomeScreen || HomeScreen} />
       <Stack.Screen name="Home/AppTile" component={AppTileScreen} />
       <Stack.Screen name="Home/AuthedAppTile" component={AuthedAppTileScreen} />
       <Stack.Screen name="Home/CustomAppTile" component={CustomAppTileScreen} />
--- a/README.md
+++ b/README.md
@@ -164,6 +164,35 @@ export default function App() {
 }
 ```
 
+### Custom Home screen
+
+```typescript
+const CustomHomeScreen = () => {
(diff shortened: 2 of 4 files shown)


src/navigators/HomeStack.tsx
```diff
@@ -22,14 +22,15 @@ import { MessageScreen } from '../screens/MessageScreen';
...
+  const { getAdditionalHomeScreens, logoHeaderConfig, CustomHomeScreen } =
+    useDeveloperConfig();
 
   return (
     <Stack.Navigator
       screenOptions={{ header: AppNavHeader }}
       screenListeners={navigationScreenListeners(logoHeaderConfig)}
     >
-      <Stack.Screen name="Home" component={HomeScreen} />
+      <Stack.Screen name="Home" component={CustomHomeScreen || HomeScreen} />
```
sternetj, line 33: This seems fine but I think we could also use `getAdditionalHomeScreens` to override the screen named `"Home"`

sternetj (approved):

src/navigators/HomeStack.tsx
jkdowdle: what is the main difference between the two?

sternetj: With the proposed solution we are adding a specific hook for overriding the home screen. By using `getAdditionalHomeScreens` we are registering two screens with the name `Home` but react navigation doesn't allow for multiple routes with the same name so it uses the most recently configured screen (LIFO).

aymanenadi: I originally tried using `getAdditionalHomeScreens` to override the screen named `Home`, but it does not like it.
Screenshot 2023-09-26 at 1 54 49 PM

aymanenadi: >so it uses the most recently configured screen (LIFO).
@sternetj Is this from their documentation? I'm asking because that's not the behavior I got. It flat out throws an exception when I provide a screen named "Home" in `getAdditionalHomeScreens`.

sternetj: Huh I must be misremembering our set up. Sorry about that!

markdlv: @aymanenadi My intention was for us to utilize `componentProps.TabBar.tabs` for our current efforts.  After all, we need to customize more than just the home tab.

markdlv: The `tabs` specify the entire stacks to use via the `component` prop, so one should be able to re-create `HomeStack` to their liking and provide it in the right `tab`.  But they could re-use stacks as well, like `NotificationsStack` or `SettingsStack`.

aymanenadi: Right. So the question is whether we want to recreate `HomeStack`, or just use `CustomHomeScreen` which essentially injects the custom home screen in the existing `HomeStack`.

markdlv: Right.  I'd have preferred the former in the case for a number of reasons.  This is already merged, so let's utilize it.  Give a little more time on these SDK PRs next round, though.

aymanenadi: That's all good. I don't need to use this in the FL app.

PR #386, opened 2023-09-27
Hide HealthKit when not on iOS platform

markdlv:


aymanenadi (approved):
Should we do something similar for Android (i.e. hide Google fit on iOS)?

src/screens/WearablesScreen.tsx
```diff
@@ -24,7 +25,13 @@ const WearablesScreen = () => {
   const { onPostToggle, onBackfill } = useWearableLifecycleHooks();
   const { enabledBackfillWearables, backfillEHR } = useWearableBackfill(data);
 
-  const wearables = data?.items || [];
+  const wearables = useMemo(() => {
+    let items = data?.items || [];
+    if (Platform.OS !== 'ios') {
+      items = items.filter((w) => w.ehrType !== EHRType.HealthKit);
```
sternetj, line 31: The iOS plugin does the same thing. Should we rely on that or have it in two places? If an app is built without the plugin, it would certainly be a confusing behavior. https://github.com/lifeomic/react-native-sdk-plugins/blob/master/src/healthKit/HealthKitLifecycleHandler.ts#L50-L59

markdlv:
> Should we do something similar for Android (i.e. hide Google fit on iOS)?

Nope- Google fit can be used on iOS.

src/screens/WearablesScreen.tsx
markdlv: Ah sweet, great catch.  I need to dig in further why this either isn't working OR perhaps it was just a temp issue before the plugin was enabled.  Looking

swain (approved):

src/screens/WearablesScreen.tsx
sternetj: Plugin isn't configured: https://github.com/lifeomic/branded-mobile-apps/blob/master/apps/fountainlifedemo/appConfig.json#L19

You would still see this locally when run without the plugin

markdlv: Again great catch and then offline convo on this.  We decided we want to go ahead and move forward for this and tweak our plugins to not worry about this sanitization.  This will come with the benefit of also not showing Apple Health on Android when running locally pre-LifeOmic-build 👍

PR #394, opened 2023-10-02
Add an Invalidate Today Tasks Hook

sternetj:
## Changes

  - Switches the refetch tasks logic to use the route listeners based on a route param, removes useFocusEffect which caused  duplicate queries on load
  - Adds a specialized hook for invalidating the today tasks cache
  - Adds a delay so today activity interactions can settle on the backend before fetching the latest count (3.5 seconds worked well during my testing)

### DIFF
--- a/src/hooks/todayTile/useTodayTasks.tsx
+++ b/src/hooks/todayTile/useTodayTasks.tsx
@@ -2,15 +2,36 @@ import { useCallback } from 'react';
 import { useGraphQLClient } from '../useGraphQLClient';
 import { gql } from 'graphql-request';
 import { useActiveAccount } from '../useActiveAccount';
-import { useQuery } from '@tanstack/react-query';
+import { useQuery, useQueryClient } from '@tanstack/react-query';
 import { useActiveProject } from '../useActiveProject';
-import { useRestQuery } from '../rest-api';
+import { useRestQuery, useRestCache } from '../rest-api';
 import { ConsentTask, SurveyResponse } from './types';
 
 export type TodayTask = SurveyResponse | ConsentTask;
 export function isConsentTask(value: TodayTask): value is ConsentTask {
   return !!(value as ConsentTask).form;
 }
+
+export const useInvalidateTodayCountCache = () => {
+  const restCache = useRestCache();
+  const queryClient = useQueryClient();
+
+  // We wait 3.5 seconds by default to give the backend a chance to update
+  return useCallback(
+    (delay = 3500) => {
+      setTimeout(() => {
+        console.log('Fetch with delay', delay);
+        restCache.invalidateQueries({
+          'GET /v1/consent/directives/me': 'all',
+          'GET /v1/survey/projects/:projectId/responses': 'all',
+        });
+
+        queryClient.refetchQueries(['getIncompleteActivitiesCount']);
+      }, delay);
+    },
+    [restCache, queryClient],
+  );
+};
 
 const useConsentTasks = () => {
   const { activeProject } = useActiveProject();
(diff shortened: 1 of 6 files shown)


namick (approved):

src/hooks/todayTile/useTodayTasks.tsx
```diff
@@ -2,16 +2,37 @@ import { useCallback } from 'react';
...
 
+export const useInvalidateTodayCountCache = () => {
+  const restCache = useRestCache();
+  const queryClient = useQueryClient();
+
+  // We wait 3.5 seconds by default to give the backend a chance to update
+  return useCallback(
+    (delay = 3500) => {
+      setTimeout(() => {
+        console.log('Fetch with delay', delay);
```
bruce-glazier, line 23: console.log

src/hooks/todayTile/useTodayTasks.tsx
```diff
@@ -2,16 +2,37 @@ import { useCallback } from 'react';
...
 export type TodayTask = SurveyResponse | ConsentTask;
 export function isConsentTask(value: TodayTask): value is ConsentTask {
   return !!(value as ConsentTask).form;
 }
 
+export const useInvalidateTodayCountCache = () => {
+  const restCache = useRestCache();
+  const queryClient = useQueryClient();
+
+  // We wait 3.5 seconds by default to give the backend a chance to update
```
bruce-glazier, line 19: I'd think that there wouldn't be this long of a delay between when we've submitted a survey and  when the data is available on the backend. I also assumed that the user would maybe take a second or two before navigating directly back to the Home Screen. Did you find this to be an issue in your testing?

1) submitting a survey or completing some activity (submit to the backend)
2) closing the activity and navigating away from the today applet and back to the Home Screen

src/hooks/todayTile/useTodayTasks.tsx
```diff
@@ -110,6 +131,8 @@ export const useTodayTasks = () => {
     ...inProgressSurveyTasks,
   ] as TodayTask[];
 
+  console.log(newTasks);
```
bruce-glazier: console.log

src/navigators/HomeStack.tsx
```diff
@@ -32,7 +34,17 @@ export function HomeStack() {
     >
       <Stack.Screen name="Home" component={CustomHomeScreen || HomeScreen} />
       <Stack.Screen name="Home/AppTile" component={AppTileScreen} />
-      <Stack.Screen name="Home/AuthedAppTile" component={AuthedAppTileScreen} />
+      <Stack.Screen
+        name="Home/AuthedAppTile"
+        component={AuthedAppTileScreen}
+        listeners={(props) => ({
+          beforeRemove: () => {
+            if (props.route.params.refreshTodayCountOnRemove) {
```
bruce-glazier, line 42: Nice, like this approach.

bruce-glazier (approved):

src/hooks/todayTile/useTodayTasks.tsx
sternetj: Yeah my testing comprised of completing an activity (user interaction, not a survey) and then navigating back as fast as I could. I saw it reliably work with 3.5 but I agree that is pretty arbitrary and could depend on a variety of factors. Maybe it would be worth assuming a 1-2 second delay on the today tile before navigating back. Open to suggestions.

bruce-glazier: If it works reliably and the task count change feels responsive enough on the home screen then 3.5 seconds is fine with me.

PR #395, opened 2023-10-03
Allow override of activity indicator view

bruce-glazier:


jkdowdle (approved):

swain (dismissed):

sternetj (dismissed):

src/common/DeveloperConfig.ts
```diff
@@ -113,6 +113,7 @@ export type DeveloperConfig = {
     HomeStack: Navigator<ParamList, NativeStackNavigatorProps>,
   ) => JSX.Element[];
   CustomHomeScreen?: () => JSX.Element;
+  CustomActivityIndicatorView?: JSX.Element;
```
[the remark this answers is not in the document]
aecorredor: I also noticed that the custom home screen takes a function that returns the element. Do we want to consolidate patterns or is there a good reason why the custom home screen wants a function?

src/screens/AccountSelectionScreen.tsx
```diff
@@ -58,6 +58,7 @@ const defaultStyles = createStyles('AccountSelection', () => ({
   button: {
     marginBottom: 28,
   },
+  accountText: {},
```
aecorredor, line 61: do we want this empty style?

src/common/DeveloperConfig.ts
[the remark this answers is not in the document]
bruce-glazier: > I wonder if we should have a component here instead so we could pass the timeout message to it.
Excellent thought, I will do that.

> is there a good reason why the custom home screen wants a function?
The custom home screen function is passed into a Stack.Screen which expects a function.

aecorredor (approved):

src/common/DeveloperConfig.ts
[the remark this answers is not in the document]
bruce-glazier: Done

src/components/ActivityIndicatorView.tsx
```diff
@@ -23,23 +23,26 @@ export function ActivityIndicatorView({
...
-        />
-        {timeOutMessage && showMessage() && (
-          <Text variant="labelSmall" style={styles.text}>
-            {timeOutMessage}
-          </Text>
-        )}
-      </View>
-    )
+  const timeOutMessage =
+    !!messageText && !!showMessage() ? (
```
aecorredor, line 27: I think I initially missed this: I saw that the useTimeout hook causes a component to re-render after the timeout. Does that just make this component show the message after X amount of time?

aecorredor (approved):

src/components/ActivityIndicatorView.tsx
sternetj: Yeah it does. We had issues early on where we would just see a loader and have no idea why it was spinning forever. Now a message pops up after some time which aids in debugging.

src/common/DeveloperConfig.ts
```diff
@@ -113,6 +113,9 @@ export type DeveloperConfig = {
     HomeStack: Navigator<ParamList, NativeStackNavigatorProps>,
   ) => JSX.Element[];
   CustomHomeScreen?: () => JSX.Element;
+  CustomActivityIndicatorView?: React.FC<{
+    timeoutMessage: JSX.Element | null;
```
sternetj, line 117: Seems okay but it might be better to just pass the message through as a string and let the custom view decide what to do with it.

sternetj (approved):

src/common/DeveloperConfig.ts
bruce-glazier: I suppose it might be difficult to write your own timeout message component if you don't have access to the message. I'll follow up in a different PR.

PR #398, opened 2023-10-05
Messages - Timestamps, sorting, new styles, preview prefix

bruce-glazier:
## Screenshots

Screenshot 2023-10-04 at 5 19 53 PM

### DIFF
--- a/src/hooks/useMyMessages.tsx
+++ b/src/hooks/useMyMessages.tsx
@@ -10,6 +10,7 @@ import {
   useLookupUsersFirstPost,
 } from './Circles/usePrivatePosts';
 import { UseQueryResult } from '@tanstack/react-query';
+import { differenceInSeconds } from 'date-fns';
 
 export const useMyMessages = (tileId: string) => {
   const [pageIndex, setPageIndex] = useState(0);
@@ -75,10 +76,26 @@ export const useMyMessages = (tileId: string) => {
         displayName: data.user.profile.displayName,
         picture: data.user.profile.picture,
         isUnread: unreadIds?.includes(data.user.userId) ?? false,
-        message: data.privatePosts.edges?.[0]?.node.message,
+        message: data.privatePosts.edges?.[0]?.node.message ?? '',
+        messageTime: data.privatePosts.edges?.[0]?.node.createdAt,
+        messagePrefix:
+          data.privatePosts.edges?.[0]?.node.authorId === userData?.id
+            ? 'You: '
+            : '',
       };
     }),
-  );
+  ).sort((a, b) => {
+    if (!a.messageTime) {
+      return 0;
+    }
+    if (!b.messageTime) {
+      return -1;
+    }
+    return differenceInSeconds(
+      new Date(b.messageTime),
+      new Date(a.messageTime),
+    );
+  });
 
   const isLoading = userQueries.some(
     ({ isInitialLoading }) => isInitialLoading,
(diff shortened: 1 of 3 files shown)


src/screens/MessageScreen.tsx
```diff
@@ -136,10 +153,51 @@ const defaultStyles = createStyles('MessageScreen', (theme) => {
...
+  } else if (differenceHours === 1) {
+    return `${differenceHours}hour`;
+  } else if (differenceHours < 24) {
+    return `${differenceHours}hrs`;
+  } else if (daysDifference === 1) {
+    return `${daysDifference}day`;
+  } else if (daysDifference < 3) {
+    return `${daysDifference}days`;
+  } else {
+    return format(date, 'MMM d');
```
aecorredor, line 201: I think most time libraries support this kind of thing by default. I think for `date-fns` we'd want https://date-fns.org/v2.30.0/docs/formatDistance. We could get the shorter text by using the custom formatting letters.

aecorredor (commented):
Quick question about the styling: will we be able to theme this to the dark app colors easily? Or will we need to add some more theme modifications within these components to support that?

bruce-glazier:
> Quick question about the styling: will we be able to theme this to the darker FL app colors easily? Or will we need to add some more theme modifications within these components to support that?

@aecorredor I believe the styling for this screen is already in the demo, we should be all set there.

aecorredor:
@bruce-glazier so it's already darked theme like we can see here: https://www.figma.com/file/yYSMvoTF16xHYCsS8wHWej/Fountain-Life-Member-App-Internal?type=design&node-id=987-7181&mode=design&t=pc2zwXDQcpUjoHaz-0?

src/screens/MessageScreen.tsx
bruce-glazier: I tried to use formatDistance but it doesn't support the custom letter from what I could tell. I'll definitely revisit this in the future and put something more standard in place but I think it's fine for now.

jkdowdle: what is the custom letter we need? the `min/hrs/hour` etc?

jkdowdle (dismissed):

src/screens/MessageScreen.tsx
jkdowdle: not sure what this suffix does exactly but I saw it quickly

```
formatDistance(subDays(new Date(), 3), new Date(), { addSuffix: true })
```

sternetj (approved):

src/hooks/useMyMessages.tsx
```diff
@@ -75,10 +76,26 @@ export const useMyMessages = (tileId: string) => {
         displayName: data.user.profile.displayName,
         picture: data.user.profile.picture,
         isUnread: unreadIds?.includes(data.user.userId) ?? false,
-        message: data.privatePosts.edges?.[0]?.node.message,
+        message: data.privatePosts.edges?.[0]?.node.message ?? '',
+        messageTime: data.privatePosts.edges?.[0]?.node.createdAt,
+        messagePrefix:
+          data.privatePosts.edges?.[0]?.node.authorId === userData?.id
+            ? 'You: '
```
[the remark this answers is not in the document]
bruce-glazier: 👍 I'll grab this in a follow up PR.

src/screens/MessageScreen.tsx
```diff
@@ -41,23 +49,26 @@ export function MessageScreen({
 
   const renderLeft = useCallback(
     (props: { style: ViewStyle }, user: User) => (
-      <View style={props.style}>
+      <View
+        style={
+          user.isUnread ? [props.style, styles.newMessageIconView] : props.style
```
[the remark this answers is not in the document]
bruce-glazier: I agree. I removed it since the badge wasn't in the latest designs but I think making the badge show by default and allow for hiding it would be preferable. I'll update in the next PR.

aecorredor (approved):

swain (approved):

PR #401, opened 2023-10-05
Code cleanup/PR feedback for Messages

bruce-glazier:
## Changes

This PR is almost only renames and code cleanup with the exception of tweaking the message timestamp abbreviations a bit.

### DIFF
--- a/src/screens/MessageScreen.tsx
+++ b/src/screens/MessageScreen.tsx
@@ -1,6 +1,6 @@
 import React, { useCallback, useLayoutEffect } from 'react';
 import { useStyles } from '../hooks/useStyles';
-import { Divider, List, Button, Text } from 'react-native-paper';
+import { Divider, List, Button, Text, Badge } from 'react-native-paper';
 import { TouchableOpacity, ScrollView, View, ViewStyle } from 'react-native';
 import { createStyles } from '../components/BrandConfigProvider';
 import { GiftedAvatar, User as GiftedUser } from 'react-native-gifted-chat';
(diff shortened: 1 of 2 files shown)


src/screens/MessageScreen.tsx
```diff
@@ -9,13 +9,8 @@ import { t } from 'i18next';
...
 import { useMyMessages } from '../hooks/useMyMessages';
-import {
-  format,
-  differenceInMinutes,
-  differenceInHours,
-  differenceInDays,
-  differenceInSeconds,
-} from 'date-fns';
+import { format, differenceInDays } from 'date-fns';
+import dayjs from 'dayjs';
```
aecorredor, line 13: My two cents here are that it'd be great if we could consolidate on a single date library for the SDK. I think that means that keeping your previous code that manually calculated the "from now" format while only keeping `date-fns` looks better than having both `date-fns` and `dayjs` at the same time just for `dayjs`'s format helper.

bruce-glazier: I wonder if we should just swap date-fns with dayjs. I think date-fns doesn't have nearly as much utility and we are not too deep with date-fns yet.

bruce-glazier: I think for now I'll backpedal on the dayjs decision and use the function I initially created.

swain: +1, sounds good. If it does come around, I'm personally in favor of `dayjs`. I find it's much more ergonomic than `date-fns`.

bruce-glazier: Done, switched back to a function.

src/screens/MessageScreen.tsx
```diff
@@ -150,6 +179,16 @@ const defaultStyles = createStyles('MessageScreen', (theme) => {
     },
     badgeView: {
       alignSelf: 'center',
+      paddingRight: 0,
+      marginRight: -10,
+    },
+    badgeColor: {
+      enabled: {
+        backgroundColor: theme.colors.primary,
```
bruce-glazier, line 187: A consumer of this can set this color to transparent if they don't want the badge.

sternetj: Or they can do `badgeView.display: 'none'` That's my go to for hiding content.

sternetj (approved):

PR #407, opened 2023-10-11
Move LoggedInProviders back to RootProviders

bruce-glazier:
## Changes

Injecting custom providers used to be as simple as putting them underneath RootProviders but many hooks are no longer available in RootProviders. This was caused by a refactor to separate hooks and screens that require authentication from unauthenticated hooks and screens. Splitting them out solved a number of issues including premature API requests but resulted in loosing access to the authenticated hooks within injected custom providers.

The issues that the original refactor hoped to solve for were achieved by splitting out the LoggedInStack from the RootStack and moving the providers to the LoggedInStack and out of the RootProviders was most likely unnecessary. 

Although the providers contained within LoggedInProviders could have been moved directly back into RootProviders I've opted to keep them in their own file. I believe there is still a benefit to having them organized this way so a developer can easily discern providers that require auth or not.

### DIFF
--- a/src/common/RootProviders.tsx
+++ b/src/common/RootProviders.tsx
@@ -12,6 +12,7 @@ import { NoInternetToastProvider } from '../hooks/NoInternetToastProvider';
 import { ActionSheetProvider } from '@expo/react-native-action-sheet';
 import { SafeAreaProvider } from 'react-native-safe-area-context';
 import { ThemedNavigationContainer } from './ThemedNavigationContainer';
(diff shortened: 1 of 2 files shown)


markdlv (approved):
Would be ideal to run through our typical auth/onboarding/invites tests locally/manually before merging.

sternetj (approved):

bruce-glazier:
Tested onboarding flow E2E - merging

PR #414, opened 2023-10-24
feat: Add the navigation linking via the developer config

aymanenadi:
This PR adds the ability to set `linking` on the `NavigationContainer` to handle deep-linking to certain screens in the app.

[https://reactnavigation.org/docs/configuring-links](https://reactnavigation.org/docs/configuring-links)


sternetj (approved):
Do we intend to create any linking configs for existing screens such as surveys?

markdlv (approved):
Nice!

aymanenadi:
> Do we intend to create any linking configs for existing screens such as surveys?

I don't know if that's an intention, but that's a possibility.

PR #415, opened 2023-10-24
feat: Add Configurable App Tabs

sternetj:
## Changes

  - Adds the ability to configure app tabs in the app config
  - Adds support for home, notifications, settings, app tile, authed app tile, and custom

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/2295908/28629013-1580-4d2a-9520-c52182611723


bruce-glazier (dismissed):

src/screens/AppTileScreen.tsx
```diff
@@ -1,26 +1,62 @@
...
+        ? () => (
+            <Appbar.Action
+              icon={ChevronLeft}
+              color={styles.backActionIcon?.color}
+              onPress={() => webViewRef.current?.goBack()}
+              style={styles.backAction}
+              disabled={!canGoBack}
+            />
+          )
+        : undefined,
```
markdlv, line 38: tabMode is [true when we're in the context of an AppTileStack-configured-tab](https://github.com/lifeomic/react-native-sdk/pull/415/files#diff-9ad2569bd5d7d98971cf7d8f384da603a8eb17bd1e9e08fe8c0f2d9bd8e1daf4R35), right?  This intuitively reads either backwards or off to me.  If inside an app tile tab, I'd expect the back button _not_ to show (at least initially).  This shows up in the PR's demo too - while the back button is disabled initially, I'd think we'd want it to not show up at all (initially).  When inside a home-screen-tile-tapped appTile (i.e. tabMode=false, right?), I'd expect we'd want the backButton shown & enabled immediately.  So that's the "backwards" thought.  The "off" feeling is that it seems like in both cases, when the appTile itself navigates to something else, we want the Appbar.Action - but this logic appears to remove that _altogether_ for an appTile that was home-screen-tile-tapped.  Am I misreading this or am I onto something?

markdlv (dismissed):
Please see my questions before merging, but approving in case I'm just misreading/misunderstanding.  This looks like it's going to be really powerful!

src/screens/AppTileScreen.tsx
markdlv: Thanks for syncing offline - all clear now.  For my future ref: while in home-screen-tile-tapped context, it's the HomeStack/Screen that's providing nav, thus we don't want this screen to do anything `headerLeft` related.  Makes sense now 👍

markdlv (approved):

PR #416, opened 2023-10-26
feat: analytics events

nicolls1:
## Changes

- Class for analytic events

This provides an interface for users to define strongly typed events that can be emitted through the `createAnalyticsEmitter` builder. Internal SDK events can be directly defined in the `SDKTrackEvents` type and then emitted through `sdkTracker`. `sdkTracker` only makes sense internally and is not exposed to consumers.

Emitted `track` events are not strongly typed for now but consumers will just be passing all of the track events along to their analytics library so it is practically not that useful. 

There will certainly be more types and details to fill in but this is a strong starting point. 

## Screenshots

N/A


src/common/Analytics/AnalyticsEvents.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  public emit<EventType extends AnalyticsEventTypes>(
+    eventType: EventType,
+    ...params: Parameters<AnalyticsEventTypeHandler<EventType>>
+  ) {
+    return this.emitter.emit(eventType, ...params);
+  }
+}
+
+export const analyticsEvents = new AnalyticsEvents();
```
swain, line 41: Should consumers be using this object, or createAnalyticsEmitter? It seems odd that we'd export both.

src/common/Analytics/AnalyticsEvents.ts
```diff
@@ -0,0 +1,70 @@
...
+    reset: () => analyticsEvents.emit('reset'),
+  };
+};
+
+export type SDKTrackEvents = {
+  LoginWithInvite: Record<string, any>;
+};
+
+// class for internal tracking use only
+export const sdkTracker = createAnalyticsEmitter<SDKTrackEvents>();
```
swain, line 70: Might be worth pre-pending with an `_` here for extra clarity that it's "internal"

src/common/Analytics/index.ts
```diff
@@ -0,0 +1,10 @@
+export {
```
swain, line 1: Can we just do `export * from '...'` instead? That will make this file much more maintainable.

tsconfig.json
```diff
@@ -20,11 +20,5 @@
       "@styles": ["./src/components/BrandConfigProvider/styles/types"]
     }
   },
-  "include": [
-    "index.ts",
-    "src/**/*",
-    "lib/**/*",
-    "example/**/*",
```
swain, line 27: It looks this line was removed. What's the motivation there?

swain (commented):
Love it! Just a couple questions.

Also, one high-level suggestion: it'd be great to have some additional documentation/examples of how to use this new functionality. That could either be in the README, or inline via JSDoc.

src/common/Analytics/AnalyticsEvents.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export const analyticsEvents = new AnalyticsEvents();
+
+export type Tracker<CustomEventMap extends Record<string, unknown>> = {
+  track<Key extends keyof CustomEventMap>(
+    name: Key,
+    event: CustomEventMap[Key],
+  ): void;
+  userPropertyUpdate: AnalyticsEventTypeHandlers['userPropertyUpdate'];
+  reset: AnalyticsEventTypeHandlers['reset'];
```
aecorredor, line 49: instead of explicitly adding these here, could we just do `& AnalyticsEventTypeHandlers`? We could then just override `track`.

src/common/Analytics/AnalyticsEvents.ts
aecorredor: If we're already exporting it, and it gets exported from the index, there's really no guarantee that this is only for internal use, correct? 

I'm also not sure what "internal tracking" would mean for our use case. I think if we don't have any internal events mapped out already we don't have to export it yet. And when we do export it, I think we should assume that it could be used by anyone who installs the SDK.

src/common/Analytics/AnalyticsEvents.ts
aecorredor: Also, looking at this more closely, I think we could do with a single generic type/interface, right? Instead of having both `AnalyticsEventTypeHandlers` and `Tracker`?

tsconfig.json
nicolls1: I meant to not commit that, I could not run tests unless the example repo was setup and did not want to go down that rabbit hole, will revert.

src/common/Analytics/AnalyticsEvents.ts
nicolls1: The way I currently have it this is for listening and the createAnalyticsEmitter is for creating, I guess I need a listening specific one as well?

src/common/Analytics/AnalyticsEvents.ts
nicolls1: Internal event are ones that would be created by this SDK. I could add a couple as first examples.

src/common/Analytics/AnalyticsEvents.ts
nicolls1: I was not able to get the types to play nicely with a single handler/tracker object.

src/common/Analytics/AnalyticsEvents.ts
```diff
@@ -0,0 +1,91 @@
...
+  };
+};
+
+export type SDKTrackEvents = {
+  Login: { usedInvite: boolean };
+  PostCreated: { messageLength: number };
+  PostEdited: { messageLength: number };
+};
+
+// class for internal tracking use only
```
swain, line 81: Bonus points for making this JSDoc so that consumers can easily know this isn't for their use.

src/common/Analytics/AnalyticsEvents.ts
```diff
@@ -0,0 +1,91 @@
...
+    return this.emitter.removeListener(eventType, listener);
+  }
+
+  public emit<EventType extends AnalyticsEventTypes>(
+    eventType: EventType,
+    ...params: Parameters<AnalyticsEventTypeHandler<EventType>>
+  ) {
+    return this.emitter.emit(eventType, ...params);
+  }
+}
```
swain, line 40: I'd recommend removing this `class` altogether, since we're not supporting creating multiple instances. Instead, we can just do this:

```typescript
const emitter = new EventEmitter();

// reference `emitter` in `createAnalyticsEmitter` etc.

type AnalyticsEmitter = {
  addListener<EventType extends AnalyticsEventTypes>(
    eventType: EventType,
    listener: AnalyticsEventTypeHandler<EventType>,
  ): void;

  removeListener<EventType extends AnalyticsEventTypes>(
    eventType: EventType,
    listener: AnalyticsEventTypeHandler<EventType>,
  ): void;
};

export const analyticsListener = {
  addListener: (...args) => emitter.addListener(...args),
  removeListener: (...args) => emitter.addListener(...args)
}
```

I think that will slim down the LOC here quite a lot

src/common/Analytics/AnalyticsEvents.ts
```diff
@@ -0,0 +1,91 @@
+import { EventEmitter } from 'events';
+
+export type AnalyticsEventTypeHandlers = {
+  track: (eventName: string, event: unknown) => void;
+  identifyUser: (id: string) => void;
+  userPropertyUpdate: (
+    key: string,
+    value: string | number | boolean | undefined,
+  ) => void;
+  reset: () => void;
```
swain, line 10: What is `reset` supposed to be used for?

swain (commented):
Nice, this is shaping up nicely!

src/common/Analytics/AnalyticsEvents.ts
nicolls1: It is the opposite of `identifyUser`, called when user logs out. I'll make it called `resetIdentifyUser`. I was following the library's naming but we can be more specific.

swain: @nicolls1 I'd suggest "resetUser"

swain (approved):
Beautiful. Awesome work!

src/common/Analytics/AnalyticsEvents.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    console.log('1', listener.mock.calls);
+    analyticsListener.addListener('track', listener);
+    console.log('2', listener.mock.calls);
+    _sdkTracker.track(eventKey, event);
+    console.log('3', listener.mock.calls);
+    analyticsListener.removeListener('track', listener);
+    console.log('4', listener.mock.calls);
+    _sdkTracker.track(eventKey, event);
+
+    console.log('listen', listener.mock.calls);
```
sternetj, line 23: Probably don't need these console logs

src/common/index.ts
```diff
@@ -5,3 +5,4 @@ export * from './Notifications';
 export * from './RootProviders';
 export * from './SecureStore';
 export * from './testID';
+export * from './Analytics';
```
sternetj, line 8: Should we avoid exporting `_sdkTracker`?

src/hooks/useActiveProject.tsx
```diff
@@ -31,6 +32,9 @@ const findProjectAndSubjectById = (
     const defaultSubject = subjects?.find(
       (s) => s.projectId === defaultProject?.id,
     );
+    defaultSubject?.subjectId &&
+      _sdkTracker.identifyUser(defaultSubject?.subjectId);
+    _sdkTracker.userPropertyUpdate('project', defaultProject?.id);
```
sternetj, line 37: Why is `project` considered a "user property"?

src/hooks/useActiveProject.tsx
```diff
@@ -46,6 +50,9 @@ const findProjectAndSubjectById = (
     }
     return getDefault();
   }
+  selectedSubject?.subjectId &&
+    _sdkTracker.identifyUser(selectedSubject?.subjectId);
+  _sdkTracker.userPropertyUpdate('project', selectedProject?.id);
```
sternetj, line 55: Feels like this logic should live where the value is actually changed. If `findProjectAndSubjectById` was ever used just to look up the value this would cause a mutation.

src/hooks/useOAuthFlow.tsx
```diff
@@ -123,13 +125,14 @@ export const OAuthContextProvider = ({
       try {
         const result = await authorize(authConfig);
         await storeAuthResult(result);
+        _sdkTracker.track('Login', { usedInvite: !!inviteId && !!evc });
```
sternetj, line 128: I feel like this might read better if we changed the instance to be `_sdkEvents` it would also help avoid overloading the term `tracker`.

sternetj (approved):
Looks awesome! This is going to be super useful.

src/hooks/useOAuthFlow.tsx
aecorredor: Something like `Events.track` would read nicely.

src/hooks/useActiveProject.tsx
aecorredor: Also, should we only be updating the property when we know there's a project? It seems like in this case we could go from setting it to `undefined` first before setting it back to an actual value.

src/hooks/useActiveProject.tsx
aecorredor: +1. I also noticed these are the exact same lines as above. It seems like they should live in different places since the subject and project don't necessarily change together.

src/common/Analytics/AnalyticsEvents.ts
```diff
@@ -0,0 +1,80 @@
...
+ * userPropertyUpdate: Set a value for the current logged in user.
+ * resetUser: Reset the current user tracking session, for when the user logs out.
+ */
+export const createAnalyticsEmitter = <
+  EventMap extends Record<string, unknown>,
+>(): Tracker<EventMap> => {
+  return {
+    track: (name, payload) => emitter.emit('track', name as string, payload),
+    identifyUser: (id) => emitter.emit('identifyUser', id),
+    userPropertyUpdate: (key, value) =>
```
aecorredor, line 43: nit: since we already have `identifyUser` and `resetUser`, it seems more idiomatic to call this `updateUser`.

src/common/Analytics/AnalyticsEvents.ts
```diff
@@ -0,0 +1,80 @@
...
+  Login: { usedInvite: boolean };
+  PostCreated: { messageLength: number };
+  PostEdited: { messageLength: number };
+};
+
+/**
+ * Class for internal SDK event creation only. You will want to create your own
+ * `createAnalyticsEmitter`
+ */
+export const _sdkTracker = createAnalyticsEmitter<SDKTrackEvents>();
```
aecorredor, line 59: To Teddy's point, maybe we call this `InternalEvents`

src/common/Analytics/AnalyticsEvents.ts
```diff
@@ -0,0 +1,80 @@
...
+    eventType: EventType,
+    listener: AnalyticsEventTypeHandler<EventType>,
+  ): void;
+};
+
+/**
+ * Clients can add/remove listeners with these functions. There is not as much
+ * type safety here but it will mostly be used to just pass along to an analytics SDK.
+ */
+export const analyticsListener: AnalyticsListener = {
```
aecorredor, line 77: What's the main goal of being able to listen for analytic events? Is it to basically provide a hook for any consumer to specify where they want to send their events?

src/common/index.ts
sternetj: I'm not really sure. I think it would still be exported but the underscore usually just indicates that it is internal. I'm not sure what would be best.

src/hooks/useOAuthFlow.tsx
nicolls1: There are a few event emitters, I think I'll do `_sdkAnalyticsEvent` to make clear 🙏

src/navigators/LoggedInStack.tsx
```diff
@@ -80,6 +81,10 @@ export function LoggedInStack() {
 
   const loadingMessage = getLoadingMessage();
 
+  useEffect(() => {
+    activeSubjectId && _sdkAnalyticsEvent.setUser(activeSubjectId);
```
sternetj, line 85: I would think we would want to make the user be the userId and set the subjectId as a property.

src/hooks/useSetUserProfileEffect.ts
```diff
@@ -4,7 +4,7 @@ import { useUpdateUser, useUser } from './useUser';
 
 export const useSetUserProfileEffect = () => {
   const { isLoading, isFetched, data: user } = useUser();
-  const { mutate: updateUser } = useUpdateUser();
```
sternetj, line 7: I think this was probably better before since we are actually updating the user's data here.

nicolls1: Ah, yeah, find all found too many 😮‍💨

sternetj (approved):

src/common/Analytics/AnalyticsEvents.ts
```diff
@@ -0,0 +1,80 @@
...
+  setUser: AnalyticsEventTypeHandlers['setUser'];
+  updateUserProperty: AnalyticsEventTypeHandlers['updateUserProperty'];
+  resetUser: AnalyticsEventTypeHandlers['resetUser'];
+};
+
+/**
+ * Create a type safe way to emit analytics events. Can do the following actions:
+ *
+ * track: Tracks an event with a given name and payload with values relevant to that event.
+ * setUser: Allows you to start tracking a logged in users session by their ID.
```
markdlv, line 33: Nit- "logged in user[']s"

tsconfig.json
markdlv: @nicolls1 is the intention to still remove this diff?

src/common/Analytics/AnalyticsEvents.ts
markdlv: I think type safety in the listener could be important over time, if not just better DX.  Not worrying about it for now / first pass seems fine as we flesh out kinks while wiring this up to the first analytics SDK 👍

markdlv (dismissed):
LGTM.  I bet we'll iteration on this as we start utilizing it to send events to xyz analytics service, which is fine.

markdlv (approved):

aecorredor (approved):

PR #420, opened 2023-10-31
feat: Add PersistQueryClientProvider and user profile hooks

bruce-glazier:
## Changes

- Adds PersistQueryProvider which is a wrapper for PersistQueryClientProvider that embeds the required props and filter functions. The filter functions make it so only specified queries or mutations are written to async-storage. This provider also includes the QueryClientProvider.
- Adds hooks to fetch user profiles


src/hooks/useUserProfiles.tsx
```diff
@@ -0,0 +1,151 @@
...
+        const cachedProfile = await getProfileFromCache(userId);
+        if (!cachedProfile) {
+          await fetchProfileFromApi(userId);
+        }
+      };
+
+      const getUserProfilePromises =
+        !isLoading && accountHeaders ? userIds.map(fetchUserProfile) : [];
+
+      Promise.all(getUserProfilePromises).then(() => {
```
sternetj, line 92: How does this perform for a user with lots of messages and an empty cache? I wonder if we should do this in chunks in the order they appear on the screen.

sternetj (approved):

src/hooks/useUserProfiles.tsx
bruce-glazier: True, I really should put some protection here to be sure this doesn't cause issues at scale. I can do this in a follow up.

src/hooks/useUserProfiles.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const profiles = useRef<Map<string, UserProfile>>(new Map());
+  const { isLoggedIn } = useAuth();
+  const placeHolderProfile = (userId: string) => ({
+    id: userId,
+    profile: {
+      displayName: userId,
+    },
+  });
+  const messageTiles = appConfig.data?.homeTab?.messageTiles;
+  const userIds = uniq(messageTiles?.flatMap((tile) => tile.userIds));
```
markdlv, line 49: This is named `useUserProfiles`, but it looks coupled to messageTiles.  Let's either rename to be more specific, or refactor to be more re-usable (e.g. by taking in `userIds` as a prop).

bruce-glazier: Yeah, good point.

src/hooks/useUserProfiles.tsx
markdlv: +1 yea let's batch this from the get-go - maybe 10-15 users at a time?

markdlv (dismissed):

swain (commented):
I'd _strongly_ recommend modeling this behavior (and caching) using `react-query` rather than rolling with custom `useEffect` calls + custom caching.

I think you could match the behavior you have here using the `useQueries` hook, and passing 1 query object per user.

bruce-glazier:
> I'd _strongly_ recommend modeling this behavior (and caching) using `react-query` rather than rolling with custom `useEffect` calls + custom caching.
> 
> I think you could match the behavior you have here using the `useQueries` hook, and passing 1 query object per user.

I'll take a stab at using `useQueries` here instead and see how it shakes out. I wanted to persist the profiles to storage and react-query does not have a mechanism for that built-in. I don't want to fetch X number of profiles every time that app is relaunched.

src/hooks/useUserProfiles.tsx
swain, line 1: Thinking more, you could actually even use the "query combiner" util from `one-query` to make this even simpler:

```typescript
const createQueryOptions = (
  userId: string,
  fetchProfile: () => Promise<UserProfile>,
  options: UseQueryOptions<UserProfile>,
): UseQueryOptions<UserProfile> => ({
  ...options,
  queryKey: ['profile', userId],
  queryFn: fetchProfile,
});

export const useUserProfiles = (userIds: string[]) => {
  const { apiClient } = useHttpClient();
  const { accountHeaders } = useActiveAccount();

  const queries = useQueries({
    queries: userIds.map((userId) =>
      createQueryOptions(
        userId,
        () =>
          apiClient
            .request(
              'GET /v1/account/users/:userId',
              { userId },
              { headers: accountHeaders },
            )
            .then((res) => res.data),
        { enabled: !!accountHeaders },
      ),
    ),
  });

  // Use this `one-query` util to "combine" the result of `useQueries` into a single query.
  return combineQueries(queries);
};

// Now, use it:

const profiles = useUserProfiles([...]);

if (profiles.status !== 'success') {
  return '...',
}

profiles.data; // UserProfile[]
```

I think taking an approach like this would provide several benefits:
- No need to add the `react-native-cache` dependency.
- We don't have to model nearly as many states, since we can build on the react-query async API.
- No provider required! You can _just_ use the hook.

swain: > I wanted to persist the profiles to storage and react-query does not have a mechanism for that built-in.

@bruce-glazier React Query actually has a solid + mature approach for "persisting" data from the cache to some external store, if you want to do that. Check out these docs: https://tanstack.com/query/v4/docs/react/plugins/persistQueryClient

bruce-glazier: Thank you @swain! I will see about getting this updated.

src/hooks/useUserProfiles.tsx
bruce-glazier: Now handled internally to react-query

src/hooks/useUserProfiles.tsx
bruce-glazier: Refactored and renamed

src/hooks/useUserProfiles.tsx
bruce-glazier: Updated, thanks for steering me in the right direction, looks much nicer.

src/hooks/PersistedQueryProvider.tsx
```diff
@@ -0,0 +1,58 @@
...
+  if (isArray(query.queryKey)) {
+    // Match on first part of key only to ignore unique ids
+    const keyPart = query.queryKey?.[0];
+    if (queriesToPersist.includes(keyPart)) {
+      return true;
+    }
+  }
+
+  return false;
+};
```
swain, line 34: Nice. Just confirming: this will only persist keys that look like `['profile', ...]`, right?

src/hooks/useMessagingProfiles.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const messageTiles = appConfig?.homeTab?.messageTiles;
+  const userIds = messageTiles?.find(
+    (messageTile) => messageTile.id === tileId,
+  )?.userIds;
+
+  const profiles = useMessagingProfiles(userIds);
+
+  const relevantProfiles = compact(
+    profiles.data?.filter((userProfile) => userIds?.includes(userProfile.id)),
+  );
```
swain, line 73: Why do we need `relevantProfiles`? Is there a possibility that `useMessagingProfiles` will return user profiles that _weren't_ included in `userIds`?

src/hooks/useMessagingProfiles.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  return {
+    all: relevantProfiles,
+    others: relevantProfiles.filter((profile) => profile.id !== userData?.id),
+    isLoading: profiles.isLoading,
+    isFetching: profiles.isFetching,
+  };
```
swain, line 80: Returning in this way will unfortunately mean _losing_ some of the great typing from react-query. Example:

```typescript
const profiles = useProfilesForTile('...');

if (profiles.isLoading) {
  return;
}

profiles.data; // UserProfile[] | undefined 😥
```

Depending on the answer to the comment above, I'd suggest trying to return the raw `react-query` result to get improved type narrowing capabilities:

```typescript
const profiles = useProfilesForTile('...');

if (profiles.status !== 'success') {
  return;
}

profiles.data; // UserProfile[]  😊
```

swain (approved):
Nice! I think this is better -- also lays the foundation for persisting other queries from `react-query` if need be.

src/hooks/useMessagingProfiles.tsx
bruce-glazier: Originally useMessagingProfiles returned all profiles, not just a subset. This is just a left over of that.

src/hooks/useMessagingProfiles.tsx
bruce-glazier: Great point, I probably don't even really need this hook anymore now that I think about it. I will address this in my next PR.

src/hooks/PersistedQueryProvider.tsx
bruce-glazier: Yep, that is the case for the time being. I really didn't want to suddenly adopt the default behavior of persisting everything, that seemed like it would be problematic.

PR #422, opened 2023-11-01
Add Configurable Message Tile Tab

sternetj:
## Changes

  - Adds a MessageTileStack so it can be displayed as a tab
    - Uses the first configured MessageTile on the home tab by default but we will want to rework this eventually to not be dependent on the home screen
  - Reworks the message screen to be stack agnostic
  - Add labels to the storybook tabs, changed name to be a guid, and makes tab labels required

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/2295908/d4443e7c-4921-497c-b45d-08f04dcc79a2


src/navigators/configurable-tabs/MessageTileStack.tsx
```diff
@@ -0,0 +1,58 @@
...
+  DirectMessageScreen: 'MessageTileStack/DirectMessage',
+});
+
+export function MessageTileStack() {
+  const { logoHeaderConfig } = useDeveloperConfig();
+  const { params } = useRoute();
+  const queryClient = useQueryClient();
+  const { data: appConfig } = useAppConfig();
+
+  const tileId = appConfig?.homeTab?.messageTiles?.[0]?.id;
```
markdlv, line 35: Do we already, or are you planning, handling what happens if `tileId` is undefined?  This would apply currently and after any break from `homeTab`.

markdlv (dismissed):

src/navigators/configurable-tabs/MessageTileStack.tsx
sternetj: Yeah right now it just shows a blank screen. I suppose we could hide the tab if there isn't one configured. Do you think that would be better? Happy to update.

Ultimately I think we need to rethink some of the message tile work.

markdlv: I'll leave it up to you - we have a number of places that have assume-happy-path logic; this definitely isn't the only place.  Another approach is to add some tasks to do some sweeping "increase defensive code & error handling" PRs when there's time.

src/navigators/configurable-tabs/MessageTileStack.tsx
bruce-glazier, line 1: Small nit: Why is `configurable-tabs` not camel case?

bruce-glazier (approved):
Looks great. I'd like to get my messaging refactor PR merged in first if that is alright. I think there will be less to resolve on this PR.

src/navigators/configurable-tabs/MessageTileStack.tsx
sternetj: I'll fix this

bruce-glazier (approved):

PR #424, opened 2023-11-01
feat: simplify useAppConfig by using one-query hook

swain:
## Changes
The `useAppConfig` hook currently uses a provider + custom logic for fetching -- we can achieve the same result with less code by just using `useRestQuery`.

I had to make a few other changes in response to the refactor. I explain them each below.

### DIFF
--- a/example/storybook/helpers/DataProviderDecorator.tsx
+++ b/example/storybook/helpers/DataProviderDecorator.tsx
@@ -6,9 +6,17 @@ import { HttpClientContextProvider } from '../../../src/hooks/useHttpClient';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { ActiveAccountContext, ActiveProjectContext } from '../../../src/hooks';
 
-export const DataProviderDecorator = (
-  builder?: (adapter: MockAdapter) => void,
-) => {
+export type DataOverrideProviderProps = {
+  children: React.ReactNode;
+  builder?: (adapter: MockAdapter) => void;
+};
+
+const mockQueryClient = new QueryClient();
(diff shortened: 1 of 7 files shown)


example/storybook/helpers/DataProviderDecorator.tsx
swain, line 1: `AppConfigContext` goes away in this PR -- it was used by two Storybook stories to override the app config for the story.

I did a no-behavior-change refactor in this file to allow modeling a "data override" using a component (and not just a story decorator). **Why**: the `ExampleApp` stories render a "full" app, with providers and all. I couldn't use the decorator overrider, because it'd override all the root level API calls. This component allows me to only override screen code.

@sternetj you might find this change interesting.

src/screens/AppTileScreen.test.tsx
```diff
@@ -36,7 +37,11 @@ beforeEach(() => {
 });
 
 test('renders webview with source prop', () => {
-  render(<AppTileScreen navigation={navigation} route={route} />);
```
swain, line 39: This screen ends up calling `useAppConfig`. Now, it needs a `QueryClientProvider`, otherwise it throws.

example/storybook/stories/SettingsScreen/SettingsScreen.stories.tsx
swain, line 1: `AppConfigContext` went away. Updated this file to use the `DataProviderDecorator` to override the data fetching 🙌🏻

jkdowdle (approved):

nicolls1 (approved):

PR #428, opened 2023-11-03
Permit one patient per group message

bruce-glazier:
## Changes

  - Only allow a single patient to be selected when composing a group message. Only providers can be added after the first patient.
  - Filter out any conversations that contain users that are not part of the message tile, i.e., if a user is removed from the tile then the conversation should no longer be displayed.  

## Screenshots

image


sternetj (dismissed):

sternetj:
Not to add more scope but I feel like a warning message somewhere might be nice to alert the provider to why some aren't available. It might also be good to sort the disabled ones to the bottom.

bruce-glazier:
> Not to add more scope but I feel like a warning message somewhere might be nice to alert the provider to why some aren't available. It might also be good to sort the disabled ones to the bottom.

Totally, I see a lot of room for improvement here and have some thoughts. I wanted to get all the PRs for core functionality merged in first before coming back to refine.

sternetj (approved):

PR #429, opened 2023-11-06
Update Tab Bar to Work with Configurable Tabs

sternetj:
⚠️ BREAKING CHANGE

## Changes

  - Makes the tab bar style-able from the tabs config on the api config
  - Removes some existing styles and consolidates them under the tab definition

## Screenshots

Default Styles:
image


UPGRADING.md
```diff
@@ -13,3 +13,126 @@ versions of this library.
...
+```
+
+After:
+
+```jsx
+<DeveloperConfig
+  developerConfig={{
+    componentProps: {
+      TabBar: {
+        tabs: {
```
markdlv, line 56: Shouldn't this be an array?

UPGRADING.md
```diff
@@ -13,3 +13,126 @@ versions of this library.
...
+          },
+        },
+      },
+    },
+  }}
+/>
+```
+
+- The Svg Props for a tabs defined in the `DeveloperConfig` have also been moved
+  into a new style property on the tab.
```
markdlv, line 84: Nit-pick: "into a new `styles` property on each tab."

src/navigators/TabBar.tsx
```diff
@@ -260,57 +262,35 @@ const defaultStyles = createStyles('TabBar', (theme) => ({
...
-  },
-  activeIndicator0View: {
-    backgroundColor: theme.colors.primary,
-  },
-  activeIndicator1View: {
-    backgroundColor: theme.colors.secondary,
-  },
-  activeIndicator2View: {
-    backgroundColor: theme.colors.tertiary,
+    backgroundColor: theme.colors.scrim,
```
markdlv, line 275: Changing this color to the theme's `scrim` seems like a change (beyond just the refactoring to not expect only 3 tabs).  Is that correct?

src/navigators/TabBar.tsx
```diff
@@ -260,57 +262,35 @@ const defaultStyles = createStyles('TabBar', (theme) => ({
...
-    backgroundColor: theme.colors.secondaryContainer,
-  },
-  tabActive2View: {
-    backgroundColor: theme.colors.tertiaryContainer,
+    backgroundColor: theme.colors.background,
   },
   tabInactiveView: {
     height: '100%',
     paddingTop: 2,
+    backgroundColor: theme.colors.background,
```
markdlv, line 285: Same here as comment above, it seems like (from the removed entries below) this used to be using `surface` and you're changing it to use `background`.  Is that correct, and if so are these intentional changes?

sternetj:
@markdlv Regarding the two style changes you called out: Correct, they are not refactors they are changes. It was mostly personal preference because I didn't like how they looked and they also didn't match the styles of the tab editor so I updated them. I don't feel super strongly about them though so I can change them back. Let me know what you think. Here is what it looked like using the existing colors:
image

markdlv:
> @markdlv Regarding the two style changes you called out: Correct, they are not refactors they are changes. It was mostly personal preference because I didn't like how they looked and they also didn't match the styles of the tab editor so I updated them. I don't feel super strongly about them though so I can change them back. Let me know what you think. Here is what it looked like using the existing colors: image

Ok, well it's great to put these into a breaking change PR then.  I also don't have a strong preference so I'm happy to move forward with your findings/changes.  It might be fair to assume anyone upgrading would run their project locally and spot any styling differences, but you might consider putting a note in the UPGRADING doc about doing that since some default colors were changed.

markdlv (dismissed):

bruce-glazier (dismissed):

markdlv (approved):

bruce-glazier (approved):

PR #431, opened 2023-11-06
fix: simplify useExchangeToken hook

swain:
## Changes
This API was simplified to _not_ require passing the `accessToken` in the request body. It's now gather from the (also required) `Authorization` header.


aecorredor (approved):
LGTM, but CI is red.

jkdowdle (dismissed):

jkdowdle (approved):

PR #432, opened 2023-11-06
Display/resize multiple avatar icons for group messages

bruce-glazier:
## Changes

  - Add support for displaying multiple icons

## Screenshots

Updated:
Screenshot 2023-11-09 at 11 23 39 AM

### DIFF
--- a/src/screens/MessageScreen.tsx
+++ b/src/screens/MessageScreen.tsx
@@ -121,20 +121,7 @@ export function MessageScreen<ParamList extends ParamListBase>({
                 }
               />
             }
-            <View>
-              {
-                // TODO: Combine multiple GiftedAvatars
-              }
-              <GiftedAvatar
-                key={selectedProfiles[0].id}
-                user={{
-                  name: selectedProfiles[0].profile.displayName,
-                  avatar: selectedProfiles[0].profile.picture,
-                  _id: selectedProfiles[0].id,
-                }}
-                textStyle={{ fontWeight: '500' }}
-              />
-            </View>
+            <MultiGiftedAvatar profiles={selectedProfiles} />
           </View>
         );
       }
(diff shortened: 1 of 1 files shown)


src/screens/MessageScreen.tsx
```diff
@@ -286,6 +315,15 @@ const defaultStyles = createStyles('MessageScreen', (theme) => {
       color: theme.colors.text,
       fontWeight: '600',
     },
+    multiGiftedAvatarView: {
+      flex: 1,
+      flexWrap: 'wrap',
+      maxHeight: 40,
+      marginRight: 50,
+    },
```
sternetj, line 323: Any chance we could make it round with a background, similar to iOS?
IMG_4065

src/screens/MessageScreen.tsx
```diff
@@ -239,6 +226,48 @@ export const createMessageScreen = <ParamList extends ParamListBase>(
...
 
+type MultiGiftedAvatarProps = {
+  profiles: User[];
+};
+
+const MultiGiftedAvatar = ({ profiles }: MultiGiftedAvatarProps) => {
+  const { styles } = useStyles(defaultStyles);
+  const diameter = profiles.length < 2 ? 40 : 20;
+  const fontSize = diameter / 2;
+  const adjustments = profiles.length === 2 ? -8 : 0;
```
sternetj, line 237: We should probably build this with injected style settings

src/screens/MessageScreen.tsx
```diff
@@ -286,6 +315,15 @@ const defaultStyles = createStyles('MessageScreen', (theme) => {
...
     },
+    multiGiftedAvatarView: {
+      flex: 1,
+      flexWrap: 'wrap',
+      maxHeight: 40,
+      marginRight: 50,
+    },
+    profileView: {
+      height: 20,
+    },
```
sternetj, line 326: I think if we throw a justify center on here it would look a bit better when there are just three. I think it would look like this
image

sternetj (commented):
Thanks for doing this I think this will be a really nice enhancement. Couple of suggestions.

markdlv:
This is looking great!  I'll wait for the feedback items, but LGTM otherwise.

sternetj (approved):
Looks awesome!

PR #433, opened 2023-11-06
Remove PersistedQueryProvider

bruce-glazier:
## Changes

  - No longer persist certain data to storage


src/common/RootProviders.tsx
```diff
@@ -24,7 +25,7 @@ export function RootProviders({
   const { apiBaseURL, theme } = useDeveloperConfig();
 
   return (
-    <PersistedQueryProvider>
+    <QueryClientProvider client={new QueryClient()}>
```
swain, line 28: We'll want to declare this query client in the top-level global scope -- it needs to be stable for the life of the app, and as-is will re-create on each render.

src/common/RootProviders.tsx
```diff
@@ -12,7 +12,8 @@ import { ActionSheetProvider } from '@expo/react-native-action-sheet';
 import { SafeAreaProvider } from 'react-native-safe-area-context';
 import { ThemedNavigationContainer } from './ThemedNavigationContainer';
 import { LoggedInProviders } from './LoggedInProviders';
-import { PersistedQueryProvider } from '../hooks/PersistedQueryProvider';
+import { QueryClientProvider } from '@tanstack/react-query';
+import { QueryClient } from '@tanstack/query-core';
```
swain, line 16: The auto-complete imports have been behaving oddly for me, so I'm not surprised this happened. But, we should import `QueryClient` from `@tanstack/react-query`, not `query-core`.

src/common/RootProviders.tsx
bruce-glazier: Gah, totally - I had this sneaking feeling that I was missing something.

swain (approved):
Nice. Pretty easy refactor!

src/hooks/PersistedQueryProvider.tsx
```diff
@@ -1,62 +0,0 @@
-import { createAsyncStoragePersister } from '@tanstack/query-async-storage-persister';
-import AsyncStorage from '@react-native-async-storage/async-storage';
-import { PersistQueryClientProvider } from '@tanstack/react-query-persist-client';
```
swain, line 3: @bruce-glazier In this PR or a follow-up, I think we can + should remove these two react-query persistence libraries. I think they'll be unused now.

bruce-glazier: I'll grab it now. Sorry for the slop, extra sleepy today.

swain (approved):

PR #435, opened 2023-11-07
feat: simplify useUser and extract useUpdateUser

swain:
## Changes
Adopting `useRestQuery` in `useUser`, and getting some nice type improvements.

### DIFF
--- a/src/hooks/useUser.ts
+++ b/src/hooks/useUser.ts
@@ -1,30 +1,20 @@
-import { useQuery, useQueryClient } from '@tanstack/react-query';
-import { useHttpClient } from './useHttpClient';
 import { useAuth } from './useAuth';
-import { useRestMutation } from './rest-api';
-import { User } from '../types';
+import { useRestCache, useRestMutation, useRestQuery } from './rest-api';
 
 export function useUser() {
   const { authResult } = useAuth();
-  const client = useQueryClient();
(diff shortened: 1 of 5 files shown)


src/hooks/useUser.ts
swain, line 1: After adding `GET /v1/user` to the rest-types, was able to land a couple improvements in this hook super easily:

- adopt `useRestQuery` in `useUser` to get auto-complete + perfect type inference (and reduced lines of code)

- use `useRestCache` to perform auto-completed + type-safe cache update, instead of having to use raw `setQueryData`

src/hooks/useUser.ts
```diff
@@ -1,30 +1,20 @@
...
+    'GET /v1/user',
+    {},
+    { enabled: !!authResult?.accessToken },
   );
+}
 
-  const { mutateAsync: updateUser } = useRestMutation('PATCH /v1/user', {
-    onSuccess: (updatedUser) => {
-      client.setQueryData(['user'], () => updatedUser);
+export const useUpdateUser = () => {
```
swain, line 13: I separated the mutation into a dedicated hook. Why:
- The spread on _old_ line 27 was actually killing some of the nice `react-query` typing of the query result.
- Avoiding that spread by separating this hook put the typing back into place.

jkdowdle (approved):

PR #436, opened 2023-11-07
Remove redundant/unused logic in useActiveAccount

swain:
## Changes
This PR simplifies a number of pieces of logic in `useActiveAccount` that seemed to be no longer useful.

I've highlighted each change below via a comment -- review appreciated.

### DIFF
--- a/src/hooks/useActiveAccount.tsx
+++ b/src/hooks/useActiveAccount.tsx
@@ -7,7 +7,6 @@ import React, {
   useMemo,
 } from 'react';
 import { Account } from '../types/rest-types';
-import { QueryObserverResult } from '@tanstack/react-query';
 import { useAsyncStorage } from './useAsyncStorage';
 import { inviteNotifier } from '../components/Invitations/InviteNotifier';
 import { ProjectInvite } from '../types';
@@ -23,15 +22,13 @@ export type ActiveAccountProps = {
 };
 
 export type ActiveAccountContextProps = ActiveAccountProps & {
-  refetch: () => Promise<QueryObserverResult<Account[], unknown>>;
-  setActiveAccountId: (accountId: string) => Promise<void>;
+  setActiveAccountId: (accountId: string) => void;
   isLoading: boolean;
   isFetched: boolean;
   error?: any;
 };
 
 export const ActiveAccountContext = createContext({
-  refetch: () => Promise.reject(),
   setActiveAccountId: () => Promise.reject(),
   isLoading: true,
   isFetched: false,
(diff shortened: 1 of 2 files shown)


src/hooks/useActiveAccount.tsx
```diff
@@ -23,15 +22,13 @@ export type ActiveAccountProps = {
 };
 
 export type ActiveAccountContextProps = ActiveAccountProps & {
-  refetch: () => Promise<QueryObserverResult<Account[], unknown>>;
```
swain, line 26: `refetch` was not being used outside of the provider

src/hooks/useActiveAccount.tsx
```diff
@@ -23,15 +22,13 @@ export type ActiveAccountProps = {
...
-  setActiveAccountId: (accountId: string) => Promise<void>;
+  setActiveAccountId: (accountId: string) => void;
```
swain, line 25: There was no need for this function to be `async` -- it's just a wrapper around a `setState` call

src/hooks/useActiveAccount.tsx
```diff
@@ -120,43 +116,28 @@ export const ActiveAccountContextProvider = ({
...
   }, [activeAccount?.account?.id, setStoredAccountId]);
 
-  // Clear selected account when
-  // we've detected that the userId has changed
-  useEffect(() => {
-    if (userId !== previousUserId) {
-      accountsResult.refetch();
-      setPreviousUserId(userId);
-    }
-  }, [previousUserId, userId, accountsResult]);
```
swain, line 130: This `useEffect` was actually not necessary (which made "previous user id" tracking unnecessary as well). Why: the entire react-query `QueryClient` gets cleared wiped when the user changes. When that happens, a refetch of the query is triggered automatically.

src/hooks/useActiveAccount.tsx
```diff
@@ -120,43 +109,28 @@ export const ActiveAccountContextProvider = ({
...
-  }, [accountsResult]);
-
   // Handle invite accept
   useEffect(() => {
     const listener = async (acceptedInvite: ProjectInvite) => {
-      cache.invalidateQueries({ 'GET /v1/accounts': 'all' });
-      await refetch();
-      await setActiveAccountId(acceptedInvite.account);
+      cache.resetQueries({ 'GET /v1/accounts': 'all' });
+      setSelectedId(acceptedInvite.account);
```
swain, line 120: `resetQueries` is a "hard reset" -- it wipes the current value from the cache and puts us back into the "initial loading" state. By doing this, we avoid needing to refetch, and more clearly signal "we're starting over"

src/hooks/useActiveAccount.tsx
```diff
@@ -19,19 +18,16 @@ export type ActiveAccountProps = {
   account?: Account;
   accountsWithProduct?: Account[];
   accountHeaders?: Record<string, string>;
-  trialExpired?: boolean;
```
swain, line 22: This `trialExpired` property was not referenced anywhere. Seems like it's fallen out of use.

src/hooks/useActiveAccount.tsx
bruce-glazier: > the entire react-query QueryClient gets cleared wiped when the user changes

That wasn't always the case. Are you certain it is setup that way?

bruce-glazier (approved):

PR #446, opened 2023-11-08
fix: Improve the compose new message screen

bruce-glazier:
## Changes

- Improve the compose new message screen by sorting providers and patients into two separate fields.
- Only providers will be given the patient field which has a limit of one patient.

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/76954025/b679b82a-896b-4d84-9bbe-1a03cf28e616

### DIFF
--- a/src/hooks/useAppConfig.tsx
+++ b/src/hooks/useAppConfig.tsx
@@ -27,7 +27,7 @@ export interface CircleTile {
 
 export interface MessageTile {
   id: string;
-  providerUserIds: string;
+  providerUserIds: string[];
   userIds: string[];
   role: string;
   displayName: string;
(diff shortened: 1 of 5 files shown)


sternetj (approved):
Looks awesome!

src/hooks/useAppConfig.tsx
```diff
@@ -27,7 +27,7 @@ export interface CircleTile {
 
 export interface MessageTile {
   id: string;
-  providerUserIds: string;
+  providerUserIds: string[];
```
markdlv, line 30: Was this just wrong before?  Or breaking change?

markdlv (approved):

src/hooks/useAppConfig.tsx
bruce-glazier: It was just wrong before 😬

PR #450, opened 2023-11-09
feat!: require specifying a top-level `account`

swain:
## Changes
**Reviewers**: I _highly_ recommend reviewing this change:
- by commit, and
- with _Hide Whitespace Changes_

I've intentionally separated the change into three commits:
- Updating the core logic in `useActiveAccount` + `useActiveProject`

- Updating the `ActiveAccountProvider` call sites.

- Updating the `useActiveAccount` call sites.

### DIFF
--- a/example/AppDemo.tsx
+++ b/example/AppDemo.tsx
@@ -21,7 +21,7 @@ function App() {
         apiBaseURL: baseURL,
       }}
     >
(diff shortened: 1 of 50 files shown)


swain:
I'm re-opening a newer, cleaner PR here.

PR #456, opened 2023-11-10
fix: Adjust profile screen padding, typography and add divider

dexterca:
This PR is trying to better match styles found in Figma file. This is my first commit to a React Native project so I thought I'd start small.

## Changes
- Adjust font sizes
- Add fields container padding
- Adjust spacing between fields
- Add divider and adjust top margin

## Screenshots
BEFORE | AFTER
-- | --
Simulator Screenshot - iPhone 15 Pro - 2023-11-10 at 10 47 29 | Simulator Screenshot - iPhone 15 Pro - 2023-11-10 at 10 46 04

## Related PR
https://github.com/lifeomic/fountain-life-member-app/pull/152

### DIFF
--- a/src/screens/ProfileScreen.tsx
+++ b/src/screens/ProfileScreen.tsx
@@ -130,10 +130,7 @@ const Field = ({ label, value }: FieldProps) => {
     <View style={styles.fieldView} testID={tID(label)}>
       <Text style={styles.fieldLabelText}>{label}</Text>
       <Text style={styles.fieldValueText}>{value}</Text>
-      <Divider
-        style={styles.divider}
-        theme={{ colors: { outlineVariant: '#727983' } }}
-      />
+      <Divider style={styles.dividerView} />
     </View>
   );
 };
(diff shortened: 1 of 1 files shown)


jkdowdle (approved):
This looks great! nice improvement, I think the only thing would be to make sure this is desired to go out to other use cases too for Branded Mobile Apps

src/screens/ProfileScreen.tsx
```diff
@@ -127,12 +128,12 @@ const Field = ({ label, value }: FieldProps) => {
...
-        {label}
-      </Text>
-      <Text variant="bodyLarge" style={styles.fieldValueText}>
-        {value}
-      </Text>
+      <Text style={styles.fieldLabelText}>{label}</Text>
+      <Text style={styles.fieldValueText}>{value}</Text>
+      <Divider
+        style={styles.divider}
+        theme={{ colors: { outlineVariant: '#727983' } }}
```
jkdowdle, line 135: Could consider pulling this theme object out of the component

aecorredor: I think the divider should take the color directly from the theme that comes from the root theme provider.

aecorredor (commented):
I'd +1 Josh's comment about not providing a theme object explicitly at that level. Like for other components, there should be a way for the divider to take the global theme color.

aecorredor (approved):

PR #467, opened 2023-11-16
feat!: require project to be loaded before rendering children

swain:
## Changes
**Reviewers:** I've separated the changes here into dedicated commits for maximum reviewability. I highly recommend reviewing by commit, with **Hide Whitespace Changes** turned on.

The commits:
- Modifications to the type contract and implementation of `useActiveProject`, as well as the other core changes to "move" the InviteRequired screen higher up the tree.

- Update internal references to `useActiveProject` to pull out any now-dead conditional logic around the return values.

- Update tests to reflect the changes above^.

- Entirely unrelated: adding a test suite to `usePushNotifications`, to put this PR above the coverage threshold requirement.

## Screenshots
I performed in-depth manual testing of this change in internal apps -- I confirmed that during the various onboarding/login flows, the previous experience is still provided.

That said, any _additional_ manual testing would of course be welcomed.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -210,3 +210,6 @@ const App = () => {
   _not_ render their children until the active account + project have been
   resolved. If there is no active account or project, the `InviteRequired`
   screen will be rendered in the place of children.
+
+- The `useActiveProject` hook now returns non-optional data. You can simplify
+  any conditional logic around those returned values.
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -203,3 +203,10 @@ const App = () => {
   );
 };
 ```
+
+### 6.x -> 7.x
+
+- `ActiveAccountContextProvider` and `ActiveProjectContextProvider` will now
+  _not_ render their children until the active account + project have been
+  resolved. If there is no active account or project, the `InviteRequired`
+  screen will be rendered in the place of children.
--- a/src/hooks/Circles/useJoinCircles.tsx
+++ b/src/hooks/Circles/useJoinCircles.tsx
@@ -11,12 +11,12 @@ export function useJoinCircles() {
   const { httpClient } = useHttpClient();
 
   return useQuery(
-    [`/v1/life-research/projects/${activeProject?.id}/app-config/circles`],
+    [`/v1/life-research/projects/${activeProject.id}/app-config/circles`],
     () => {
       if (data?.homeTab?.circleTiles?.some((c) => !c.isMember)) {
         httpClient
           .patch<CircleTile[]>(
-            `/v1/life-research/projects/${activeProject?.id}/app-config/circles`,
+            `/v1/life-research/projects/${activeProject.id}/app-config/circles`,
             data.homeTab.circleTiles.map((c) => ({ ...c, isMember: true })),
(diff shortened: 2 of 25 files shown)


src/hooks/useActiveProject.test.tsx
swain, line 1: Something nice to notice here: even though `useActiveProject` looks like it has shifted a lot, the existing test cases are basically unchanged, aside from the loading behavior + new test validating the `InviteRequired` screen 🙌🏻.

src/hooks/usePushNotifications.test.tsx
swain, line 1: The changes in this PR are **entirely unrelated** to the rest of the PR. _Why they're here_: the nature of the changes above _reduced_ the code coverage of the project. I'm not sure how. But, in order to get back above the coverage requirement, I had to add new tests. I chose to add this suite to bring the coverage back up, since the logic was previously untested.

src/hooks/useActiveAccount.tsx
```diff
@@ -82,6 +83,13 @@ export const ActiveAccountContextProvider = ({
     };
   }, [cache, setPreferredId]);
 
+  /**
+   * This check is temporary. It will be refactored out in a subsequent PR.
+   */
+  if (accountsResult.status === 'success' && !selectedAccount) {
+    return <InviteRequiredScreen />;
+  }
```
swain, line 91: This approach is not as hacky as I thought it was when I wrote this comment. But, regardless: when the shift to "static account definition" happens, this code will be factored out.

UPGRADING.md
```diff
@@ -203,3 +203,13 @@ const App = () => {
...
 ```
+
+### 6.x -> 7.x
+
+- `ActiveAccountContextProvider` and `ActiveProjectContextProvider` will now
+  _not_ render their children until the active account + project have been
+  resolved. If there is no active account or project, the `InviteRequired`
+  screen will be rendered in the place of children.
+
+- The `useActiveProject` hook now returns non-optional data. You can simplify
```
jkdowdle, line 214: maybe worth adding a small example to illustrate what we mean by this

```diff
--- a/use.ts
+++ b/use.ts
-      project?.id
+      project.id



```

jkdowdle (approved):

UPGRADING.md
swain: @jkdowdle Great idea. Unless there are other changes, I'll add that in a follow-up.

src/hooks/Circles/useJoinCircles.tsx
```diff
@@ -11,12 +11,12 @@ export function useJoinCircles() {
   const { httpClient } = useHttpClient();
 
   return useQuery(
-    [`/v1/life-research/projects/${activeProject?.id}/app-config/circles`],
+    [`/v1/life-research/projects/${activeProject.id}/app-config/circles`],
```
aecorredor, line 14: Strange that this wasn't erroring out before. I'm guessing that the repo doesn't have that strict template rule set in eslint?

aecorredor (commented):
LGTM. I'll defer merging to BMA folks though.

sternetj (approved):

PR #470, opened 2023-11-22
fix: refine modeling of active project provider

swain:
## Changes
Landing a quick refactor of the modeling in #467. I realized that using a higher-order component would make the flow a bit clearer.

**This PR contains no behavior changes**, as evidenced by the unchanged tests.


swain:
Talked offline about this with @aecorredor -- he showed me the light. Not doing this.

PR #478, opened 2023-11-30
feat: support custom consent screen

jkdowdle:
## Changes

  - Support passing a custom Consent screen, similar to custom login and home screens 
    - The custom component will be provided props to be able to render and act on the consent

## Screenshots

image

### DIFF
--- a/src/screens/ConsentScreen.tsx
+++ b/src/screens/ConsentScreen.tsx
@@ -3,7 +3,6 @@ import { View, ScrollView, Text, Alert } from 'react-native';
 import { t } from 'i18next';
 import Markdown from 'react-native-markdown-display';
 import { Button } from 'react-native-paper';
-import { Consent, Questionnaire } from 'fhir/r3';
 import { ActivityIndicatorView, createStyles } from '../components';
 import {
   useStyles,
@@ -11,6 +10,7 @@ import {
   useOAuthFlow,
   useOnboardingCourse,
 } from '../hooks';
+import type { ConsentAndForm } from '../hooks/useConsent';
 import { useDeveloperConfig } from '../hooks/useDeveloperConfig';
 import { LoggedInRootScreenProps } from '../navigators/types';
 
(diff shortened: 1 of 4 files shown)


src/screens/ConsentScreen.tsx
```diff
@@ -157,3 +171,24 @@ declare module '@styles' {
...
+export type CustomConsentScreenProps = {
+  /**
+   * The full Consent and Questionnaire FHIR Resources pertaining to the
+   * consent form. Terms and acceptance text can be derived from here
+   */
+  consentForm:
+    | (Consent & {
+        form: Questionnaire;
+      })[]
+    | undefined;
```
swain, line 184: Two nits:
- I think there is already a shared `ConsentAndForm` type we should use here.
- I think we should only pass a single instance, not an array. Just grab the first one, like the SDK logic does. Thoughts?

jkdowdle: >I think we should only pass a single instance, not an array. Just grab the first one, like the SDK logic does. Thoughts?

I can do that, but I feel like it would be nice in case anyone has needs to act on multiple consents. There is a todo here mentioning that the sdk may eventually support it

>   // TODO: If needed, allow for accepting multiple consents in a row.

swain: I'd say it's fine to ignore that scenario for now for simplicity. The platform in general does not support multiple consents. Supporting that would be a lot of work.

sternetj (approved):

swain (approved):

src/screens/ConsentScreen.tsx
```diff
@@ -157,3 +171,20 @@ declare module '@styles' {
...
+export type CustomConsentScreenProps = {
+  /**
+   * The full Consent and Questionnaire FHIR Resources pertaining to the
+   * consent form. Terms and acceptance text can be derived from here
+   */
+  consentForm: ConsentAndForm | undefined;
+  /** Mutation to accept consent is in flight */
+  isLoadingUpdateConsent: boolean;
+  /** Mutation to accept consent */
+  acceptConsent: () => Promise<void>;
```
swain, line 184: It would be nice if we used mutate instead of mutateAsync here so that this doesn't return a promise. That way, there's no floating promise to handle, and it's clear that error handling is not required.

swain (approved):

aecorredor (approved):

PR #479, opened 2023-12-01
refactor: change use consent hook to be powered by api hook

jkdowdle:
## Changes

  - use rest api hook to power `useConsent`
  - support the type differences of `includeForm: boolean` in looking up consent directives for user


src/types/rest-types.ts
```diff
@@ -85,13 +86,43 @@ export type RestAPIEndpoints = FhirAPIEndpoints & {
...
     Request: {
       projectId: string;
-      includeForm: boolean;
     };
     Response: {
       items: ConsentDirective[];
     };
   };
 
+  'GET /v1/consent/directives/me?includeForm=true': {
```
jkdowdle, line 95: I wasn't quite sure if this would work, but I did test it and appears to work. 

I couldn't think of another way to differentiate the response type (generic didn't seem appropriate since it would be on the top level Rest types) - but part of me kinda likes this

swain: I think there's a simpler change. There are currently no use cases in the SDK for fetching _without_ the form. So, we can just update [this line](https://github.com/lifeomic/react-native-sdk/pull/479/files#diff-bdad403381ead9862cb8dc89cbb6f2ab33b533c48f158f1a06038a11359296f8L88) to be `includeForm: true`, and make the response always `ConsentAndForm[]`.

Does that make sense?

swain:
@jkdowdle Also, this should definitely be a `fix:` or `feat:` commit, not a `refactor:`.

`refactor:` commits don't cause a new release to be cut.

jkdowdle:
@swain I did a force push to fix the semantic commit message, sorry about that. but I do think it is better to review that way since some of the other changes are just not necessary any more

swain (approved):

PR #482, opened 2023-12-01
fix: consent screen use mutate over async mutate

jkdowdle:
## Changes

  - Following up on an early suggestion to switch this to use `mutate` over `mutateAsync`, makes using it a bit simpler especially in the custom screen flow

### DIFF
--- a/src/hooks/useConsent.ts
+++ b/src/hooks/useConsent.ts
@@ -2,7 +2,6 @@ import { useActiveProject } from './useActiveProject';
 import { Consent, Questionnaire } from 'fhir/r3';
 import { useRestQuery, useRestMutation } from './rest-api';
 import { RestAPIEndpoints } from '../types/rest-types';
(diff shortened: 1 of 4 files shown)


src/hooks/useConsent.ts
```diff
@@ -13,9 +17,18 @@ export const useConsent = () => {
...
 
-  const useUpdateProjectConsentDirective = () => {
+  const useUpdateProjectConsentDirective = (
+    options: UseMutationOptions<
+      PatchConsentDirectives['Response'],
+      unknown,
+      PatchConsentDirectives['Request'] & {
+        directiveId: string;
+      }
+    > = {},
```
swain, line 27: Rather than provide the full options here, can we _just_ provide an `onSuccess` option? That will simplify the typing a lot, and make a soon-to-come change from me a lot easier.

PR #487, opened 2023-12-04
fix: remove unused invite events

swain:
## Changes
I'm working the existing issue that flashes the InviteRequired screen. To make subsequent changes cleaner, removing these currently defunct invitation events.


nicolls1 (approved):

swain:
Spoiler alert: the fact that these events became defunct is the root cause of an ongoing bug. But, I think the fastest thing to do is remove them for now. My next PR is going to be a fix for the bug.

PR #488, opened 2023-12-04
feat!: hide app while accepting invite

swain:
## Changes

  - Your changes...


src/hooks/useMe.ts
```diff
@@ -21,9 +20,8 @@ interface MeResponse {
 
 export function useMe() {
   const { httpClient } = useHttpClient();
-  const { lastAcceptedId } = usePendingInvite();
 
-  const useMeQuery = useQuery(['fhir/dstu3/$me', lastAcceptedId], () =>
```
swain, line 26: @bruce-glazier You added this. I think it's no longer needed, since we're going to clear the QueryClient entirely when accepting a new invite now, thanks to this line: https://github.com/lifeomic/react-native-sdk/pull/488/files#diff-c7bf47785b64fa6a95226baae705ac2b51d046cd1846c3c02fca15b17d7a6733R77-R80

Would love a sanity check there.

nicolls1 (approved):

src/common/LoggedInProviders.tsx
```diff
@@ -49,21 +49,23 @@ export const LoggedInProviders = ({ children }: LoggedInProvidersProps) => {
...
-                  {children}
-                </NotificationsManagerProvider>
-              </PushNotificationsProvider>
-              <CreateEditPostModal />
-            </OnboardingCourseContextProvider>
-          </CircleTileContextProvider>
-        </WearableLifecycleProvider>
-      </TrackTileProvider>
-    </ActiveProjectContextProvider>
+    <InviteProvider>
```
aecorredor, line 52: I think I'm missing something: conceptually The `LoggedInProviders` component seems to indicate that the user is already logged in, so at that point an invite has already been accepted. Doesn't the `InviteProvider` play a role before we have a logged in user?

aecorredor (commented):
LGTM, but had a question since I'm probably missing something. Also, I'll defer a second approval to @bruce-glazier.

src/common/LoggedInProviders.tsx
swain: > at that point an invite has already been accepted

This is not correct. The typical flow for a brand new user is:

1. Log in via `evc` + the OpenID issuer.

2. Once logged in, accept the invite (this is an authenticated API call)

3. Once invite is accepted, show consent + rest of app.

jkdowdle (approved):

package.json
```diff
@@ -159,6 +159,7 @@
     "react-native-paper": "^5.3.1",
     "react-native-vector-icons": "^9.2.0",
     "react-navigation-header-buttons": "^9.0.1",
-    "react-use": "^17.4.0"
+    "react-use": "^17.4.0",
+    "zustand": "^4.4.7"
```
jkdowdle, line 163: maybe my only concern is adding this as a dep to the sdk

jkdowdle (approved):

package.json
swain: @jkdowdle Can you elaborate on the concern?

src/common/LoggedInProviders.tsx
swain: @aecorredor So, the role of the `InviteProvider` is to accomplish step 2.

src/hooks/useMe.ts
bruce-glazier: This seems correct to me.

src/common/LoggedInProviders.tsx
aecorredor: Thanks for the clarification.

bruce-glazier (approved):

src/components/Invitations/InviteProvider.tsx
```diff
@@ -1,144 +1,105 @@
...
+export type PendingInvite = {
+  inviteId: string;
   evc?: string;
 };
 
-type InviteState = {
-  inviteParams: InviteParams;
-  clearPendingInvite: () => Promise<void>;
-  lastAcceptedId: string;
+const store = create<PendingInvite | undefined>(() => undefined);
```
sternetj, line 16: Should we call this `usePendingInviteStore`? That seems to be the convention that zustand uses in its readme.

sternetj (approved):

src/components/Invitations/InviteProvider.tsx
swain: @sternetj I like it. I'll follow-up with that improved naming.

swain:
Thanks so much for the expedited review @bruce-glazier @sternetj @nicolls1 @aecorredor @jkdowdle 🙏🏻

PR #493, opened 2023-12-06
FLME-300 [part 1]: extract AuthedWebApplet component

swain:
## Changes
We need to do tighter customization of the surveys screen in an internal app. This PR extracts the "core" logic of `AuthedAppTileScreen` into an `AuthedWebApplet` component.

This PR contains **no behavior changes**, as evidenced by the unchanged tests.

PR #494, opened 2023-12-06
Revert #493, and simplify extracted logic

swain:
## Changes
In adopting #493, I realized it was not a super helpful extraction. So, this PR:
- reverts #493 (we got lucky, it was never released due to an NPM issue
AND
- extracts a more focused piece of logic: the logic to fetch the exchange code + construct an exchange url.


UPGRADING.md
```diff
@@ -298,3 +298,9 @@ const MyComponent = () => {
 - The `react-query` query key for the `useMe()` data changed. The hook is now
   powered by `useRestQuery`. If needing to interact with the cached data, use
   the `useRestCache` hook.
+
+### 11.x -> 12.x
```
markdlv, line 302: I don't think this is too big of a deal, but FWIW it looks like this this PR didn't result in a major semantic release.  e.g. Latest is 11.9.1.  Just documenting here so we can fix this up next time we're in UPGRADING

PR #495, opened 2023-12-07
fix: improve dev log message for failed requests

swain:
## Changes
Improving the local-dev-only error log message for failed HTTP requests.


bruce-glazier (approved):
Nice QOL!

aecorredor (approved):

PR #499, opened 2023-12-12
feat: add operations for fhir procedure request resource

jkdowdle:
## Changes

Unfortunately, this resource type does _not_ currently support a status filter or date sorting. Tested this by hitting the api directly with different options

as fyi sorting by date for observations works something like this

```
    const { data } = await this.client.get(
      `/${encodeURIComponent(account)}/dstu3/Observation`,
      {
        params: {
          patient: 'ef876b82-92d8-433b-b0e6-86365116401d',
          date: ['gt2022-12-12T19:55:24.079Z', 'lt2023-29-12T19:55:24.079Z'],
          _sort: '-date' | 'date',
        },
      },
    );
```

### DIFF
--- a/src/types/fhir-api-types.ts
+++ b/src/types/fhir-api-types.ts
@@ -7,6 +7,7 @@ import {
   Practitioner,
   Questionnaire,
   QuestionnaireResponse,
+  ProcedureRequest,
 } from 'fhir/r3';
 
 // These "better" types strongly type the bundle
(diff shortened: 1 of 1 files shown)


aecorredor (approved):
Regarding the API implementation, do we know what would it take to add add support for further filtering/sorting? I ask because since the endpoint is already generic based on the resource type, I'm assuming that some meta fields from `Observation` are shared among other resources and could make the implementation easier.

swain (approved):

jkdowdle:
> Regarding the API implementation, do we know what would it take to add add support for further filtering/sorting? I ask because since the endpoint is already generic based on the resource type, I'm assuming that some meta fields from `Observation` are shared among other resources and could make the implementation easier.

its a good question, but I did not look - I would hope so too

PR #502, opened 2023-12-15
feat: add subject param

swain:
## Changes
See code.

### DIFF
--- a/src/types/rest-types.ts
+++ b/src/types/rest-types.ts
@@ -122,6 +122,12 @@ export type RestAPIEndpoints = FhirAPIEndpoints & {
 
   'GET /v1/fhir/files/projects/:projectId/files/:fileId': {
     Request: {
+      /**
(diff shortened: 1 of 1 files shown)


nicolls1 (approved):

src/types/rest-types.ts
```diff
@@ -122,6 +122,12 @@ export type RestAPIEndpoints = FhirAPIEndpoints & {
 
   'GET /v1/fhir/files/projects/:projectId/files/:fileId': {
     Request: {
+      /**
+       * If accessing a file owned by a subject, it is recommended to
+       * specify the subject id here.
+       */
```
aecorredor, line 128: Why is it "recommended" and not enforced?

jkdowdle (approved):

src/types/rest-types.ts
swain: Because someone with a very permissive policy won't necessarily need it.

PR #503, opened 2023-12-19
Add Sleep Data Chart

sternetj:
## Changes

  - Adds a new my data chart type called `SleepChart`
  - Sleep chart has different views for the single day and multi day views
  - Single day shows a minute by minute break down of sleep levels for corresponding sleep observations
  - Multi day shows total sleep duration for week and month by day and shows an average sleep duration for year by each month

## Screenshots

<table>
<tr>
 <td>
 <td>
<tr>
 <td>
 <td>
</table>


src/components/MyData/SleepChart/useSleepChartData.ts
```diff
@@ -0,0 +1,88 @@
...
+          ),
+        }),
+        {} as { start?: Date; end?: Date },
+      );
+
+      const newXDomain = scaleTime()
+        .range([0, common.plotAreaWidth])
+        .domain([new Date(start ?? 0), new Date(end ?? 0)]);
+
+      setChartData({ sleepData: newSleepData, xDomain: newXDomain, dateRange });
```
markdlv, line 77: Like we've talked about, whether in this hook or elsewhere, I foresee the need to filter `newSleepData` down to handle multiple sleep records per evening (e.g. one from Oura, one from Garmin -> both synced in from Apple Health).  Unless that logic is already in place elsewhere, just keep it in mind (as perhaps a separate task to eventually get to).  Excited to try this out with my own data, which would serve as an example of that scenario.

src/components/MyData/SleepChart/DailyChart.tsx
```diff
@@ -0,0 +1,284 @@
...
+  default: 0,
+});
+
+const valToName = (value: number) =>
+  ({
+    4: t('sleep-analysis-type-awake', 'Awake'),
+    3: t('sleep-analysis-type-rem', 'REM'),
+    2: t('sleep-analysis-type-light', 'Light'),
+    1: t('sleep-analysis-type-deep', 'Deep'),
+  }[value] ?? '');
```
markdlv, line 209: 🔥 so cool to see these normalized component codings come together!

markdlv (approved):
🔥

src/screens/MyDataScreen.tsx
```diff
@@ -146,12 +147,23 @@ export const MyDataScreen = () => {
...
+                {...component}
+                dateRange={range}
+                padding={Number(styles.container?.paddingHorizontal) * 2}
+                onShare={setExportData}
+              />
+            )}
+            {component.type === 'SleepChart' && (
+              <SleepChart
+                {...component}
+                title="Sleep Analysis"
```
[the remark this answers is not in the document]
sternetj: Oh good catch. That shouldn't be there. The chart title should come from the app config. I'll update this.

src/components/MyData/SleepChart/useSleepChartData.ts
sternetj: I kind of forgot about that discussion. I think it will likely have some display issues with the current solution. I look into it and follow up with a fix.

src/screens/MyDataScreen.tsx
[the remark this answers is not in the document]
sternetj: I'll fix this in a followup pull request.

PR #504, opened 2023-12-19
Add Sleep Chart Data Selection

sternetj:
## Changes

  - Adds data selection tooltips to the sleep data chart

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/2295908/c58fc986-ea3e-47f4-a448-a1fa46e178a0

### DIFF
--- a/src/components/MyData/SleepChart/DataSelector.tsx
+++ b/src/components/MyData/SleepChart/DataSelector.tsx
@@ -1,0 +1,201 @@
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import { NativeTouchEvent, View } from 'react-native';
+import { useCommonChartProps } from '../useCommonChartProps';
+import {
+  G,
+  Rect,
+  Svg,
+  Text,
+  Circle,
+  RectProps,
+  TextProps,
+} from 'react-native-svg';
+import { differenceInMilliseconds } from 'date-fns';
+import { createStyles } from '../../BrandConfigProvider';
+import { useStyles } from '../../../hooks';
+
+export type SleepDataToolTipPreparer = (x: number) =>
+  | undefined
+  | {
+      header: string;
+      title: string;
+      subtitle: string;
+      color: string;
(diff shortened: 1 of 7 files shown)


bruce-glazier (approved):
Looks really nice 👍

src/components/MyData/SleepChart/DataSelector.tsx
```diff
@@ -0,0 +1,201 @@
...
+                  {tooltip.title}
+                </Text>
+                <Text {...styles.tooltipSubtitle} x={16} y={16}>
+                  {tooltip.subtitle}
+                </Text>
+              </G>
+            </G>
+          )}
+        </Svg>
+      )}
```
markdlv, line 135: Lot of hard-coded numbers in here for x and y.  I suppose they're all relative to the parent `<G x={tooltip.tooltipX} y={tooltip.y}>`, and that's how/why we don't want those to be dynamic?

markdlv (approved):
🔥

markdlv:
There's an unacceptable UX delay in the demo video when you tap `DAY` ~0:31.  Is that consistent or was it a fluke?  The first state has nothing changed except `DAY` is greyed, the 2nd state shows a loading indicator on top of an empty chart, and then we see the graph.  The key is that we want the user to see the 2nd state immediately after tap; it's the first state that's I think would be unacceptable/confusing to the average user.

I'd bet you already know about this and agree, but wanted to note it just in case.  Lmk if you're not sure how to make that transition more responsive.

sternetj:
@markdlv The period toggle slowness should now be fixed. Here is a demo:

https://github.com/lifeomic/react-native-sdk/assets/2295908/b437b92a-22a5-4f5c-aa9a-8f2ebd7d57b2

markdlv (approved):

markdlv:
> @markdlv The period toggle slowness should now be fixed. Here is a demo:

Nice cleanup!

PR #506, opened 2023-12-26
fix: improve invite already accepted handling

swain:
## Changes
This handling seems to have become incorrect over the past few months. This makes it correct.

### DIFF
--- a/src/components/Invitations/InviteProvider.tsx
+++ b/src/components/Invitations/InviteProvider.tsx
@@ -37,13 +37,13 @@ export type InviteProviderProps = {
 
 const isInviteAlreadyAcceptedErrorResponse =
   /* istanbul ignore next */
-  (response: unknown) => {
+  (response: unknown): boolean => {
     if (!response) {
       return false;
     }
 
     if (typeof response !== 'object') {
-      return;
+      return false;
     }
 
     if ('code' in response && response.code === 'INVITATION_ALREADY_ACCEPTED') {
--- a/src/components/Invitations/InviteProvider.tsx
+++ b/src/components/Invitations/InviteProvider.tsx
@@ -35,6 +35,32 @@ export type InviteProviderProps = {
   children: React.ReactNode;
 };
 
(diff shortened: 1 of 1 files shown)


src/components/Invitations/InviteProvider.tsx
```diff
@@ -35,6 +35,32 @@ export type InviteProviderProps = {
...
 
+const isInviteAlreadyAcceptedErrorResponse =
+  /* istanbul ignore next */
+  (response: unknown) => {
+    if (!response) {
+      return false;
+    }
+
+    if (typeof response !== 'object') {
+      return;
```
mdlavin, line 46: Should this return a true or false?

src/components/Invitations/InviteProvider.tsx
```diff
@@ -35,6 +35,32 @@ export type InviteProviderProps = {
...
+    }
+
+    if ('code' in response && response.code === 'INVITATION_ALREADY_ACCEPTED') {
+      return true;
+    }
+
+    if (
+      'error' in response &&
+      typeof response.error === 'string' &&
+      response.error.includes('already accepted')
```
mdlavin, line 56: Should the API be modified to have a consistent behavior instead of two ways to return the same thing?

src/components/Invitations/InviteProvider.tsx
swain: Nice catch.

src/components/Invitations/InviteProvider.tsx
swain: @mdlavin Yes, it definitely should. I'd like to consider that change out-of-scope for this PR.

FWIW, as far as I can tell, the behavior is consistent. But, the current SDK logic is simply not correct. I'd like to get this sub-optimal extra handling in place, then follow-up with a deeper investigation to simplify the logic (client-side, server-side, or both)

mdlavin (approved):

PR #518, opened 2024-01-16
Upgrade React Native Paper and Fix Warnings

sternetj:
## Changes

  - Upgrades React Native Paper to fix the warning about `RCTView has a shadow set but cannot calculate shadow efficiently`
  - Removes the `hidden` style from PillarsTile as it also triggered a `RCTView has a shadow set but cannot calculate shadow efficiently` warning
  - Adds a background and radius to the tracker circles to prevent the `RCTView has a shadow set but cannot calculate shadow efficiently` warning
   - Adds a background to the message compose modal to avoid the `RCTView has a shadow set but cannot calculate shadow efficiently` warning
  - Changes the screen params for `AdvancedTrackerEditor` to be serializable to avoid a `Non-serializable values were found in the navigation state.` warning
  - Reworks the `AdvancedTrackerEditor` save button to avoid a `React has detected a change in the order of Hooks` error when setting the right header option
  - Fixes a NaN error when trendline has zero data points

## Screenshots

Here is a video of everything work correctly:

https://github.com/lifeomic/react-native-sdk/assets/2295908/db433f36-4840-41a5-ae91-a9bff8a7cd39


src/components/Messaging/SearchRecipientsModal.tsx
```diff
@@ -125,8 +124,6 @@ const defaultStyles = createStyles('SearchRecipientsModal', (theme) => {
     rootView: {
       backgroundColor: theme.colors.elevation.level1,
       height: Platform.select({ android: '90%', default: '100%' }),
-      width: '90%',
-      marginHorizontal: '5%',
```
sternetj, line 129: Spacing is now handled by the modal. Spacing on the modal was necessary to achieve the same style while also setting a background to avoid the RTCView warning.

src/components/Messaging/SearchRecipientsModal.tsx
```diff
@@ -152,7 +149,6 @@ const defaultStyles = createStyles('SearchRecipientsModal', (theme) => {
       marginVertical: 16,
       marginRight: 16,
       height: 35,
-      width: 55,
```
sternetj, line 155: After the upgrade this was causing the text to be truncated.

src/components/MyData/LineChart/TraceLine.tsx
```diff
@@ -32,7 +32,7 @@ export const TraceLine = (props: Props) => {
   const yDomain = [domainMin, domainMax] as [number, number];
 
   const trend = useMemo(() => {
-    if (showTrend) {
+    if (showTrend && data.length > 1) {
```
sternetj, line 35: We should only show the trend if it has 2+ points. A single point produces a flat trend which isn't meaningful and 0 points causes a NaN error because of division by zero.

src/components/ScreenSurface.tsx
```diff
@@ -35,6 +35,7 @@ export function ScreenSurface({
 const defaultStyles = createStyles('ScreenSurface', (theme) => ({
   surfaceView: {
     flex: 1,
+    height: '100%',
```
sternetj, line 38: After the upgrade a react native paper surface must have a height set to flex properly.

src/navigators/BottomNavigationBar.tsx
```diff
@@ -56,7 +56,7 @@ export const BottomNavigationBar = ({
             ? options.tabBarLabel
             : options.title !== undefined
             ? options.title
-            : route.title;
+            : route.name;
```
sternetj, line 59: After the upgrade the title property was removed. We now default to name, however, in practice this should never happen.

src/screens/MessageScreen.tsx
```diff
@@ -122,8 +122,7 @@ export function MessageScreen<ParamList extends ParamListBase>({
           <View
             style={{
               flexDirection: 'row',
-              flex: 1,
-              maxWidth: 60,
+              width: 60,
```
sternetj, line 125: After the upgrade the message avatars did not display correctly. Setting a static width solved this problem.

markdlv (approved):

PR #520, opened 2024-01-23
fix: do not make app config stale

swain:
## Changes

  - Your changes...


src/hooks/useAppConfig.tsx
```diff
@@ -123,6 +123,10 @@ export const useAppConfig = () => {
   const query = useRestQuery(
     'GET /v1/life-research/projects/:projectId/app-config',
     { projectId: activeProject.id },
+    {
+      // Don't refetch this query, unless it is manually refetched.
+      staleTime: Infinity,
```
sternetj, line 128: I wonder if this is too long. Maybe we reduce this to a few hours so a user doesn't have to kill the app to get a fresh config?

swain: @sternetj fair enough. I'll do something even more reasonable -- 10 minutes. A 10 minute session is a long tiem.

sternetj: Yeah that sounds great to me. I think that will probably be the best of both worlds.

sternetj (approved):

PR #521, opened 2024-01-23
feat: allow for providing dynamic auth config just in time

jkdowdle:
## Changes
Most users of the sdk will not want or need to use this, but this change allows for intercepting the auth config an key moments and allows consumers to make changes to the auth config used 

```
<RootProviders
  authConfig={authConfig}
  modifyAuthConfig={(action, config, previousAuthResult) => {
    if (...) {
      return {
        ...config,
        // make auth config change
      }
    }

    return config
  }}
>
```

### DIFF
--- a/src/common/RootProviders.tsx
+++ b/src/common/RootProviders.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { AuthConfiguration } from 'react-native-app-auth';
 import { AuthContextProvider } from '../hooks/useAuth';
 import { HttpClientContextProvider } from '../hooks/useHttpClient';
-import { OAuthContextProvider } from '../hooks/useOAuthFlow';
+import { ModifyAuthConfig, OAuthContextProvider } from '../hooks/useOAuthFlow';
 import { GraphQLClientContextProvider } from '../hooks/useGraphQLClient';
 import { useDeveloperConfig } from '../hooks/useDeveloperConfig';
 import { BrandConfigProvider } from '../components/BrandConfigProvider';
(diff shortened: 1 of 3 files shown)


src/common/RootProviders.tsx
```diff
@@ -37,7 +39,10 @@ export function RootProviders({
         <AuthContextProvider>
           <HttpClientContextProvider baseURL={apiBaseURL}>
             <GraphQLClientContextProvider baseURL={apiBaseURL}>
-              <OAuthContextProvider authConfig={authConfig}>
+              <OAuthContextProvider
+                authConfig={authConfig}
+                modifyAuthConfig={modifyAuthConfig}
```
aecorredor, line 44: My one thought about this interface is that it would be nicer to have a single getter prop: a function that receives the action being performed + an auth result (if any) and returns an auth config. I'd use the same name, `authConfig`. 

I don't think there's any technical reason that requires us to have two separate props.

jkdowdle: @aecorredor I'll see what I can do - I did go down this path first - but it become a little more convoluted because internally the app config is updated with some extra props https://github.com/lifeomic/react-native-sdk/blob/11461d938c9f4a33e98dde99871a296b2d3c528c/src/hooks/useOAuthFlow.tsx#L89-L97

jkdowdle: here is the alternative https://github.com/lifeomic/react-native-sdk/pull/523

jkdowdle:
merged other https://github.com/lifeomic/react-native-sdk/pull/523

PR #522, opened 2024-01-24
feat: add apps api hooks

jkdowdle:
## Changes

  - Your changes...

### DIFF
--- a/src/types/apps-rest-types.ts
+++ b/src/types/apps-rest-types.ts
@@ -1,0 +1,50 @@
+/**
+ * While performing authorization through an external user-agent (browser)
+ * is recommended and the only way to be OAuth 2.0 compliant, there are
+ * publicly available endpoints that can be used to build a fully native
+ * experience.
+ */
+export type AppsAPIEndpoints = {
+  'POST /auth/v1/api/invite-validations': {
+    Request: {
+      inviteId: string;
+      evc: string;
+    };
+    Response: {
+      userState: {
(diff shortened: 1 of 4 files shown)


sternetj (approved):

src/types/apps-rest-types.ts
```diff
@@ -0,0 +1,50 @@
+/**
```
aecorredor, line 1: This file should be named `apps-auth-rest-types.ts`, right?

src/types/apps-rest-types.ts
```diff
@@ -0,0 +1,50 @@
...
+ * While performing authorization through an external user-agent (browser)
+ * is recommended and the only way to be OAuth 2.0 compliant, there are
+ * publicly available endpoints that can be used to build a fully native
+ * experience.
+ */
+export type AppsAPIEndpoints = {
```
aecorredor, line 7: We should name this `AppsAuthAPIEndpoints`, right?

aecorredor (commented):
Don't we want `AppsAuth` instead of just `Apps` in all occurrences in this PR?

src/types/apps-rest-types.ts
jkdowdle: @aecorredor I think Apps is more appropriate if we ever wanted to add anything else under this origin `https://apps.us.lifeomic.com`, not just the `/auth` routes

PR #523, opened 2024-01-24
feat: allow for providing dynamic auth config based on current auth result

jkdowdle:
## Changes
Alternative pr to #521. Instead of allowing a consumer to intercept the config at the time an action is about to take place, this provides a function that takes the current auth result and then will return the auth config 

Also noticed the setup function for the tests here was async and didn't need to be. Likely drifted from a time it was required


aecorredor (approved):
I like the simplicity of this change better. In regards to what you said about losing the flexibility of knowing what action is about to take place, I feel like that's a separate issue, parallel to selecting the right auth config. I imagine that in that scenario we'd expose some hooks/handlers for listening in to certain actions, like revoke, refresh, etc.

I'm approving this one since I do like it better and I think it still serves our immediate purpose.

sternetj (approved):

PR #524, opened 2024-01-24
feat: support unsubscribe method from add listener

jkdowdle:
## Changes

  - return an short cut to remove event listeners to sdk event notifiers
  - update internal uses instead

To accomplish this a wrapper around node event's `EventEmitter` was added and then used in place

The commits are split between first, adding that wrapper and making use of it then the second was calling `unsubscribe` in place of `removeEventListener` 

I think the next move would be to deprecate the `removeEventListener` methods and then eventually removing it completely - for now it is left in place as to not break any consumers

### DIFF
--- a/src/common/SDKEventEmitter.ts
+++ b/src/common/SDKEventEmitter.ts
@@ -1,0 +1,42 @@
+import { EventEmitter } from 'events';
+
+/**
+ * A wrapper around node's events module for convenience.
+ * Adds unsubscribe method to the returned result of `addListener`
+ */
+export class SDKEventEmitter {
+  public emitter = new EventEmitter();
+
+  public addListener<
+    Type extends string,
+    Listener extends (...args: any[]) => void,
+  >(eventType: Type, listener: Listener) {
+    const subscription = this.emitter.addListener(eventType, listener);
+    const unsubscribe = () => {
+      this.emitter.removeListener(eventType, listener);
+    };
+
+    return {
+      ...subscription,
+      /**
+       * Removes this subscription from the emitter that registered it.
+       * Similar to using emitter.removeEventListener(eventName, listener)
+       * except it does not return the EventEmitter instance.
+       */
+      unsubscribe,
+      /**  @alias unsubscribe */
+      remove: unsubscribe,
+    };
+  }
+
+  public removeListener<
+    Type extends string,
+    Listener extends (...args: any[]) => void,
+  >(eventType: Type, listener: Listener) {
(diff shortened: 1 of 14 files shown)


src/common/SDKEventEmitter.ts
```diff
@@ -0,0 +1,42 @@
...
+      /**  @alias unsubscribe */
+      remove: unsubscribe,
+    };
+  }
+
+  public removeListener<
+    Type extends string,
+    Listener extends (...args: any[]) => void,
+  >(eventType: Type, listener: Listener) {
+    return this.emitter.removeListener(eventType, listener);
```
aecorredor, line 36: Why do we need to expose `removeListener` if `addListener` is now returning `unsubscribe`, which does the same thing?

src/common/SDKEventEmitter.ts
```diff
@@ -0,0 +1,42 @@
...
+    return {
+      ...subscription,
+      /**
+       * Removes this subscription from the emitter that registered it.
+       * Similar to using emitter.removeEventListener(eventName, listener)
+       * except it does not return the EventEmitter instance.
+       */
+      unsubscribe,
+      /**  @alias unsubscribe */
+      remove: unsubscribe,
```
aecorredor, line 28: Why do we need to return both of these if they are the same?

src/common/SDKEventEmitter.ts
jkdowdle: As to not cause a breaking change at this time

src/common/SDKEventEmitter.ts
jkdowdle: I believe I have seen this pattern in other places - some prefer one over the other. `EventEmitter` also returns different aliases like `on` etc

swain: I'd strongly prefer only returning one, to minimize any confusion. And, maybe _only_ returning the `unsubscribe` function (and leaving the `...subscription` out).

jkdowdle: @swain I would like to leave off `...subscription` - but that would cause a breaking change as the way things are today. idk if anyone would be using it right now, but probably best to deprecate over time

Any one have a preference which one do we prefer to keep? I think I like `unsubscribe` but I think most react-native api's now go with `remove`

aecorredor (approved):
In a follow-up, could we comment on why we kept the old methods that had the same effect? It'd be nice to even mark them as `@deprecated`.

PR #526, opened 2024-02-06
feat: allow for providing safe area provider props

jkdowdle:


aecorredor (commented):
Curious, why did we need this in the FL app?

aecorredor (approved):

aecorredor (dismissed):

sternetj (dismissed):
Looks good. Where we not able to just modify the theme we pass via the DeveloperConfigProvider to update the navigator's theme?

jkdowdle:
@aecorredor @sternetj sorry for the force push, but I think the change set will now be more straight forward. Just the safe area props, nothing with the navigation

sternetj (approved):

PR #529, opened 2024-02-14
feat: Add Invite Loading State

sternetj:
## Changes

  - Adds the ability to set some loading state for an incoming invite
  - Adds logic to the login screen to disable and trigger a loading wheel when an invite is being loaded
  - Adds the ability to inspect the invitation error with a long press

## Screenshots

https://github.com/lifeomic/react-native-sdk/assets/2295908/4d7f8a14-61b0-47b3-b79b-55a2f7ec03b8


bruce-glazier (approved):
Nice! This should help tremendously.

PR #531, opened 2024-02-14
first round of updates

nicolls1:
## Changes

  - Your changes...


nicolls1:
I believe I updated all for all of the current features in `@lifeomic/react-client` but let me know if I missed any

src/hooks/useHttpClient.tsx
```diff
@@ -6,8 +6,8 @@ import axios, {
 } from 'axios';
 import { useAuth } from './useAuth';
 import { APIClient } from '@lifeomic/one-query';
-import { RestAPIEndpoints } from '../types/rest-types';
```
aecorredor, line 9: Can we now delete the types defined in this file?

src/hooks/useConsent.ts
```diff
@@ -104,6 +113,36 @@ export const useConsent = () => {
     useConsentDirectives,
     useUpdateProjectConsentDirective,
     useShouldRenderConsentScreen,
+  } as {
```
aecorredor, line 116: Why did we need this cast?

src/hooks/rest-api.tsx
```diff
@@ -13,7 +32,18 @@ const hooks = createAPIHooks<RestAPIEndpoints>({
...
+  Route extends keyof Endpoints & string,
+  Data = Endpoints[Route]['Response'],
+>(
+  route: Route,
+  payload: RequestPayloadOf<Endpoints, Route>,
+  options?: RestrictedUseQueryOptions<
+    Endpoints[Route]['Response'],
+    unknown,
+    Data
+  >,
```
aecorredor, line 45: Why did we need this explicit type? I think we didn't have it when we used the library on the FL member app.

src/hooks/useHttpClient.tsx
swain: Yep, +1 -- there are several big type defs defined in this project that should be deleted.
- fhir api types
- rest api types
- auth api types

src/hooks/useConsent.ts
nicolls1: See https://lifeomic.slack.com/archives/C05KMKQNTAR/p1707948259517759 or the failed earlier builds. Happy to fix some other way but that is what was decided there.

src/hooks/useHttpClient.tsx
nicolls1: I had a bunch of test errors come up when deleting and moving over so I assumed we needed both but will remove.

src/hooks/rest-api.tsx
nicolls1: See https://lifeomic.slack.com/archives/C05KMKQNTAR/p1707948259517759 or the failed earlier builds. Happy to fix some other way but that is what was decided there.

nicolls1: We are not building the FL member app as an SDK is probably the biggest difference.

src/hooks/useConsent.ts
nicolls1: We are not building the FL member app as an SDK is probably the biggest difference.

nicolls1:
To be clear, tests are working but the build is what is requiring these explicit types because of this setting ts config setting `"declaration": true,`. I am not able to find another solution but open for suggestions.

swain (approved):

aecorredor (approved):

PR #535, opened 2024-02-29
Allow skip invite params

bruce-glazier:
## Changes

  - Add config option to not forward inviteId and evc to the auth session


sternetj (approved):

rob-snyder (approved):
Is this so Readout members can get into the app without receiving an invite from our platform?

bruce-glazier:
> Is this so Readout members can get into the app without receiving an invite from our platform?

It is so Readout members do not trigger the evc flow in auth-apps. Since they are using SSO we do not want a user account automatically created for them by the invite/evc.

PR #541, opened 2024-03-21
fix: Add messaging placeholder

bruce-glazier:
## Changes

  - Add explanation for when the messages screen is empty
  - Add explanation for when the compose messaging screen has no recipients

## Screenshots

Screenshot 2024-03-21 at 9 33 51 AM
Screenshot 2024-03-21 at 9 44 43 AM

### DIFF
--- a/src/screens/MessageScreen.tsx
+++ b/src/screens/MessageScreen.tsx
@@ -236,6 +236,18 @@ export function MessageScreen<ParamList extends ParamListBase>({
             </TouchableOpacity>
           </Swipeable>
         ))}
+        {!conversations?.length && !isLoading && (
+          <Swipeable>
+            <List.Item
+              titleStyle={styles.listItemText}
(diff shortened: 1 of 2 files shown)


src/screens/MessageScreen.tsx
```diff
@@ -236,6 +236,18 @@ export function MessageScreen<ParamList extends ParamListBase>({
             </TouchableOpacity>
           </Swipeable>
         ))}
+        {!conversations?.length && !isLoading && (
+          <Swipeable>
```
sternetj, line 240: What happens if I try to swipe this?

sternetj (approved):

src/screens/MessageScreen.tsx
bruce-glazier: Since no swipeable left or right props are passed, nothing happens.

PR #545, opened 2024-04-01
feat: Add app switcher button

bruce-glazier:
## Changes

  - Your changes...

### DIFF
--- a/src/hooks/useAppConfig.test.tsx
+++ b/src/hooks/useAppConfig.test.tsx
@@ -1,6 +1,10 @@
 import React from 'react';
 import { renderHook, waitFor } from '@testing-library/react-native';
-import { useAppConfig, AppTile } from './useAppConfig';
+import {
+  useAppConfig,
+  AppTile,
+  AppConfigContextProvider,
+} from './useAppConfig';
 import { useActiveProject } from './useActiveProject';
 import { HttpClientContextProvider } from './useHttpClient';
 import { createRestAPIMock } from '../test-utils/rest-api-mocking';
@@ -27,7 +31,9 @@ const renderHookInContext = async () => {
     wrapper: ({ children }) => (
       <QueryClientProvider client={new QueryClient()}>
         <ActiveAccountProvider account="mockaccount">
-          <HttpClientContextProvider>{children}</HttpClientContextProvider>
+          <AppConfigContextProvider>
+            <HttpClientContextProvider>{children}</HttpClientContextProvider>
+          </AppConfigContextProvider>
         </ActiveAccountProvider>
       </QueryClientProvider>
     ),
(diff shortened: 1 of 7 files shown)


src/screens/HomeScreen.tsx
```diff
@@ -32,21 +34,34 @@ export const HomeScreen = (navProps: HomeStackScreenProps<'Home'>) => {
...
+        <AppConfigSwitcherButton />
+      </>
     ),
     [refreshing, refresh],
   );
 
+  const headerLeftAppPicker = useCallback(
+    () => <AppConfigSwitcherButton />,
+    [],
+  );
```
sternetj, line 48: I think this would be a breaking change since it will now always show the app switcher. I wonder if we could make it opt in, similar to the refresh button.

src/hooks/useAppConfig.tsx
```diff
@@ -118,15 +118,27 @@ export interface AppConfig {
   };
 }
 
-export const useAppConfig = () => {
+export const useAppConfigList = () => {
   const { activeProject } = useActiveProject();
 
   const query = useRestQuery(
-    'GET /v1/life-research/projects/:projectId/app-config',
+    'GET /v1/life-research/projects/:projectId/app-configs/list',
```
sternetj, line 125: Somewhere we will need to make this smarter, probably at the API level. Currently this list will contain all configs that exist for the project but an App user should likely only be able to change between configs that they belong to (cohort or user list). This is probably okay until we identify how to manage that logic. I imagine it could end up being different per client.

sternetj (approved):

src/screens/HomeScreen.tsx
bruce-glazier: Indeed, I mean't to remove this from here.

src/hooks/useAppConfig.tsx
bruce-glazier: Total oversight on my part, I'll work on updating API side

src/screens/HomeScreen.tsx
bruce-glazier: Done

sternetj (dismissed):

src/hooks/useAppConfig.test.tsx
```diff
@@ -42,17 +48,21 @@ beforeEach(() => {
...
 test('configured appTiles are returned', async () => {
   const mockAppTiles = ['appTile-1', 'appTile-2', 'appTile-3'].map(mockAppTile);
-  api.mock('GET /v1/life-research/projects/:projectId/app-config', {
-    status: 200,
-    data: {
-      homeTab: {
-        appTiles: mockAppTiles,
-      },
+
+  // Bug in mocker seems to not be able to tell /:id and /list apart
```
sternetj, line 52: I think it is thinking that `list` is the `:id`. Did you try swapping the definition order of the route mocks? It seems like `msw/node` [prepends](https://mswjs.io/docs/api/setup-worker/use/) the routes so I would think/hope that mocking `/:id` then mocking `/list` would produce the desired result. Regardless though this seems fine.

sternetj (approved):

src/hooks/useAppConfig.test.tsx
bruce-glazier: I'd expect that if `/list` matches multiple routes that the mocking library would prefer the exact match over the wildcard which is sort of why I described this as a bug but perhaps you are right about that.

PR #547, opened 2024-04-03
Upgrade apps auth

bruce-glazier:
## Changes

  - Your changes...


bruce-glazier:
Prefer another PR that brings in the same change

PR #557, opened 2024-05-17
fix: Modify image message bubble styles

sternetj:
## Changes

  - Adjusts the default image styles to not have a background and not display the time
  - Adds logic to make the image display with the correct aspect ratio
  - Tweaks the lightbox settings to avoid the black background display when interacting with the image (open/close)

## Screenshots

<table>
<tr>
 <td><b>Before
 <td><b>After
<tr>
 <td>
 <td>
</table>

Text messages still have bubbles:
image


rob-snyder (approved):

bruce-glazier (approved):
Looks great!

rob-snyder (approved):

PR #562, opened 2024-05-23
fix; Keyboard interactions with messages

sternetj:
## Changes

  - Fixes some display issues with the keyboard showing and hiding
  - Hide the keyboard when selecting images to make sure the accessory view isn't cut off
  image

  - Moves the accessory view styling to the correctly typed place

## Screenshots

<table>
<tr>
 <td><b>Before
 <td><b>After
<tr>
 <td>image
 <td>image
<tr>
 <td>image
 <td>image
</table>

### DIFF
--- a/src/screens/DirectMessagesScreen.tsx
+++ b/src/screens/DirectMessagesScreen.tsx
@@ -23,6 +23,7 @@ import {
   ScrollView,
   Image,
   Dimensions,
+  Keyboard,
 } from 'react-native';
 import {
   useInfinitePrivatePosts,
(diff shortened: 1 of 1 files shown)


sternetj:
I set out to fix an issue where some extra spacing is shown after uploading an image. I think the gifted chat library isn't updating it's height when the image is uploaded but I can't figure out what needs to change to fix it.
image

src/screens/DirectMessagesScreen.tsx
```diff
@@ -134,36 +145,38 @@ export const DirectMessagesScreen = ({
...
+      await Promise.all(
+        newMessages.map(async (m) => {
+          const isLastMessage =
+            newMessages.indexOf(m) === newMessages.length - 1;
+
+          await mutateAsync({
+            userIds: users.map((user) => user.id),
+            post: {
+              message: m.text,
+              ...(isLastMessage
```
bruce-glazier, line 157: Just for my own knowledge and I should have asked this on the original PR, what is the reason attachments are only included if the message is the last message?

bruce-glazier (approved):
Left a question but LGTM

src/screens/DirectMessagesScreen.tsx
sternetj: It's just a quirk of the gifted chat interface to say that there could be multiple messages but for our use-case I believe there is only ever one message. I just wanted to make sure the attachments were only added once in case I was wrong about why there could be multiple messages passed to the callback.

PR #563, opened 2024-05-23
Retry queries on interval

bruce-glazier:
## Changes

  - Allow specifying the example apps account parameter from the config file
  - Throttle query retries

### DIFF
--- a/src/hooks/useActiveProject.tsx
+++ b/src/hooks/useActiveProject.tsx
@@ -156,13 +156,19 @@ export const ActiveProjectContextProvider: React.FC<ActiveProjectContextProvider
     const state = calculateState();
 
     useEffect(() => {
-      if (
-        !query.isRefetching &&
-        state.status === 'not-a-patient' &&
-        keepWaitingForPatientId
-      ) {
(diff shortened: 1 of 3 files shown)


src/hooks/useActiveProject.tsx
```diff
@@ -156,13 +156,19 @@ export const ActiveProjectContextProvider: React.FC<ActiveProjectContextProvider
...
 
     useEffect(() => {
-      if (
-        !query.isRefetching &&
-        state.status === 'not-a-patient' &&
-        keepWaitingForPatientId
-      ) {
-        query.refetchAll();
-      }
+      const intervalId = setInterval(() => {
```
sternetj, line 159: Don't we have logic elsewhere to check for the subject existence? Maybe it would make sense to add an event emitter. Not sure if we have other cases where this might be needed.

sternetj (approved):

src/hooks/useActiveProject.tsx
bruce-glazier: I will take this is a follow up task after this PR.

